PR #18, opened 2022-01-07
Circuit construction api improvements

Kixiron:


ryzhyk:
@Kixiron , are there any parts of this that haven't been merged as part of another PR?  If so, would you like to create a new PR for that?

Kixiron:
I think everything else has been more or less covered

PR #30, opened 2022-02-02
Improved algebraic trait impls.

ryzhyk:
See commit messages.

### DIFF
--- a/src/algebra/finite_map.rs
+++ b/src/algebra/finite_map.rs
@@ -49,7 +49,7 @@ where
     fn support_size(&self) -> usize;
 
     /// Increase the value associated to `key` by the specified `value`
-    fn increment(&mut self, key: &KeyType, value: &ValueType);
+    fn increment(&mut self, key: &KeyType, value: ValueType);
 
     /// Create a map containing a singleton value.
     fn singleton(key: KeyType, value: ValueType) -> Self;
@@ -171,7 +171,7 @@ where
     {
         let mut result = FiniteHashMap::new();
         for (k, v) in iter {
-            result.increment(&k, &v);
+            result.increment(&k, v);
         }
         result
     }
@@ -198,7 +198,7 @@ where
         self.value.len()
     }
 
-    fn increment(&mut self, key: &KeyType, value: &ValueType) {
+    fn increment(&mut self, key: &KeyType, value: ValueType) {
         if value.is_zero() {
             return;
         }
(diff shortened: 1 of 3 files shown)


src/algebra/finite_map.rs
```diff
@@ -317,39 +317,102 @@ where
...
+                let entry = self.value.entry(k);
+                match entry {
+                    Entry::Vacant(e) => {
+                        e.insert(v);
+                    }
+                    Entry::Occupied(mut e) => {
+                        e.get_mut().add_assign(v);
+                        if e.get().is_zero() {
+                            e.remove_entry();
+                        };
```
Kixiron, line 340:
```suggestion
                        }
```

src/algebra/finite_map.rs
```diff
@@ -317,39 +317,102 @@ where
...
+                let entry = other.value.entry(k);
+                match entry {
+                    Entry::Vacant(e) => {
+                        e.insert(v);
+                    }
+                    Entry::Occupied(mut e) => {
+                        e.get_mut().add_assign(v);
+                        if e.get().is_zero() {
+                            e.remove_entry();
+                        };
```
Kixiron, line 356:
```suggestion
                        }
```

src/algebra/finite_map.rs
```diff
@@ -317,39 +317,102 @@ where
...
+                let entry = result.value.entry(k.clone());
+                match entry {
+                    Entry::Vacant(e) => {
+                        e.insert(v.clone());
+                    }
+                    Entry::Occupied(mut e) => {
+                        e.get_mut().add_assign_by_ref(v);
+                        if e.get().is_zero() {
+                            e.remove_entry();
+                        };
```
Kixiron, line 384:
```suggestion
                        }
```

src/algebra/finite_map.rs
```diff
@@ -317,39 +317,102 @@ where
...
```
Kixiron, line 403:
```suggestion
                        }
```

Kixiron (commented):
I think the arms of the two sides could be factored out into functions where you just reverse the order of the arguments, that'd be nice in removing some of the duplicate code.
There's also some more if statements with semicolons around the code you changed that GitHub wouldn't let me suggest on if you feel like fixing those up.
I'm also not sure I like the `.support_size()` function name, it's not immediately apparent as to what it means

ryzhyk:
> I think the arms of the two sides could be factored out into functions where you just reverse the order of the arguments, that'd be nice in removing some of the duplicate code. There's also some more if statements with semicolons around the code you changed that GitHub wouldn't let me suggest on if you feel like fixing those up. I'm also not sure I like the `.support_size()` function name, it's not immediately apparent as to what it means

Thanks, fixed these issues and added some more optimizations.

src/algebra/finite_map.rs
```diff
@@ -317,35 +315,109 @@ where
...
+                            e.remove_entry();
+                        }
+                    }
+                }
+            }
+            this
+        }
+
+        if self.support_size() > other.support_size() {
+            do_add(self.clone(), other)
```
mbudiu-vmw, line 392: both these do the same amount of work algorithmically. this optimization is justified only if clone is much faster than inserting all entries.

mbudiu-vmw (approved):

src/algebra/finite_map.rs
ryzhyk: That's right, cloning the map should be more efficient than element-by-element insertion.

PR #31, opened 2022-02-03
A few zset tests

mbudiu-vmw:


src/algebra/zset.rs
```diff
@@ -301,14 +301,14 @@ fn checked_zset_integer_weights_tests() {
...
-    right: i64,
+pub(crate) struct TestTuple {
+    pub(crate) left: i64,
+    pub(crate) right: i64,
 }
 
 #[cfg(test)]
 impl TestTuple {
-    fn new(left: i64, right: i64) -> Self {
+    pub fn new(left: i64, right: i64) -> Self {
```
ryzhyk, line 311: I suggest moving all tests into a module (within the same file), so you only need to write one `#[cfg(test)]` annotation for the whole module. 

From https://doc.rust-lang.org/book/ch11-03-test-organization.html:
```
The convention is to create a module named tests in each file to contain the test functions and to annotate the module with cfg(test).
```

src/circuit/circuit_builder.rs
```diff
@@ -139,10 +146,10 @@ impl<C, D> Stream<C, D> {
     /// Returns `Some` if the operator has produced output for the current timestamp and `None`
     /// otherwise.
     ///
-    /// #Safety
+    /// # Safety
     ///
     /// The caller must have exclusive access to the current stream.
-    pub(super) unsafe fn get(&self) -> &Option<D> {
+    pub unsafe fn get(&self) -> &Option<D> {
```
ryzhyk, line 152: This should not be public. Consider `pub(crate)` instead.

ryzhyk: PS.  Also, it is now called `take` in my PR.

src/circuit/operator/apply.rs
```diff
@@ -37,3 +37,31 @@ where
         (self.func)(i1)
     }
 }
+
```
ryzhyk, line 40: Why do we need this operator when we can just write `stream.apply(|x| x.to_string())`?

src/circuit/operator/differentiatior.rs
```diff
@@ -0,0 +1,57 @@
+// Differentiator operator: direct implementation, without an explicit z
```
ryzhyk, line 1:
```suggestion
//! Differentiator operator: direct implementation, without an explicit z
```

src/circuit/operator/apply.rs
```diff
@@ -37,3 +37,31 @@ where
...
+/// Converts any type to a string
+pub struct Stringifier {}
+
+impl Stringifier {
+    #[allow(dead_code)]
```
ryzhyk, line 45: You need to re-export this operator from `circuit/operator/mod.rs`. Then this won't be dead code. Same for `Differentiator`.

src/circuit/operator/differentiatior.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use crate::algebra::GroupValue;
+use crate::circuit::operator_traits::{Operator, UnaryOperator};
+use std::borrow::Cow;
+
+pub struct Differentiator<V>
```
ryzhyk, line 7: I've been naming operators as actions ("Integrate", "Inspect", "Apply", etc.).  For consistency this would have to be `Differentiate`.  (I know there is also an "Integrator", but that one will go away before we merge this PR).

src/circuit/operator/differentiatior.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl<V> UnaryOperator<V, V> for Differentiator<V>
+where
+    V: GroupValue,
+{
+    fn eval(&mut self, i: &V) -> V {
+        let result = i.add_by_ref(&self.previous.neg_by_ref());
+        self.previous = i.clone();
+        result
+    }
```
ryzhyk, line 56: I think we can do better using `neg` instead of `neg_by_ref`:

```
    fn eval(&mut self, i: &V) -> V {
        let mut old = i.clone();
        swap(&mut old, &mut self previous);
        self.previous.add_by_ref(&self.old.neg());
        self.previous = i.clone();
        result
    }
```

ryzhyk: We should also write an `eval_owned` that avoids cloning `i`.

src/circuit/operator/generator.rs
```diff
@@ -8,48 +8,55 @@ use std::borrow::Cow;
...
     T: Clone,
 {
-    /// Creates a generator with initial `seed` and state transformer function.
-    pub fn new(seed: T, update: F) -> Self {
+    /// Creates a generator with initial value `current` and state transformer function.
+    pub fn new(current: T, state: S, update: F) -> Self {
         Self {
-            seed: seed.clone(),
-            val: seed,
+            current: current.clone(),
```
ryzhyk, line 25: The naming is of fields is confusing. It seems that `current` should be called `initial`,  and `val` should be called `output`.
Also, do we need to store the `output` inside the operator?  We already have `state` for anything that should survive till the next clock tick.

ryzhyk: An alternative way to generalize this operator is to remove all state except closure and have the closure capture any state variables upon operator creation.  I should have thought of that earlier.

src/circuit/operator/inspect.rs
```diff
@@ -47,3 +47,32 @@ where
         (self.callback)(i)
     }
 }
+
+// Sink operator that prints all elements in its input stream.
```
ryzhyk, line 51: I really hate to maintain all these trivial operators.  This should be just `stream.inspect(|x| println!("new input {}", x))`.

src/circuit/zsetoperator/add.rs
```diff
@@ -0,0 +1,56 @@
+use crate::algebra::finite_map::KeyProperties;
```
ryzhyk, line 1: This is the same thing as the `Plus` in my PR, but less general.

src/circuit/zsetoperator/filter.rs
```diff
@@ -0,0 +1,71 @@
+//! Defines an operator whose input is a stream of z-sets and which applies
```
ryzhyk, line 1: This operator could be much more generic.  You only need these trait bounds on elements of the input stream:
```
Z: IntoIterator<Item=(T,W)>,
&Z: IntoIterator<Item=&(T,W)>,
```
and the output stream (which does not have to be the same ZSet type as input):
```
O: FromIterator<Item=(T,W)>
```

Unless we expect that Z sets can be more efficiently filtered without converting them into an iterator.

src/circuit/zsetoperator/filter.rs
```diff
@@ -0,0 +1,71 @@
...
+impl<T, W, F, Z> UnaryOperator<Z, Z> for ZSetFilter<T, W, F, Z>
+where
+    T: KeyProperties,
+    W: ZRingValue,
+    Z: ZSet<T, W>,
+    F: Fn(&T) -> bool + 'static,
+{
+    fn eval(&mut self, i: &Z) -> Z {
+        i.filter(&self.filter)
+    }
```
ryzhyk, line 70: Add an optimized `eval_owned`.

src/test/mod.rs
```diff
@@ -0,0 +1,255 @@
+/*
```
ryzhyk, line 1: Since you're writing all these nice tests anyway, lets wrap them in a `Runtime` and attach a trace monitor to each test for extra testing.  See `circuit_builder.rs` for examples.

ryzhyk (changes requested):
I am happy to make changes proposed in my comments and also rebase this on my PR to eliminate duplication.

src/circuit/operator/integrate.rs
```diff
@@ -0,0 +1,177 @@
...
+    ///
+    /// Streams in the above example will contain the following values:
+    ///
+    /// ```text
+    /// stream:      1, 1, 1, 1, 1, ...
+    /// sum:         1, 2, 3, 4, 5, ...
+    /// delayed_sum: 0, 1, 2, 3, 4, ...
+    /// ```
+    pub fn integrate_core(&self) -> (Stream<Circuit<P>, D>, Stream<Circuit<P>, D>) {
+        // Integration circuit:
```
mbudiu-vmw, line 57: maybe you want to draw the second output on this diagram

src/circuit/operator/plus.rs
```diff
@@ -0,0 +1,290 @@
...
+            OwnershipPreference::prefer_owned(),
+        )
+    }
+}
+
+/// Operator that computes the sum of each value in the stream with itself.
+// TODO: A more efficient implementation would rely on a specialized trait.
+// E.g., doubling weights in a z-set can be done more efficiently than adddin
+// it to itself.
+pub struct Times2<D> {
```
mbudiu-vmw, line 112: can't imagine many queries where this would show up. More likely to be written as a map (apply).

src/circuit/operator/plus.rs
```diff
@@ -0,0 +1,290 @@
...
+    ///     source1.plus(&source2).inspect(|n| assert_eq!(*n, 0));
+    /// })
+    /// .unwrap();
+    ///
+    /// # for _ in 0..5 {
+    /// #     root.step().unwrap();
+    /// # }
+    /// ```
+    pub fn plus(&self, other: &Stream<Circuit<P>, D>) -> Stream<Circuit<P>, D> {
+        if self.origin_node_id() == other.origin_node_id() {
```
mbudiu-vmw, line 43: you are embedding a little optimizer here, I think this should really be the job of the compiler, which may have a cost model.

src/test/mod.rs
mbudiu-vmw: you didn't do this, did you?

src/test/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+                                TestTuple::new(l.left, r.right)
+                            })
+                        };
+                    let join_op = Apply2::new(
+                        move |left: &ZSetHashMap<TestTuple, i64>,
+                              right: &ZSetHashMap<TestTuple, i64>| {
+                            join_func(left, right)
+                        },
+                    );
+                    let join_output = child
```
mbudiu-vmw, line 208: we don't have a nice syntax for inserting an Apply2?

mbudiu-vmw (commented):
Can't approve my own PR

src/circuit/operator/plus.rs
ryzhyk: Yep, this is why I didn't try to optimize it now.

src/circuit/operator/plus.rs
ryzhyk: I agree. This is why I only implemented a really simplistic thing that's necessary to efficiently implement methods like `integrate`.  Anything beyond that is up to the comiler.

src/test/mod.rs
ryzhyk: Not yet.  I haven't come up with a nice name for the method.  Any suggestions?

mbudiu-vmw: it is doing a zip of the streams and then a map/apply.
ZipMap.

ryzhyk: `zip_with()` maybe?

PR #34, opened 2022-02-08
Format comments.

ryzhyk:
Enable the doc comment feature of `rustfmt`.  These options should be in
`rustfmt.toml`, but that won't be supported until they are stabilized.


mbudiu-vmw (approved):
you can't argue with rust fmt even if you don't like it

Kixiron (approved):

PR #35, opened 2022-02-09
Refactor tests, ZSets and CheckedInts

Kixiron:
* Split out ZSets from their tests (helps compilation times)
* Split `CheckedInt` out into its own file, fixed up some of the traits
* Heavily refactored tests, stopped using string comparisons and redundant `TestTuple`
* Added `finite_map! {}` macro for creating `FiniteHashMap`s
* Removed funky references from various ZSet closures
* Fixed a bunch of clippy lints (how did they slip through?)
* Edited a bunch of overly verbose generics (The fact that they're a type is implicit, they're generics after all)


ryzhyk:
- Re missed lints:  Are you on nightly by any chance?
- I support not using strings for tests.

PR #39, opened 2022-02-15
Several improvements to `FiniteMap`.

ryzhyk:

### DIFF
--- a/src/algebra/finite_map/mod.rs
+++ b/src/algebra/finite_map/mod.rs
@@ -217,7 +217,9 @@ where
             RawEntryMut::Vacant(ve) => {
                 let mut val = Value::zero();
                 f(&mut val);
-                ve.insert(key.clone(), val);
+                if !val.is_zero() {
+                    ve.insert(key.clone(), val);
+                }
             }
         }
     }
@@ -288,9 +290,13 @@ where
     }
 
     fn singleton(key: Key, value: Value) -> Self {
-        let mut result = Self::with_capacity(1);
-        result.value.insert(key, value);
-        result
+        if !value.is_zero() {
+            let mut result = Self::with_capacity(1);
+            result.value.insert(key, value);
+            result
+        } else {
+            Self::new()
+        }
     }
 
     fn map<F, NewKey>(&self, mapper: &F) -> FiniteHashMap<NewKey, Value>
--- a/src/algebra/finite_map/mod.rs
+++ b/src/algebra/finite_map/mod.rs
@@ -6,8 +6,11 @@ mod tests;
 use crate::algebra::{
     Add, AddAssign, AddAssignByRef, AddByRef, GroupValue, HasZero, Neg, NegByRef,
 };
+use hashbrown::{
+    hash_map,
+    hash_map::{Entry, HashMap, RawEntryMut},
+};
 use std::{
-    collections::{hash_map, hash_map::Entry, HashMap},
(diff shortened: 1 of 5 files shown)


src/algebra/finite_map/mod.rs
```diff
@@ -19,8 +22,14 @@ pub trait KeyProperties: 'static + Clone + Eq + Hash {}
 
 impl<T> KeyProperties for T where T: 'static + Clone + Eq + Hash {}
 
-// FIXME: this is not abstract enough.
-type SupportIter<'a, KeyType, ValueType> = hash_map::Keys<'a, KeyType, ValueType>;
+pub trait WithSupport<'a, KeyType>
```
mbudiu-vmw, line 25: a comment would be nice.

src/algebra/finite_map/mod.rs
```diff
@@ -39,16 +48,32 @@ where
     /// Find the value associated to the specified key
     fn lookup(&self, key: &Key) -> Value;
 
-    /// Return the set of values that are mapped to non-zero values.
-    fn support(&self) -> SupportIter<'_, Key, Value>;
+    /// Find the value associated to the specified key
+    ///
+    /// Return value `None` indicates that the key is not stored in the map,
+    /// and its associated value is `Value::zero()`.
+    fn get(&self, key: &Key) -> Option<&Value>;
```
mbudiu-vmw, line 55: but this is wrong. FiniteMaps always map to a group, and if a value is not there it is effectively mapped to 0.

mbudiu-vmw: FiniteMaps are total functions, so they always return a value.

src/algebra/finite_map/mod.rs
```diff
@@ -178,8 +195,48 @@ where
         self.value.get(key).cloned().unwrap_or_else(Value::zero)
     }
 
-    fn support<'a>(&self) -> SupportIter<'_, Key, Value> {
-        self.value.keys()
+    fn get(&self, key: &Key) -> Option<&Value> {
```
mbudiu-vmw, line 198: see above

src/algebra/finite_map/mod.rs
mbudiu-vmw: you can have a private helper function, but this should not be part of the trait.

src/algebra/finite_map/mod.rs
```diff
@@ -178,8 +195,48 @@ where
...
+                let val = oe.get_mut();
+                f(val);
+                if val.is_zero() {
+                    oe.remove();
+                }
+            }
+            RawEntryMut::Vacant(ve) => {
+                let mut val = Value::zero();
+                f(&mut val);
+                ve.insert(key.clone(), val);
```
mbudiu-vmw, line 217: we should not store val if it is zero

mbudiu-vmw (changes requested):

src/algebra/finite_map/mod.rs
ryzhyk: Renaming to `get_in_support`.

src/algebra/finite_map/mod.rs
ryzhyk: Right!  I am also fixing a similar issue in `ZSetHashMap::singleton()`.

mbudiu-vmw (approved):

PR #40, opened 2022-02-15
Index operator.

ryzhyk:
The `Index` operator generates an indexed representation of a Z-set.

The input of the operator is a Z-set where the value type is
a key/value pair.  The output is an indexed representation of
the Z-set, i.e., a finite map from keys to Z-sets of
values associated with each key.

The input Z-set can be represented by any type that can be
converted into a sequence of `((key, value), weight)` pairs.

### DIFF
--- a/src/operator/index.rs
+++ b/src/operator/index.rs
@@ -1,0 +1,209 @@
+use crate::{
+    algebra::{finite_map::KeyProperties, FiniteMap, ZRingValue, ZSet},
+    circuit::{
+        operator_traits::{Operator, UnaryOperator},
+        Circuit, Scope, Stream,
+    },
+};
+use std::{borrow::Cow, marker::PhantomData};
+
+/// Trait for types that can be converted into a pair of references
+///
+/// This trait unifies `(&K, &V)` and `&(K, V)` types, so that the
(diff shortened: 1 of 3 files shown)


src/operator/index.rs
```diff
@@ -0,0 +1,209 @@
...
+        for _ in 0..2 {
+            root.step().unwrap();
+        }
+    }
+
+    #[test]
+    fn index_zset() {
+        let root = Root::build(move |circuit| {
+            let mut inputs = vec![
+                finite_map!{
```
mbudiu-vmw, line 179: how about a test where you really partition the zset using some key function?

mbudiu-vmw (changes requested):

src/operator/index.rs
ryzhyk: Yes, would be nice, but we'll need a `Map` operator first.  I'll add a couple more operators (Map, Join) and write more tests using them.

mbudiu-vmw: Isn't Apply a map?

ryzhyk: Yes, but not a very good one.  I think we can do better performance-wise.

mbudiu-vmw (approved):

PR #41, opened 2022-02-17
Join operator.

ryzhyk:
Join operator and some Z-set API refactoring.  See individual commits.


src/algebra/finite_map/map_macro.rs
```diff
@@ -8,6 +8,7 @@ macro_rules! finite_map {
 
     // Create a map from elements
     ($($key:expr => $value:expr),+ $(,)?) => {{
+        use $crate::algebra::MapBuilder;
```
Kixiron, line 11: Github won't let me comment on it, but I'd instead change this macro arm to have
```rust
$( <$crate::algebra::FiniteHashMap<_, _> as $crate::algebra::MapBuilder>::increment(&mut map, &$key, $value); )+
```
instead of
```rust
$( $crate::algebra::FiniteHashMap::increment(&mut map, &$key, $value); )+
```

src/operator/join.rs
```diff
@@ -0,0 +1,223 @@
...
+    for<'a> &'a I1: IntoIterator<Item = (&'a K, &'a Z1)>,
+    for<'a> &'a Z1: IntoIterator<Item = (&'a V1, &'a W)>,
+    Z1: 'static,
+    I2: FiniteMap<K, Z2>,
+    for<'a> &'a I2: IntoIterator<Item = (&'a K, &'a Z2)>,
+    for<'a> &'a Z2: IntoIterator<Item = (&'a V2, &'a W)>,
+    Z2: 'static,
+    F: Fn(&K, &V1, &V2) -> V + 'static,
+    V: 'static,
+    Z: MapBuilder<V, W> + 'static,
```
Kixiron, line 147: 😅

src/operator/join.rs
```diff
@@ -0,0 +1,223 @@
...
+/// * `V2` - value type in the second input stream.
+/// * `V` - value type in the output Z-set.
+/// * `W` - weight type used by both inputs.
+/// * `F` - join function type: maps key and a pair of values from input Z-sets
+///   to an output value.
+/// * `I1` - indexed Z-set type in the first input stream.
+/// * `I1` - indexed Z-set type in the second input stream.
+/// * `Z1` - value Z-set type in the first input stream.
+/// * `Z2` - value Z-set type in the second input stream.
+/// * `Z` - output Z-set type.
```
Kixiron, line 65: This *screams* "traits with associated types" to me

src/algebra/finite_map/map_macro.rs
ryzhyk: yes, I tried that, and it works, but for some reason this impacts type inference and the compiler asks for type annotations wherever I use the macro

src/operator/join.rs
ryzhyk: I am not a fan of this signature either, but I don't see how to use associated types here without restricting the set of supported input/output types.  Any suggestions?

src/algebra/finite_map/map_macro.rs
ryzhyk: I take it back, it works now.  Maybe something else was broken last time I tried this.

src/operator/join.rs
ryzhyk: On a second thought, I think there's a way. I'll submit a revised version of this shortly.

ryzhyk: Ok, I got rid of two of the type arguments (`Z1` and `Z2`) and am open for suggestions to improve things further.

src/algebra/finite_map/map_macro.rs
Kixiron: Hum, that's very strange

src/operator/join.rs
Kixiron: Maybe make the index type a param on a trait for the input streams? That seems like it should be fairly related, but I'm also not sure how that may impact varying index types (may end up requiring noop wrapper types to change index types or something).

Although on the *inverse*, the index type could make the input stream type a parameter on itself a la
```rust
trait Index
    type Value;
    type Weight;
}
```

ryzhyk: The inverse does not feel right, `IndexedZSet` does not need to know about circuits or streams.  A trait for streams with index as an associated types might work.  I'll try that.

src/operator/join.rs
```diff
@@ -0,0 +1,215 @@
...
+    },
+};
+use std::{borrow::Cow, marker::PhantomData};
+
+impl<P, I1> Stream<Circuit<P>, I1>
+where
+    P: Clone + 'static,
+{
+    /// Apply [`Join`] operator to `self` and `other`.
+    pub fn join<K, V1, V2, V, W, F, I2, Z>(
```
mbudiu-vmw, line 15: I feel dizzy only looking at this list. Should it be documented?

src/operator/join.rs
```diff
@@ -0,0 +1,215 @@
...
+        f: F,
+    ) -> Option<Stream<Circuit<P>, Z>>
+    where
+        K: KeyProperties,
+        V1: KeyProperties,
+        V2: KeyProperties,
+        W: ZRingValue,
+        I1: IndexedZSet<K, V1, W>,
+        for<'a> &'a I1: IntoIterator<Item = (&'a K, &'a I1::ZSet)>,
+        for<'a> &'a I1::ZSet: IntoIterator<Item = (&'a V1, &'a W)>,
```
mbudiu-vmw, line 27: I didn't even know you could do this.

src/operator/join.rs
ryzhyk: Meh, it's doable (in fact either way, as you suggested), but feels like bending backward.  Sure, we can reduce the number of generic arguments, but I'm not sure it gets simpler overall.

src/operator/join.rs
```diff
@@ -0,0 +1,215 @@
...
+                (2, "c h".to_string()) => 12,
+                (2, "d g".to_string()) => 12,
+                (2, "d h".to_string()) => 16,
+                (3, "e i".to_string()) => 25,
+                (3, "e j".to_string()) => -10,
+                (3, "f i".to_string()) => -10,
+                (3, "f j".to_string()) => 4
+            }]
+            .into_iter();
+            let index1: Stream<_, FiniteHashMap<usize, FiniteHashMap<&'static str, isize>>> =
```
mbudiu-vmw, line 192: it's a bummer that you have to specify the type for index1.
How come DD doesn't?

mbudiu-vmw (approved):

src/algebra/finite_map/mod.rs
```diff
@@ -189,6 +225,67 @@ where
...
+            }
+        }
+    }
+
+    fn increment_owned(&mut self, key: Key, value: Value) {
+        if value.is_zero() {
+            return;
+        }
+
+        // TODO: the HashMap API does not support avoiding this clone.
```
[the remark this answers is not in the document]
ryzhyk: oops. Should have removed the comment.

src/operator/join.rs
ryzhyk: It is, in the `Join` operator.

src/operator/join.rs
ryzhyk: Because DD has exactly one type to represent Z-Sets (`Collection`) and one type to represent indexed Z-sets (`Arrangement`). We can have a convenience API on top of this that avoids type annotations by using fixed types.

src/operator/join.rs
Kixiron: I'd add a "See the `Join` operator for more info" note here just so it's a wee more discoverable

ryzhyk: Done

src/operator/join.rs
```diff
@@ -0,0 +1,217 @@
...
+    for (k, vals1) in i1.into_iter() {
+        if let Some(vals2) = i2.get_in_support(k) {
+            for (v1, w1) in vals1.into_iter() {
+                for (v2, w2) in vals2.into_iter() {
+                    map.increment_owned(join_func(k, v1, v2), w1.mul_by_ref(w2));
+                }
+            }
+        }
+    }
+    map
```
Kixiron, line 123:
```suggestion
    }

    map
```

src/operator/join.rs
```diff
@@ -0,0 +1,217 @@
...
+    Z: 'static,
+{
+    fn name(&self) -> Cow<'static, str> {
+        Cow::from("Join")
+    }
+    fn clock_start(&mut self, _scope: Scope) {}
+    fn clock_end(&mut self, _scope: Scope) {}
+}
+
+fn join_inner<K, V1, V2, V, W, F, I1, I2, Z>(i1: &I1, i2: &I2, join_func: &F) -> Z
```
Kixiron, line 99: Small nit, we probably want to join *into* a pre-existing output index instead of creating our own

Kixiron: I'm talking about allowing reusing physical allocations, just making `&mut Z` a parameter on the function

ryzhyk: I can refactor the function that way, but we wouldn't be able to take advantage of this in our current setup.  Where would a Z-set to reuse come from? The set produced at the previous iteration has been consumed by downstream operators...

PR #42, opened 2022-02-18
Incremental join

ryzhyk:
Incremental join operator and a bunch of support infrastructure for it.  See individual commits.

### DIFF
--- a/src/algebra/mod.rs
+++ b/src/algebra/mod.rs
@@ -50,6 +50,15 @@ where
     }
 }
 
+/// A trait for objects with size.
+pub trait WithNumEntries {
+    /// Returns the number of entries in `self`.
+    ///
+    /// Container types have a well-defined notion of size
+    /// (the number of entries in the container).  Scalars have size 1.
+    fn num_entries(&self) -> usize;
+}
+
 /// Like the Add trait, but with arguments by reference.
 pub trait AddByRef {
     fn add_by_ref(&self, other: &Self) -> Self;
--- a/src/algebra/mod.rs
+++ b/src/algebra/mod.rs
@@ -123,13 +123,13 @@ where
 /// We trust the implementation to have an associative addition.
 /// (this cannot be checked statically).
 pub trait MonoidValue:
-    Clone + Eq + 'static + HasZero + Add + AddByRef + AddAssign + AddAssignByRef
+    Clone + Eq + 'static + HasZero + Add<Output = Self> + AddByRef + AddAssign + AddAssignByRef
 {
 }
 
 /// Default implementation for all types that have an addition and a zero.
 impl<T> MonoidValue for T where
-    T: Clone + Eq + 'static + HasZero + Add + AddByRef + AddAssign + AddAssignByRef
+    T: Clone + Eq + 'static + HasZero + Add<Output = Self> + AddByRef + AddAssign + AddAssignByRef
 {
 }
 
@@ -144,7 +144,7 @@ impl<T> GroupValue for T where
         + Eq
         + 'static
         + HasZero
-        + Add
+        + Add<Output = Self>
         + AddByRef
         + AddAssign
         + AddAssignByRef
@@ -163,7 +163,7 @@ impl<T> RingValue for T where
         + Eq
         + 'static
         + HasZero
-        + Add
+        + Add<Output = Self>
         + AddByRef
         + AddAssign
         + AddAssignByRef
@@ -188,7 +188,7 @@ where
         + Eq
         + 'static
         + HasZero
-        + Add
+        + Add<Output = Self>
         + AddByRef
         + AddAssign
         + AddAssignByRef
--- a/src/circuit/circuit_builder.rs
+++ b/src/circuit/circuit_builder.rs
@@ -166,6 +166,15 @@ where
         }
     }
 
+    /// Get a reference to the content of the stream without
+    /// decrementing the refcount.
+    pub(crate) unsafe fn peek(&self) -> &D {
+        let (val, refs) = &*self.val.get();
+        debug_assert_ne!(*refs, 0);
+
+        val.as_ref().unwrap()
+    }
+
     /// Puts a value in the stream, overwriting the previous value if any.
     ///
     /// #Safety
@@ -1052,7 +1061,7 @@ where
         operator: Op,
         input_stream1: &Stream<Self, I1>,
         input_stream2: &Stream<Self, I2>,
-    ) -> Option<Stream<Self, O>>
+    ) -> Stream<Self, O>
     where
         I1: Data,
         I2: Data,
@@ -1079,17 +1088,14 @@ where
         input_stream2: &Stream<Self, I2>,
         input_preference1: OwnershipPreference,
         input_preference2: OwnershipPreference,
-    ) -> Option<Stream<Self, O>>
+    ) -> Stream<Self, O>
     where
         I1: Data,
         I2: Data,
         O: Data,
         Op: BinaryOperator<I1, I2, O>,
     {
-        if input_stream1.origin_node_id() == input_stream2.origin_node_id() {
-            return None;
-        }
-        Some(self.add_node(|id| {
+        self.add_node(|id| {
             self.log_circuit_event(&CircuitEvent::operator(
                 GlobalNodeId::child_of(self, id),
                 operator.name(),
@@ -1106,7 +1112,7 @@ where
             self.connect_stream(input_stream1, id, input_preference1);
             self.connect_stream(input_stream2, id, input_preference2);
             (node, output_stream)
-        }))
+        })
     }
 
     /// Add a N-ary operator (see [`NaryOperator`]).
@@ -1204,13 +1210,11 @@ where
     /// // is a placeholder where we can later plug the input to `z1`.
     /// let (z1_output, z1_feedback) = circuit.add_feedback(Z1::new(0));
     /// // Connect outputs of `source` and `z1` to the plus operator.
-    /// let plus = circuit
-    ///     .add_binary_operator(
-    ///         Apply2::new(|n1: &usize, n2: &usize| n1 + n2),
-    ///         &source,
-    ///         &z1_output,
-    ///     )
-    ///     .unwrap();
+    /// let plus = circuit.add_binary_operator(
+    ///     Apply2::new(|n1: &usize, n2: &usize| n1 + n2),
+    ///     &source,
+    ///     &z1_output,
+    /// );
     /// // Connect the output of `+` as input to `z1`.
     /// z1_feedback.connect(&plus);
     /// # });
@@ -1370,13 +1374,11 @@ where
     ///                 res
     ///             });
     ///             let (z1_output, z1_feedback) = child.add_feedback_with_export(Z1::new(1));
-    ///             let mul = child
-    ///                 .add_binary_operator(
-    ///                     Apply2::new(|n1: &usize, n2: &usize| n1 * n2),
-    ///                     &countdown,
-    ///                     &z1_output.local,
-    ///                 )
-    ///                 .unwrap();
+    ///             let mul = child.add_binary_operator(
+    ///                 Apply2::new(|n1: &usize, n2: &usize| n1 * n2),
+    ///                 &countdown,
+    ///                 &z1_output.local,
+    ///             );
     ///             z1_feedback.connect(&mul);
     ///             Ok((move || *counter.borrow() <= 1, z1_output.export))
     ///         })
@@ -1727,6 +1729,8 @@ struct BinaryNode<C, I1, I2, O, Op> {
     input_stream1: Stream<C, I1>,
     input_stream2: Stream<C, I2>,
     output_stream: Stream<C, O>,
+    // `true` if both input streams are aliases of the same stream.
+    is_alias: bool,
 }
 
 impl<P, I1, I2, O, Op> BinaryNode<Circuit<P>, I1, I2, O, Op>
@@ -1741,11 +1745,13 @@ where
         circuit: Circuit<P>,
         id: NodeId,
     ) -> Self {
+        let is_alias = input_stream1.origin_node_id() == input_stream2.origin_node_id();
         Self {
             id,
             operator,
             input_stream1,
             input_stream2,
+            is_alias,
             output_stream: Stream::new(circuit, id),
         }
     }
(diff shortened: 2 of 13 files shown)


src/algebra/mod.rs
```diff
@@ -50,6 +50,15 @@ where
     }
 }
 
+/// A trait for objects with size.
```
mbudiu-vmw, line 53: this is too close to "size in bytes"
I would prefer WithCapacity.

src/algebra/mod.rs
```diff
@@ -114,13 +123,13 @@ where
 /// We trust the implementation to have an associative addition.
 /// (this cannot be checked statically).
 pub trait MonoidValue:
-    Clone + Eq + 'static + HasZero + Add + AddByRef + AddAssign + AddAssignByRef
+    Clone + Eq + 'static + HasZero + Add<Output = Self> + AddByRef + AddAssign + AddAssignByRef
```
mbudiu-vmw, line 126: Isn't this default?

src/operator/integrate.rs
```diff
@@ -5,6 +5,46 @@ use crate::{
...
+///
+/// * `current` - the current value of the integral, i.e., the sum of all values
+///   in the stream since the last `clock_start` including the current clock
+///   cycle.
+/// * `delayed` - the previous value of the integral, i.e., the sum of all
+///   values in the stream since the last `clock_start` up to the previous clock
+///   cycle.
+/// * `export` - stream exported to the parent circuit that contains the final
+///   value of the integral at the end of the nested clock epoch.
+/// * `delta` - value added to the integral at the last clock cycle.  This is
```
mbudiu-vmw, line 21: that's a bit strange.
This could really be done for all operators if you find it convenient.

mbudiu-vmw: For unary operators it should be called "input"; delta is very confusing, it suggests that something has been done to the input.

src/operator/join.rs
```diff
@@ -39,6 +40,40 @@ where
     }
 }
 
+impl<P, I1> StreamIntegral<P, I1>
+where
+    P: Clone + 'static,
+{
+    pub fn join_incremental<K, V1, V2, V, W, F, I2, Z>(
```
mbudiu-vmw, line 47: I find the placement of this method here a bit strange.
I think more natural would have to have a static method (no Self) in some Join operator that takes two StreamIntegral arguments. People will not find it here.

src/operator/sum.rs
```diff
@@ -0,0 +1,201 @@
...
+        for input in inputs {
+            input_vec.push(input);
+        }
+        input_vec.sort_by(|x, y| y.size().cmp(&x.size()));
+
+        if input_vec.is_empty() {
+            return D::zero();
+        }
+        let mut res = D::zero();
+        for input in input_vec.drain(..) {
```
mbudiu-vmw, line 89: So I guess this saves n-1 intermediate results and scheduler invocations.
But on the other hand you sort this vector every time. It's true it has size 3 so far.

src/operator/sum.rs
```diff
@@ -0,0 +1,201 @@
...
+            let mut s = ZSetHashMap::new();
+            let source3 = circuit.add_source(Generator::new(move || {
+                let res = s.clone();
+                s.increment(&6, -1);
+                res
+            }));
+
+            // Supply `source3` twice to test the handling of aliases.
+            source3
+                .sum([source2.clone(), source1.clone(), source3.clone()].iter())
```
mbudiu-vmw, line 146: is the iter() necessary?

mbudiu-vmw (approved):

src/algebra/mod.rs
ryzhyk: It is, but sometimes compiler becomes unsure.

src/operator/integrate.rs
ryzhyk: It's the delta between delayed and current.

mbudiu-vmw: which is the current input, isn't it?

src/operator/sum.rs
ryzhyk: I'll check

src/operator/sum.rs
ryzhyk: This doesn't reduce the number of intermediate results, but hopefully produces them in a more optimal order due to the sorting...

src/operator/integrate.rs
ryzhyk: But yeah, it can be confusing, I'll rename

src/operator/sum.rs
ryzhyk: it is

src/operator/sum.rs
```diff
@@ -0,0 +1,201 @@
...
+    D: Add<Output = D> + AddAssignByRef + Clone + HasZero + WithNumEntries + 'static,
+{
+    fn eval<'a, Iter>(&mut self, inputs: Iter) -> D
+    where
+        Iter: Iterator<Item = Cow<'a, D>>,
+    {
+        let input_vec = unsafe {
+            transmute::<&mut Vec<Cow<'static, D>>, &mut Vec<Cow<'a, D>>>(&mut self.inputs)
+        };
+        input_vec.clear();
```
Kixiron, line 79: You probably want to put this clear call at the end of the function instead of the beginning

src/operator/sum.rs
```diff
@@ -0,0 +1,201 @@
...
+where
+    D: Add<Output = D> + AddAssignByRef + Clone + HasZero + WithNumEntries + 'static,
+{
+    fn eval<'a, Iter>(&mut self, inputs: Iter) -> D
+    where
+        Iter: Iterator<Item = Cow<'a, D>>,
+    {
+        let input_vec = unsafe {
+            transmute::<&mut Vec<Cow<'static, D>>, &mut Vec<Cow<'a, D>>>(&mut self.inputs)
+        };
```
Kixiron, line 78: I'm fairly sure this is UB, you probably want to do something like this:
```suggestion
        let input_vec = unsafe {
            let buffer = ManuallyDrop::new(take(&mut self.inputs));
            let (ptr, len, cap) = (buffer.as_mut_ptr(), buffer.len(), buffer.capacity());

            Vec::from_raw_parts(ptr, len, cap)
        };
```
and then the inverse at the end of the function to replace `self.inputs`

src/operator/sum.rs
```diff
@@ -0,0 +1,201 @@
...
+        input_vec.clear();
+        for input in inputs {
+            input_vec.push(input);
+        }
+        input_vec.sort_by(|x, y| y.num_entries().cmp(&x.num_entries()));
```
Kixiron, line 83: I think you can use `.sort_by_key(|value| value.num_entries())` here

src/operator/sum.rs
```diff
@@ -0,0 +1,201 @@
...
+
+impl<P, D> Stream<Circuit<P>, D>
+where
+    P: Clone + 'static,
+    D: Add<Output = D> + AddByRef + AddAssignByRef + Clone + HasZero + WithNumEntries + 'static,
+{
+    /// Apply the [`Sum`] operator to `self` and all streams in `streams`.
+    pub fn sum<'a, I>(&'a self, streams: I) -> Stream<Circuit<P>, D>
+    where
+        I: Iterator<Item = &'a Self>,
```
Kixiron, line 20: `IntoIterator` is more appropriate here
```suggestion
        I: IntoIterator<Item = &'a Self>,
```

src/operator/sum.rs
ryzhyk: Why is your alternative safer? It still changes the lifetime while the compiler isn't looking.

src/operator/sum.rs
ryzhyk: Trouble is, I need the descending order, so I'd have to reverse the vector or alternatively iterate over the vector backwards.

src/operator/sum.rs
Kixiron: It's how it does it, funky stuff to do with variance. Additionally, my code actually takes ownership of the vector instead of mutably aliasing it while another mutable reference is alive to the struct containing it (meaning `self.inputs` is implicitly borrowed via the borrow on self) which avoids the good ole double mut refs problem

ryzhyk: Ok, thanks, let me try this.

src/operator/sum.rs
ryzhyk: This is really a no-op, since the vector gets drained every time.  I think, I'll just replace it with an assertion that the vector is empty.

src/operator/sum.rs
ryzhyk: Turns out clippy knows a trick:
```
input_vec.sort_by_key(|x| Reverse(x.num_entries()));
```

ryzhyk:
@Kixiron , thanks for the review, I think I fixed everything.

src/circuit/circuit_builder.rs
```diff
@@ -1722,14 +1788,129 @@ where
...
+
+impl<P, I, O, Op> NaryNode<Circuit<P>, I, O, Op>
+where
+    I: Clone + 'static,
+    Op: NaryOperator<I, O>,
+    P: Clone,
+{
+    fn new<Iter>(operator: Op, input_streams: Iter, circuit: Circuit<P>, id: NodeId) -> Self
+    where
+        Iter: Iterator<Item = Stream<Circuit<P>, I>>,
```
Kixiron, line 1849:
```suggestion
        Iter: IntoIterator<Item = Stream<Circuit<P>, I>>,
```

src/circuit/circuit_builder.rs
```diff
@@ -1722,14 +1788,129 @@ where
...
+        for i in 0..input_streams.len() {
+            for j in 0..i {
+                if input_streams[i].0.ptr_eq(&input_streams[j].0) {
+                    input_streams[i].1 = true;
+                    aliases.push(i);
+                    break;
+                }
+            }
+        }
+        Self {
```
Kixiron, line 1862:
```suggestion
        }
        aliases.shrink_to_fit();
        input_streams.shrink_to_fit();

        Self {
```

src/circuit/circuit_builder.rs
```diff
@@ -1722,14 +1788,129 @@ where
...
+                        if *alias {
+                            Cow::Borrowed(stream.peek())
+                        } else {
+                            stream.take()
+                        }
+                    })),
+            );
+        for i in self.aliases.iter() {
+            let _ = self.input_streams[*i].0.take();
+        }
```
Kixiron, line 1913:
```suggestion
            );

        for i in self.aliases.iter() {
            let _ = self.input_streams[*i].0.take();
        }

```

src/operator/sum.rs
```diff
@@ -0,0 +1,215 @@
...
+                    Cow::Borrowed(v) => res.add_assign_by_ref(v),
+                    Cow::Owned(v) => {
+                        res = res + v;
+                    }
+                }
+            }
+        }
+
+        self.inputs = unsafe {
+            let mut buffer = ManuallyDrop::new(take(&mut input_vec));
```
Kixiron, line 114: The take is redundant since we already have ownership
```suggestion
            let mut buffer = ManuallyDrop::new(input_vec);
```

src/operator/sum.rs
```diff
@@ -0,0 +1,215 @@
...
+            let (ptr, len, cap) = (buffer.as_mut_ptr(), buffer.len(), buffer.capacity());
+            Vec::from_raw_parts(ptr as usize as *mut Cow<'static, D>, len, cap)
```
Kixiron, line 116:
```suggestion
            Vec::from_raw_parts(ptr.cast(), len, cap)
```

src/operator/sum.rs
```diff
@@ -0,0 +1,215 @@
...
+    where
+        Iter: Iterator<Item = Cow<'a, D>>,
+    {
+        let mut input_vec: Vec<Cow<'a, _>> = unsafe {
+            let mut buffer = ManuallyDrop::new(take(&mut self.inputs));
+            let (ptr, len, cap) = (buffer.as_mut_ptr(), buffer.len(), buffer.capacity());
+            Vec::from_raw_parts(ptr, len, cap)
+        };
+
+        assert!(input_vec.is_empty());
```
Kixiron, line 88:
```suggestion
        debug_assert!(input_vec.is_empty());
```

src/operator/sum.rs
```diff
@@ -0,0 +1,215 @@
...
+                res = match input {
+                    Cow::Borrowed(v) => v.clone(),
+                    Cow::Owned(v) => v,
+                };
+            } else {
+                match input {
+                    Cow::Borrowed(v) => res.add_assign_by_ref(v),
+                    Cow::Owned(v) => {
+                        res = res + v;
+                    }
```
Kixiron, line 108: We might as well require `AddAssign` instead of `Add`
```suggestion
                    Cow::Owned(v) => res = res + v,
```

src/operator/sum.rs
Kixiron, line 114: The more critical thing is really making sure that we don't let any items with a temporary lifetime escape into our new vec with a static one
```suggestion
        self.inputs = unsafe {
            input_vec.clear();

            let mut buffer = ManuallyDrop::new(take(&mut input_vec));
```

src/circuit/circuit_builder.rs
```diff
@@ -166,6 +170,15 @@ where
         }
     }
 
+    /// Get a reference to the content of the stream without
+    /// decrementing the refcount.
+    pub(crate) unsafe fn peek(&self) -> &D {
+        let (val, refs) = &*self.val.get();
+        debug_assert!(*refs > 0);
```
Kixiron, line 177:
```suggestion
        debug_assert_ne!(*refs, 0);
```

src/circuit/circuit_builder.rs
```diff
@@ -1101,7 +1116,58 @@ where
...
+                input_streams.clone().into_iter(),
+                self.clone(),
+                id,
+            );
+            let output_stream = node.output_stream();
+            for stream in input_streams.iter() {
+                self.connect_stream(stream, id, input_preference);
+            }
+            (node, output_stream)
+        })
```
Kixiron, line 1170: If these methods just immediately clone their inputs they should take owned values instead of references so we can let the caller decide when cloning needs to happen

src/operator/sum.rs
ryzhyk: How's `AddAssign` better?  It's seems less flexible performance-wise, as it doesn't let you pick the side to add to.

src/operator/sum.rs
ryzhyk: I'll add another `assert`.

src/operator/sum.rs
Kixiron: That's fair I guess

PR #43, opened 2022-02-23
Operators over streams: Z1, integrate.

ryzhyk:
Versions of z^-1 and integrator that work on streams rather than
individual values.  Mathematically, these are identical to regular `Z1`
and `integrate` operators, but implementation-wise these operators
consume the inputs value-by-value as the (conceptually infinite) input
stream gets generated.

This commit also introduces an adapter mechanism that makes it easy to
wrap values in streams in shared references (`Rc`, `Arc`, `Intern`,
etc).  Adapter operators encapsulate regular operators, which do not need
to know about such references, and take care of unwrapping inputs and
wrapping outputs.  This mechanism allows shared values to be stored
(read-only) in multiple operators without memory bloat.  We introduce
this capability now since it will be useful in  combination with the new
`integrate_nested` operator.  Consider the following circuit, that will
arise in the implementation of nested incremental join:

```
       integrate_nested
      ┌───────────────┐
      │               │
      │     ┌───┐     │
      │     │   │     │
──────┼─────┤ + ├───┬─┼─ Z ─────►
      │     │   │   │ │
      │     └───┘   │ │
      │       ▲     │ │
      │       │     │ │
      │ ZNested ◄───┘ │
      │               │
      └───────────────┘
```

Here the `ZNested` that is part of the nested integrator stores a vector
of Z-sets, one per nested timestamp.  The "regular" `Z` operator
(\lift(Z)) contains a copy of the second-last element of the vector.

### DIFF
--- a/src/operator/adapter.rs
+++ b/src/operator/adapter.rs
@@ -1,0 +1,177 @@
+//! Adapter operators wrap around regular operators and decapsulate
+//! inputs wrapped in shared references.
+//!
+//! Adapters in this module allow existing operators such as `Plus<T>`
+//! to be applied to streams that carry data of type `Ref<T>`, where
+//! `Ref` is any type that implements trait [`SharedRef`](`crate::SharedRef`).
+//! The adapter tries to extract an owned value from `Ref` if possible
+//! and borrows the value otherwise before invoking the inner operator.
+//! We provide these generic adapters so we do not need to generalize the
+//! implementation of individual operators to handle shared references.
+//!
+//! For good measure, the adapter also optionally wraps the output of
+//! the inner operator in `Ref` or converts it to any other type that
+//! implements `From<T>`.  This isn't strictly necessary, as the same
+//! can be achieved by chaining the operator with a simple transformer
+//! operator.
+
+use std::{borrow::Cow, marker::PhantomData};
+
+use crate::{
+    circuit::{
+        operator_traits::{BinaryOperator, Operator, UnaryOperator},
+        OwnershipPreference, Scope,
+    },
+    SharedRef,
+};
+
+/// Unary operator adapter unwraps input values of type
+/// `I` wrapped in a shared reference.  See
+/// [module-level documentation](`crate::operator::adapter`) for details.
+pub struct UnaryOperatorAdapter<I, O, Op> {
+    op: Op,
+    _types: PhantomData<(I, O)>,
+}
+
+impl<I, O, Op> Operator for UnaryOperatorAdapter<I, O, Op>
+where
+    Op: Operator,
+    I: 'static,
+    O: 'static,
+{
+    fn name(&self) -> Cow<'static, str> {
+        self.op.name()
+    }
+    fn clock_start(&mut self, scope: Scope) {
+        self.op.clock_start(scope);
+    }
+    fn clock_end(&mut self, scope: Scope) {
+        self.op.clock_end(scope);
+    }
+    fn is_async(&self) -> bool {
+        self.op.is_async()
+    }
+    fn ready(&self) -> bool {
+        self.op.ready()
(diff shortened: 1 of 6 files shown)


src/shared_ref.rs
```diff
@@ -0,0 +1,31 @@
+//! Shared reference trait.
+
+use std::{borrow::Borrow, rc::Rc};
+
+/// A trait that generalizes shared pointers like `Rc` and `Arc`.
+///
+/// The holder of a shared reference can extract an owned copy
+/// of the reference object from it if there are not other references
```
mbudiu-vmw, line 8: no other

src/shared_ref.rs
```diff
@@ -0,0 +1,31 @@
...
+    /// If `self` is the last reference to the object, returns
+    /// the object; otherwise, returns `Err(self)`.
+    fn try_into_owned(self) -> Result<T, Self>;
+}
+
+impl<T> SharedRef<T> for T
+where
+    T: Clone,
+{
+    fn try_into_owned(self) -> Result<T, Self> {
```
mbudiu-vmw, line 22: I find it a little distasteful that the Err constructor is used for something which isn't really an error.

src/operator/adapter.rs
```diff
@@ -0,0 +1,177 @@
...
+where
+    Op: BinaryOperator<I1, I2, O>,
+    RI1: SharedRef<I1>,
+    RI2: SharedRef<I2>,
+    I1: 'static,
+    I2: 'static,
+    RO: From<O>,
+    O: 'static,
+{
+    fn eval(&mut self, lhs: &RI1, rhs: &RI2) -> RO {
```
mbudiu-vmw, line 147: usually lhs is left-hand side, I think left and right are more appropriate, but it's not a big deal.

src/operator/integrate.rs
```diff
@@ -186,4 +270,65 @@ mod test {
             root.step().unwrap();
         }
     }
+
+    #[rustfmt::skip]
```
mbudiu-vmw, line 274: this is not a real plus, given the definition in your program below.
it is a plus with state

src/operator/z1.rs
```diff
@@ -110,47 +110,211 @@ where
...
+}
+
+impl<T> StrictOperator<T> for Z1Nested<T>
+where
+    T: Clone + 'static,
+{
+    fn get_output(&mut self) -> T {
+        if self.timestamp >= self.val.len() {
+            assert_eq!(self.timestamp, self.val.len());
+            self.val.push(self.zero.clone());
```
mbudiu-vmw, line 191: If this is sparse there may be a better way to represent it.

mbudiu-vmw (approved):

src/shared_ref.rs
ryzhyk: This mimics the behavior of `Rc::try_unwrap` and `Arc::try_unwrap`

src/operator/integrate.rs
ryzhyk: It's the exact same plus actually.  Why do you say it has state?

src/operator/z1.rs
ryzhyk: Let's not optimize prematurely.  My current assumption is there won't be many nested iterations (at least not enough to worry about the memory used by zeros).

src/operator/integrate.rs
mbudiu-vmw: so there is a z^-1 and a special max(a-1, 0) operator which I didn't notice, which you denoted by "-1".

mbudiu-vmw: there is no box around the z

ryzhyk:
Merged with #44

PR #44, opened 2022-02-24
 More robust reference counting for streams. 

ryzhyk:
Only the last commit is new; the other one is PR #43

### DIFF
--- a/src/circuit/circuit_builder.rs
+++ b/src/circuit/circuit_builder.rs
@@ -55,6 +55,33 @@ use crate::circuit::{
     trace::{CircuitEvent, SchedulerEvent},
 };
 
+/// Value stored in the stream.
+struct StreamValue<D> {
+    /// Value written to the stream at the current clock cycle;
+    /// `None` after the last consumer has retrieved the value from the stream.
+    val: Option<D>,
+    /// The number of consumers connected to the stream.  Each consumer
+    /// reads from the stream exactly once at every clock cycle.
+    consumers: usize,
+    /// The number of remaining consumers still expected to read from the stream
+    /// at the current clock cycle.  This value is reset to `consumers` when
+    /// a new value is written to the stream.  It is decremented on each access.
+    /// The last consumer to read from the stream (`tokens` drops to 0) obtains
+    /// an owned value rather than a borrow.  See description of
+    /// [ownership-aware scheduling](`OwnershipPreference`) for details.
+    tokens: usize,
+}
+
+impl<D> StreamValue<D> {
+    const fn empty() -> Self {
+        Self {
+            val: None,
+            consumers: 0,
+            tokens: 0,
+        }
+    }
+}
+
 /// A stream stores the output of an operator.  Circuits are synchronous,
 /// meaning that each value is produced and consumed in the same clock cycle, so
 /// there can be at most one value in the stream at any time.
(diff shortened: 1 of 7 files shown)


mbudiu-vmw (approved):

src/operator/z1.rs
```diff
@@ -110,47 +110,211 @@ where
...
+        Self {
+            zero,
+            timestamp: 0,
+            val: vec![],
+        }
+    }
+
+    fn reset(&mut self) {
+        self.timestamp = 0;
+        self.val = vec![];
```
Kixiron, line 158: Keep this buffer allocated
```suggestion
        self.val.clear();
```

src/operator/z1.rs
```diff
@@ -110,47 +110,211 @@ where
...
+
+impl<T> Z1Nested<T>
+where
+    T: Clone,
+{
+    pub fn new(zero: T) -> Self {
+        Self {
+            zero,
+            timestamp: 0,
+            val: vec![],
```
Kixiron, line 152:
```suggestion
            val: Vec::new(),
```

src/operator/z1.rs
```diff
@@ -110,47 +110,211 @@ where
...
+
+    fn clock_start(&mut self, scope: Scope) {
+        if scope == 0 {
+            self.val.truncate(self.timestamp);
+        }
+        self.timestamp = 0;
+    }
+
+    fn clock_end(&mut self, scope: Scope) {
+        if scope > 0 {
```
Kixiron, line 178:
```suggestion
        if scope != 0 {
```

src/operator/z1.rs
```diff
@@ -110,47 +110,211 @@ where
...
+{
+    fn get_output(&mut self) -> T {
+        if self.timestamp >= self.val.len() {
+            assert_eq!(self.timestamp, self.val.len());
+            self.val.push(self.zero.clone());
+        }
+
+        let mut zero = self.zero.clone();
+        swap(
+            unsafe { self.val.get_unchecked_mut(self.timestamp) },
```
Kixiron, line 196: Needs safety comment

src/operator/integrate.rs
```diff
@@ -186,4 +270,65 @@ mod test {
             root.step().unwrap();
         }
     }
+
+    #[rustfmt::skip]
```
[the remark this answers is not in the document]
ryzhyk: It's actually for the comment.

src/operator/z1.rs
ryzhyk: This is reset for the outer timestamp, basically the whole parent circuit starting from scratch, so there's no reason to expect that we'll need similar amount of inner iterations.  In any case, savings are negligible across many parent clock ticks.

src/operator/integrate.rs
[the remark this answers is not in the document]
Kixiron: What do you mean, what does rustfmt do to it? Have you tried wrapping it in a code block instead?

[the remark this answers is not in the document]
ryzhyk: Good point, I was not thinking.

src/operator/z1.rs
Kixiron: I mean, there's no reason not to keep the memory allocated, it's only a benefit really

src/circuit/circuit_builder.rs
```diff
@@ -55,6 +55,39 @@ use crate::circuit::{
...
 };
 
+/// Value stored in the stream.
+struct StreamValue<D> {
+    /// Value written to the stream at the current clock cycle;
+    /// `None` after the last consumer has retrieved the value from the stream.
+    val: Option<D>,
+    /// The number of consumers connected to the stream.  Each consumer
+    /// reads from the stream exactly once at every clock cycle.
+    readers: usize,
```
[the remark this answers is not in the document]
ryzhyk: `readers` -> `consumers` is nice. I think `tokens` is adequate though.

PR #45, opened 2022-03-01
Stream caching.

ryzhyk:
We use a form of caching to reuse previously constructed fragments of
the circuit. For instance, a `stream.integrate()` call creates a
feedback loop with an adder and a `z^-1` operator.  The output of this
circuit is cached, so that the next identical call will return the same
stream without creating a new copy of the same operators.  Stream
integral is one example of a derived stream.  Others include nested
integral, delayed stream, etc.  Caching will allow simplifying
high-level APIs like nested incremental join that combine several
derived streams.  Without caching, such methods would have to either
construct all derived streams internally, potentially creating duplicate
operators and thus wasting CPU and memory or take multiple derived
streams as input, which would require the caller to track these
derivatives manually.

### DIFF
--- a/src/circuit/cache.rs
+++ b/src/circuit/cache.rs
@@ -1,0 +1,90 @@
+//! Circuit cache.
+//!
+//! # Background
+//!
+//! We use a form of caching to reuse previously constructed fragments of the
+//! circuit. For instance, a
+//! [`stream.integrate()`](`crate::circuit::Stream::integrate`) call creates a
+//! feedback loop with an adder and a `z^-1` operator.  The output of this
+//! circuit is cached, so that the next identical call will return the same
+//! stream without creating a new copy of the same operators.  Stream integral
+//! is one example of a derived stream.  Others include
+//! [nested integral](`crate::circuit::Stream::integrate_nested`), [delayed
+//! stream](`crate::circuit::Stream::delay`), etc.  Caching allows simplifying
+//! high-level APIs like nested incremental join that combine several derived
+//! streams.  Without caching, such methods would have to either construct all
+//! derived streams internally, potentially creating duplicate operators and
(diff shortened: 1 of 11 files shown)


src/circuit/cache.rs
```diff
@@ -0,0 +1,90 @@
+//! Circuit cache.
+//!
+//! # Background
+//!
+//! We use a form of caching to reuse previously constructed fragments of the
```
mbudiu-vmw, line 5: to discover and reuse circuits that compute equivalent streams.

src/circuit/cache.rs
```diff
@@ -0,0 +1,90 @@
...
+//! circuit. For instance, a
+//! [`stream.integrate()`](`crate::circuit::Stream::integrate`) call creates a
```
mbudiu-vmw, line 7: I think you can be even more general: applying equivalent operators to the same stream produces equivalent streams. In particular, the z inside an integrator is an example.

src/circuit/cache.rs
```diff
@@ -0,0 +1,90 @@
...
+/// # Example
+///
+/// ```ignore
+/// circuit_cache_key!(IntegralId<C, D>(NodeId => Stream<C, D>));
+/// ```
+///
+/// declares `struct IntegralId(NodeId)` key type with associated value type
+/// `Stream<C, D>`.
+#[macro_export]
+macro_rules! circuit_cache_key {
```
mbudiu-vmw, line 59: Is this necessary?
If you have an Eq and Hash function on Operators wouldn't a normal hashmap just work?
You just need to keep a hashmap for each stream.

mbudiu-vmw: For binary commutative operations you need to be a bit more sophisticated, by choosing an ordering of the inputs.

PR #46, opened 2022-03-02
Stream caching.

ryzhyk:
We use a form of caching to reuse previously constructed fragments of
the circuit. For instance, a `stream.integrate()` call creates a
feedback loop with an adder and a `z^-1` operator.  The output of this
circuit is cached, so that the next identical call will return the same
stream without creating a new copy of the same operators.  Stream
integral is one example of a derived stream.  Others include nested
integral, delayed stream, etc.  Caching will allow simplifying
high-level APIs like nested incremental join that combine several
derived streams.  Without caching, such methods would have to either
construct all derived streams internally, potentially creating duplicate
operators and thus wasting CPU and memory or take multiple derived
streams as input, which would require the caller to track these
derivatives manually.

### DIFF
--- a/src/circuit/cache.rs
+++ b/src/circuit/cache.rs
@@ -1,0 +1,91 @@
+//! Circuit cache.
+//!
+//! # Background
+//!
+//! We use a form of caching to reuse previously constructed fragments of the
+//! circuit. For instance, a
+//! [`stream.integrate()`](`crate::circuit::Stream::integrate`) call creates a
+//! feedback loop with an adder and a `z^-1` operator.  The output of this
+//! circuit is cached, so that the next identical call will return the same
+//! stream without creating a new copy of the same operators.  Stream integral
+//! is one example of a derived stream.  Others include
+//! [nested integral](`crate::circuit::Stream::integrate_nested`), [delayed
+//! stream](`crate::circuit::Stream::delay`), etc.  Caching allows simplifying
+//! high-level APIs like nested incremental join that combine several derived
+//! streams.  Without caching, such methods would have to either construct all
+//! derived streams internally, potentially creating duplicate operators and
+//! thus wasting CPU and memory or take multiple derived streams as input, which
+//! would require the caller to track these derivatives manually.
+//!
+//! # API
+//!
+//! This module provides a low-level mechanism that individual operators use to
+//! perform operator-specific caching.  The mechanism consists of a key-value
+//! store associated with each circuit.  We use `TypedMap`, which supports
+//! multiple key and value types, for the store.  An operator registers a new
+//! key type and associated value type using the [`circuit_cache_key`] macro,
+//! e.g.,
+//!
+//! ```ignore
+//! circuit_cache_key!(IntegralId<C, D>(NodeId => Stream<C, D>));
+//! ```
+//!
+//! declares a mapping from `NodeId` to `Stream` used by the
+//! [`Stream::integrate`](`crate::circuit::Stream::integrate`) method to cache
+//! the output of the integrator.  Internally, the method first performs a map
+//! lookup and returns the cached stream, if one exists; otherwise it
+//! instantiates the integration circuit and stores its output stream in the
+//! cache before returning it to the caller.
+
+use typedmap::TypedMap;
+
+pub struct CircuitStoreMarker;
+
+/// Per-circuit cache.
+pub type CircuitCache = TypedMap<CircuitStoreMarker>;
+
+/// Declare an anonymous struct type to be used as a key in the cache and
+/// associated value type.
+///
+/// # Example
+///
+/// ```ignore
+/// circuit_cache_key!(IntegralId<C, D>(NodeId => Stream<C, D>));
+/// ```
+///
+/// declares `struct IntegralId(NodeId)` key type with associated value type
+/// `Stream<C, D>`.
+#[macro_export]
+macro_rules! circuit_cache_key {
+    ($constructor:ident<$($typearg:ident),*>($key_type:ty => $val_type:ty)) => {
+        #[repr(transparent)]
+        pub struct $constructor<$($typearg: 'static),*>(pub $key_type, std::marker::PhantomData<($($typearg),*)>);
+
+        impl<$($typearg),*> $constructor<$($typearg),*> {
+            pub fn new(key: $key_type) -> Self {
+                Self(key, std::marker::PhantomData)
+            }
+        }
+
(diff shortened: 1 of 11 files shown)


src/circuit/cache.rs
```diff
@@ -0,0 +1,90 @@
...
+//!
+//! ```ignore
+//! circuit_cache_key!(IntegralId<C, D>(NodeId => Stream<C, D>));
+//! ```
+//!
+//! declares a mapping from `NodeId` to `Stream` used by the
+//! [`Stream::integrate`](`crate::circuit::Stream::integrate`) method to cache
+//! the output of the integrator.  Internally, the method first performs a map
+//! lookup and returns the cached stream, if one exists; otherwise it
+//! instantiates the integration circuit and stores its output stream in the
```
mbudiu-vmw, line 37: only the output stream or also all internal streams

src/operator/z1.rs
```diff
@@ -1,11 +1,159 @@
...
+        let Self {
+            feedback,
+            output,
+            export,
+        } = self;
+        let circuit = output.circuit().clone();
+
+        feedback.connect_with_preference(input, OwnershipPreference::STRONGLY_PREFER_OWNED);
+        circuit
+            .cache()
```
mbudiu-vmw, line 61: you always write cache().insert(...), perhaps this should be renamed to just cache(...), or memoize(...).

src/operator/z1.rs
```diff
@@ -1,11 +1,159 @@
...
+        delayed
+    }
+
+    /// Applies [`Z1Nested`] operator to `self`.
+    pub fn nested_delay(&self) -> Stream<Circuit<P>, D>
+    where
+        P: Clone + 'static,
+        D: Clone + HasZero + 'static,
+    {
+        if let Some(delayed) = self
```
mbudiu-vmw, line 138: also this pattern is very common: if (lookup) return; create; cache; return.
But there is common code between lookup and cache; return. 
I wonder whether that can be factored using lambdas similar to the computeIfAbsent Java API https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#computeIfAbsent-K-java.util.function.Function-

mbudiu-vmw (approved):

src/circuit/cache.rs
```diff
@@ -0,0 +1,90 @@
...
+/// ```ignore
+/// circuit_cache_key!(IntegralId<C, D>(NodeId => Stream<C, D>));
+/// ```
+///
+/// declares `struct IntegralId(NodeId)` key type with associated value type
+/// `Stream<C, D>`.
+#[macro_export]
+macro_rules! circuit_cache_key {
+    ($constructor:ident<$($typearg:ident),*>($key_type:ty => $val_type:ty)) => {
+        pub struct $constructor<$($typearg: 'static),*>(pub $key_type, std::marker::PhantomData<($($typearg),*)>);
```
Kixiron, line 61:
```suggestion
        #[repr(transparent)]
        pub struct $constructor<$($typearg: 'static),*>(pub $key_type, std::marker::PhantomData<($($typearg),*)>);
```

src/circuit/circuit_builder.rs
```diff
@@ -155,6 +160,29 @@ impl<P, D> Stream<Circuit<P>, D> {
...
+        P: Clone + 'static,
+        D: 'static,
+    {
+        if let Some(export) = self
+            .circuit()
+            .cache()
+            .get(&ExportId::new(self.local_node_id()))
+        {
+            return export.clone();
+        };
```
Kixiron, line 182:
```suggestion
        }
```

src/operator/integrate.rs
```diff
@@ -308,17 +254,20 @@ mod test {
...
                 .iterate_with_condition(|child| {
                     let source = source.delta0(&child);
-                    let (z, feedback) = child.add_feedback(Z1::new(0));
-                    let plus = source.plus(&z.apply(|n| if *n > 0 { *n - 1 } else { *n }));
+                    let feedback = DelayedFeedback::new(child);
+                    let plus =
+                        source.plus(
+                            &feedback
+                                .stream()
+                                .apply(|n| if *n > 0 { *n - 1 } else { *n }),
```
Kixiron, line 262:
```suggestion
                                .apply(|&n| if n > 0 { n - 1 } else { n }),
```

src/operator/z1.rs
```diff
@@ -181,6 +329,42 @@ where
     }
 }
 
+impl<T> UnaryOperator<T, T> for Z1Nested<T>
+where
+    T: Clone + 'static,
+{
+    fn eval(&mut self, i: &T) -> T {
+        assert!(self.timestamp <= self.val.len());
```
Kixiron, line 337: These asserts could probably be `debug_assert!()`s since it's not required for safety and it's more of a "keep us from writing buggy code" kind of thing

src/operator/z1.rs
```diff
@@ -1,11 +1,159 @@
...
+        {
+            return delayed.clone();
+        }
+
+        let delayed = self.circuit().add_unary_operator(Z1::new(D::zero()), self);
+        self.circuit()
+            .cache()
+            .insert(DelayedId::new(self.local_node_id()), delayed.clone());
+
+        delayed
```
Kixiron, line 129: If you wanted you could condense this whole thing
```suggestion
        self
            .circuit()
            .cache()
            .get(&DelayedId::new(self.local_node_id()))
            .cloned()
            .unwrap_or_else(|| {
                let delayed = self.circuit().add_unary_operator(Z1::new(D::zero()), self);
                self.circuit()
                    .cache()
                    .insert(DelayedId::new(self.local_node_id()), delayed.clone());

                delayed
            })
```

src/circuit/cache.rs
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+
+        impl<$($typearg),*> std::hash::Hash for $constructor<$($typearg),*> {
+            fn hash<H>(&self, state: &mut H)
+                where
+                    H: std::hash::Hasher
+                    {
+                        self.0.hash(state);
+                    }
```
Kixiron, line 75:
```suggestion
                    H: std::hash::Hasher,
                {
                    self.0.hash(state);
                }
```

src/circuit/cache.rs
ryzhyk: Directly, yes, although it may instantiate other operators that also perform caching.  But I think the description is adequate.

ryzhyk:
I implemented most suggestions.  I will improve the ergonomics of the caching API in a separate PR.

PR #47, opened 2022-03-07
Nested incremental join.

ryzhyk:
We implement incremental join over nested streams according to the
following formula (see the full DBSP paper):

```
(↑((↑(a <> b))∆))∆ =
    I(↑I(a)) <> b            +
    ↑I(a) <> I(z^-1(b))      +
    a <> I(↑I(↑z^-1(b)))     +
    I(z^-1(a)) <> ↑I(↑z^-1(b)).
```

This required a series of other refactorings, which unfortunately ended
up bundled with this CV:

- `HasZero`, `HasOne` - remove default implementation of these traits
  for types that implement `num::Zero` and `num::One`, as these
  impls make it impossible to write impl
  `impl<T: HasZero> HasZero for Rc<T>`.  Instead we add macros to
  write such impls as one-liners for relevant types.
- Added `Minus` operator, currently implemented on top of `Add` and
  `Neg`, but the plan is that `GroupValue` will in the future support
  `Sub`, which can be slightly more efficient.
- Added `differentiate_nested`; also re-implemented `differentiate`
  as a coposition of `Minus` and `z-1`.
- Add caching support for `index`, `differentiate`,
  `differentiate_nested`.
- Generalize `join` to work with `SharedRef`.
- Change `Z1Nested` implementation to output the last value in the
  stored prefix instead of zero when asked to produce output for
  timestamps (the assumption here is that the circuit was evaluated to
  a fixed point at the previous parent timestamp, and hence the rest of
  the stream will contain the same fixed point value).

### DIFF
--- a/src/algebra/mod.rs
+++ b/src/algebra/mod.rs
@@ -10,7 +10,11 @@ pub use checked_int::CheckedInt;
 pub use finite_map::{FiniteHashMap, FiniteMap, MapBuilder, WithSupport};
 pub use zset::{IndexedZSet, IndexedZSetHashMap, ZSet, ZSetHashMap};
 
-use std::ops::{Add, AddAssign, Mul, Neg};
+use std::{
+    num::Wrapping,
+    ops::{Add, AddAssign, Mul, Neg},
+    rc::Rc,
+};
 
 /// A trait for types that have a zero value.
 /// This is simlar to the standard Zero trait, but that
@@ -20,16 +24,55 @@ pub trait HasZero {
     fn zero() -> Self;
 }
 
-/// Implementation of HasZero for types that already implement Zero
-impl<T> HasZero for T
+/// Implement `HasZero` for types that already implement `Zero`.
+macro_rules! impl_has_zero {
+    ($type:ty) => {
+        impl $crate::algebra::HasZero for $type {
+            fn is_zero(&self) -> bool {
+                <Self as num::traits::Zero>::is_zero(self)
+            }
+            fn zero() -> Self {
+                <Self as num::traits::Zero>::zero()
+            }
+        }
+    };
+}
+
+impl_has_zero!(u8);
+impl_has_zero!(u16);
+impl_has_zero!(u32);
+impl_has_zero!(u64);
+impl_has_zero!(u128);
+impl_has_zero!(usize);
+
+impl_has_zero!(i8);
+impl_has_zero!(i16);
+impl_has_zero!(i32);
+impl_has_zero!(i64);
+impl_has_zero!(isize);
+
+// TODO: Implement for `std::num::Saturating` once stable
+impl<T> HasZero for Wrapping<T>
 where
-    T: num::traits::Zero,
+    T: HasZero,
 {
     fn is_zero(&self) -> bool {
-        T::is_zero(self)
+        self.0.is_zero()
     }
     fn zero() -> Self {
-        T::zero()
+        Self(T::zero())
+    }
+}
+
+impl<T> HasZero for Rc<T>
+where
+    T: HasZero,
+{
+    fn is_zero(&self) -> bool {
+        T::is_zero(self.as_ref())
+    }
+    fn zero() -> Self {
+        Rc::new(T::zero())
     }
 }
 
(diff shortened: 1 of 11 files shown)


src/algebra/mod.rs
```diff
@@ -20,16 +23,42 @@ pub trait HasZero {
...
+impl_has_zero!(isize);
+
+impl<T> HasZero for Rc<T>
 where
-    T: num::traits::Zero,
+    T: HasZero,
 {
     fn is_zero(&self) -> bool {
-        T::is_zero(self)
+        <T as HasZero>::is_zero(self.as_ref())
```
Kixiron, line 58:
```suggestion
        T::is_zero(self.as_ref())
```

src/algebra/mod.rs
```diff
@@ -20,16 +23,42 @@ pub trait HasZero {
...
     }
     fn zero() -> Self {
-        T::zero()
+        Rc::new(<T as HasZero>::zero())
```
Kixiron, line 61:
```suggestion
        Rc::new(T::zero())
```

src/algebra/mod.rs
```diff
@@ -40,13 +69,36 @@ pub trait HasOne {
...
+impl_has_one!(u32);
+impl_has_one!(u64);
+impl_has_one!(u128);
+impl_has_one!(usize);
+
+impl_has_one!(i8);
+impl_has_one!(i16);
+impl_has_one!(i32);
+impl_has_one!(i64);
+impl_has_one!(isize);
```
Kixiron, line 94: Could add the `NonZero` variants to this

src/algebra/mod.rs
```diff
@@ -20,16 +23,42 @@ pub trait HasZero {
...
+impl_has_zero!(u64);
+impl_has_zero!(u128);
+impl_has_zero!(usize);
+
+impl_has_zero!(i8);
+impl_has_zero!(i16);
+impl_has_zero!(i32);
+impl_has_zero!(i64);
+impl_has_zero!(isize);
+
```
Kixiron, line 52:
```suggestion
// TODO: Implement for `std::num::Saturating` once stable
impl<T> HasZero for Wrapping<T>
where
    T: HasZero,
{
    fn is_zero(&self) -> bool {
        self.0.is_zero()
    }
    
    fn zero() -> Self {
        Self(T::zero())
    }
}
```

src/operator/join.rs
```diff
@@ -4,38 +4,57 @@ use crate::{
...
+        SR2: SharedRef + 'static,
+        <SR2 as SharedRef>::Target: IndexedZSet<K, V2, W>,
+        for<'a> &'a <SR2 as SharedRef>::Target: IntoIterator<
+            Item = (
+                &'a K,
+                &'a <<SR2 as SharedRef>::Target as IndexedZSet<K, V2, W>>::ZSet,
+            ),
+        >,
+        for<'a> &'a <<SR2 as SharedRef>::Target as IndexedZSet<K, V2, W>>::ZSet:
+            IntoIterator<Item = (&'a V2, &'a W)>,
```
Kixiron, line 48: Yikes

ryzhyk: I know, it keeps getting better.

src/operator/adapter.rs
```diff
@@ -28,15 +28,14 @@ use crate::{
 /// Unary operator adapter unwraps input values of type
 /// `I` wrapped in a shared reference.  See
 /// [module-level documentation](`crate::operator::adapter`) for details.
-pub struct UnaryOperatorAdapter<I, O, Op> {
+pub struct UnaryOperatorAdapter<O, Op> {
```
mbudiu-vmw, line 31: there is still an I in the comments.

src/operator/join.rs
mbudiu-vmw: I wonder if there is a Guinness record for the longest chain of constraints we can apply for.

src/operator/join.rs
```diff
@@ -311,4 +377,99 @@ mod test {
...
+                    .integrate()
+                    .integrate_nested()
+                    .apply(|i| i.distinct())
+                    .differentiate()
+                    .differentiate_nested();
+                paths_delayed.connect(&paths);
+                let output = paths.integrate();
+                Ok((
+                    vec![
+                        paths.condition(|delta| delta.support_size() == 0),
```
mbudiu-vmw, line 457: why are both of these needed?
Isn't support_size() == 0 as efficient as is_zero()? I find is_zero clearer and cleaner.

src/operator/adapter.rs
ryzhyk: ack

src/operator/join.rs
ryzhyk: > why are both of these needed?

This is subtle stuff and we should discuss it, ideally in person.  It is related to the semantics of z-1 with finite prefixes.

> Isn't support_size() == 0 as efficient as is_zero()? I find is_zero clearer and cleaner.

yep, thanks!

PR #48, opened 2022-03-07
More ergonomic caching API.

ryzhyk:
Add `Circuit::cache_lookup_or_insert_with` that combines cache lookup
with insertion on cache miss.

### DIFF
--- a/src/circuit/circuit_builder.rs
+++ b/src/circuit/circuit_builder.rs
@@ -45,7 +45,7 @@ use std::{
 
 use crate::{
     circuit::{
-        cache::CircuitCache,
+        cache::{CircuitCache, CircuitStoreMarker},
         operator_traits::{
             BinaryOperator, Data, ImportOperator, NaryOperator, SinkOperator, SourceOperator,
             StrictUnaryOperator, UnaryOperator,
@@ -58,7 +58,7 @@ use crate::{
     },
     circuit_cache_key,
 };
-use typedmap::TypedMap;
+use typedmap::{TypedMap, TypedMapKey};
 
 /// Value stored in the stream.
 struct StreamValue<D> {
@@ -173,15 +173,9 @@ impl<P, D> Stream<Circuit<P>, D> {
         P: Clone + 'static,
         D: 'static,
     {
-        if let Some(export) = self
-            .circuit()
-            .cache()
-            .get(&ExportId::new(self.local_node_id()))
-        {
-            return export.clone();
-        }
-
-        unimplemented!();
+        self.circuit()
+            .cache_get_or_insert_with(ExportId::new(self.local_node_id()), || unimplemented!())
+            .clone()
     }
 }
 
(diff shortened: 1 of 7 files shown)


mbudiu-vmw (approved):

src/circuit/circuit_builder.rs
```diff
@@ -892,19 +886,39 @@ where
...
+    pub(crate) fn cache_lookup_or_insert_with<K, F>(&self, key: K, mut f: F) -> RefMut<'_, K::Value>
+    where
+        K: 'static + TypedMapKey<CircuitStoreMarker>,
+        F: FnMut() -> K::Value,
+    {
+        // Don't use `store.entry()`, since `f` may need to perform
+        // its own cache lookup.
+        if self.inner().store.contains_key(&key) {
+            return RefMut::map(self.inner_mut(), |c| c.store.get_mut(&key).unwrap());
+        }
```
Kixiron, line 901: This causes two lookups in the happy path which is less than ideal. There's two ways to address this, one would be using hashbrown to allow dancing around the hashing somewhat and the other would be to use [`RefMut::filter_map()`](https://github.com/rust-lang/rust/issues/81061) which doesn't seem to be stable yet which is very unfortunate.
Ideally we'd do both, so I'll just put that as a todo
```suggestion
        // its own cache lookup.
        // TODO: Only perform hashing once by using hashbrown's raw api
        // TODO: Use `RefMut::filter_map()` to only perform one lookup in the happy path
        //       https://github.com/rust-lang/rust/issues/81061
        if self.inner().store.contains_key(&key) {
            return RefMut::map(self.inner_mut(), |c| c.store.get_mut(&key).unwrap());
        }
```

src/circuit/circuit_builder.rs
```diff
@@ -892,19 +886,39 @@ where
         self.inner().parent.clone()
     }
 
-    /// Returns a reference to the cache associated with the circuit.
+    /// Lookup a value in the circuit cache or insert a value on cache miss.
     ///
     /// See [`cache`] module documentation for details.
-    pub(crate) fn cache(&self) -> RefMut<'_, CircuitCache> {
-        let circuit = self.inner_mut();
+    pub(crate) fn cache_lookup_or_insert_with<K, F>(&self, key: K, mut f: F) -> RefMut<'_, K::Value>
```
Kixiron, line 892:
```suggestion
    pub(crate) fn cache_get_or_insert_with<K, F>(&self, key: K, mut f: F) -> RefMut<'_, K::Value>
```
`.get_or_insert` is a more standard (and shorter) naming scheme

src/circuit/circuit_builder.rs
```diff
@@ -892,19 +886,39 @@ where
         self.inner().parent.clone()
     }
 
-    /// Returns a reference to the cache associated with the circuit.
+    /// Lookup a value in the circuit cache or insert a value on cache miss.
```
Kixiron, line 889:
```suggestion
    /// Lookup a value in the circuit cache or create and insert a new value if it does not exist
```

src/circuit/circuit_builder.rs
ryzhyk: This API is used at circuit construction time. I don't care about the overhead of extra lookup.  `filter_map` would be nice, but raw API isn't worth it.

src/circuit/circuit_builder.rs
ryzhyk: but it's "cache lookup", not "cache getting".

Kixiron: It's getting a value from the cache, right? That's what the return value is, at least

PR #49, opened 2022-03-09
Distinct operator.

ryzhyk:
Implemented three flavors of the `distinct` operator:

- Non-incremental: `Stream::distinct`.

- Incremental: `Stream::distinct_incremental`.  Functionally this
  operator is equivalent to `D(distinct(I(stream))`, but the implementation is
  incremental, only looking at values in the support of the latest delta.

- Incremental nested: `Stream::distinct_incremental_nested`.  Complete
  implementation:
  ```
  self.integrate_nested()
      .distinct_incremental()
      .differentiate_nested()
  ```

We also add a version of the `Generator` operator (`GeneratorNested`)
that generates nested streams.  This is useful for testing nested
operators.

### DIFF
--- a/src/operator/distinct.rs
+++ b/src/operator/distinct.rs
@@ -1,0 +1,292 @@
+//! Distinct operator.
+
+use std::{borrow::Cow, marker::PhantomData};
+
+use crate::{
+    algebra::{finite_map::KeyProperties, ZRingValue, ZSet},
+    circuit::{
+        operator_traits::{BinaryOperator, Operator, UnaryOperator},
+        Circuit, NodeId, Scope, Stream,
+    },
+    circuit_cache_key,
+    operator::{BinaryOperatorAdapter, UnaryOperatorAdapter},
+    SharedRef,
+};
+
+circuit_cache_key!(DistinctId<C, D>(NodeId => Stream<C, D>));
+circuit_cache_key!(DistinctIncrementalId<C, D>(NodeId => Stream<C, D>));
+
+impl<P, Z> Stream<Circuit<P>, Z>
+where
+    P: Clone + 'static,
+{
+    /// Apply [`Distinct`] operator to `self`.
+    pub fn distinct<V, W>(&self) -> Stream<Circuit<P>, <Z as SharedRef>::Target>
+    where
+        V: KeyProperties,
+        W: ZRingValue,
+        Z: SharedRef + 'static,
+        <Z as SharedRef>::Target: ZSet<V, W>,
+    {
+        self.circuit()
+            .cache_get_or_insert_with(DistinctId::new(self.local_node_id()), || {
+                self.circuit()
+                    .add_unary_operator(UnaryOperatorAdapter::new(Distinct::new()), self)
+            })
+            .clone()
+    }
+
+    /// Incremental version of the [`Distinct`] operator.
+    ///
+    /// This is equivalent to `self.integrate().distinct().differentiate()`, but
+    /// is more efficient.
+    pub fn distinct_incremental<V, W>(&self) -> Stream<Circuit<P>, <Z as SharedRef>::Target>
+    where
+        V: KeyProperties,
+        W: ZRingValue,
+        Z: SharedRef + 'static,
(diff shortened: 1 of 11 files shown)


src/operator/distinct.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+
+/// Incremental version of the distinct operator.
+///
+/// Takes a stream `a` of changes to relation `A` and a stream with delayed
+/// value of `A`: `z^-1(A) = a.integrate().delay()` and computes
+/// `distinct(A) - distinct(z^-1(A))` incrementally, by only considering
+/// values in the support of `a`.
+struct DistinctIncremental<V, W, Z> {
```
mbudiu-vmw, line 128: Technically this is LiftedDistinctIncremental.

src/operator/distinct.rs
```diff
@@ -0,0 +1,288 @@
...
+    for<'a> &'a Z: IntoIterator<Item = (&'a V, &'a W)>,
+    Z: ZSet<V, W> + 'static,
+{
+    fn eval(&mut self, delta: &Z, delayed_integral: &Z) -> Z {
+        let mut result = Z::empty();
+
+        for (v, w) in delta.into_iter() {
+            let weight = delayed_integral.lookup(v);
+
+            if weight.is_zero() {
```
mbudiu-vmw, line 170: I don't think this is general enough.
This works if the input is a set, but once you move distincts around inputs may be multisets.
So it is safer to use the formulas in the paper, with <= 0 and >= 0.
I realize that ZRingValue does not have le0, could be a bit more efficient than negating.

src/operator/index.rs
```diff
@@ -161,7 +161,7 @@ mod test {
                 finite_map!{ 1 => finite_map!{"a" => 3, "d" => 1, "e" => 1}, 2 => finite_map!{"c" => 1}, 3 => finite_map!{"a" => 2}},
             ].into_iter();
             circuit.add_source(Generator::new(move || inputs.next().unwrap() ))
-                   .index()
+                   .index::<_,_,_,FiniteHashMap<_,_>>()
```
mbudiu-vmw, line 164: this is certainly uglier

src/operator/integrate.rs
```diff
@@ -78,7 +85,7 @@ where
                 // ```
                 let feedback = DelayedFeedback::new(self.circuit());
                 let integral = self.circuit().add_binary_operator_with_preference(
-                    Plus::new(),
+                    <BinaryOperatorAdapter<D::Target, _>>::new(<Plus<D::Target>>::new()),
```
mbudiu-vmw, line 88: so is this.

mbudiu-vmw (approved):

mbudiu-vmw (changes requested):
See comments.

src/operator/distinct.rs
ryzhyk: Sorry, I don't understand, why won't this work for multisets?

mbudiu-vmw: The paper says that the weight is 1 when it changes from negative to positive. You require it to be exactly 0.

ryzhyk: I guess because we want distinct to map negative values to 0. Fair enough.  Is that an arbitrary choice or is there a reason theory wouldn't work if distinct(-1) was was 1 instead of 0.

mbudiu-vmw: It won't work because it does not compute set difference correctly.

src/operator/distinct.rs
ryzhyk: I have so far omitted the `Lifted` part in all operator names.  It somehow doesn't seem to make anything clearer.

mbudiu-vmw: Because for most it doesn't make a difference

src/operator/distinct.rs
ryzhyk: Should be fixed now.

src/operator/index.rs
Kixiron, line 164: Formatting
```suggestion
                   .index::<_, _, _, FiniteHashMap<_, _>>()
```

mbudiu-vmw (approved):

PR #50, opened 2022-03-10
`map_keys`, `filter_map_keys` operators.

ryzhyk:
Please look at the **last** commit only.

We implement two flavors of both operators that work on borrowed and owned keys respectively.
 
As always, we make operators generic over input and output collections, which means that the user needs to write type annotations when instantiating the operators.

### DIFF
--- a/src/operator/join.rs
+++ b/src/operator/join.rs
@@ -436,11 +436,11 @@ mod test {
                 //      join_incremental_nested
                 // ```
                 let edges = edges.delta0(child);
-                let paths_delayed = DelayedFeedback::new(child);
+                let paths_delayed = <DelayedFeedback<_, FiniteHashMap<_, _>>>::new(child);
 
                 let paths_inverted: Stream<_, FiniteHashMap<(usize, usize), isize>> = paths_delayed
                     .stream()
-                    .apply(|paths: &FiniteHashMap<(usize, usize), isize>| paths.into_iter().map(|((x,y), w)| ((*y, *x), *w)).collect());
+                    .map_keys(|&(x, y)| (y, x));
 
                 let paths_inverted_indexed: Stream<_, FiniteHashMap<usize, FiniteHashMap<usize, isize>>> = paths_inverted.index();
                 let edges_indexed: Stream<_, FiniteHashMap<usize, FiniteHashMap<usize, isize>>> = edges.index();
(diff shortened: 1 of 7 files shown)


mbudiu-vmw (approved):
I wonder whether this map-filter is more efficient than a chain of two. I have no idea how expensive Option is

ryzhyk:
> I wonder whether this map-filter is more efficient than a chain of two. I have no idea how expensive Option is

`Option` will likely have trivial overhead. A chain of two will involve building an intermediate collection, which is going to be more expensive.

mbudiu-vmw:
That is not obvious to me if the colection is just the iterator

ryzhyk:
> That is not obvious to me if the colection is just the iterator

It is at least a vector, we don't pass iterators between operators.

ryzhyk:
> > That is not obvious to me if the colection is just the iterator
> 
> It is at least a vector, we don't pass iterators between operators.

But things like this are easy to benchmark.

Kixiron:
> That is not obvious to me if the colection is just the iterator

Iterators are lazy and most generally don't require any sort of allocation, they're generally compiled into loops

src/operator/join.rs
```diff
@@ -436,11 +436,11 @@ mod test {
                 //      join_incremental_nested
                 // ```
                 let edges = edges.delta0(child);
-                let paths_delayed = DelayedFeedback::new(child);
+                let paths_delayed = <DelayedFeedback<_, FiniteHashMap<_,_>>>::new(child);
```
Kixiron, line 439:
```suggestion
                let paths_delayed = <DelayedFeedback<_, FiniteHashMap<_, _>>>::new(child);
```

src/operator/map.rs
```diff
@@ -0,0 +1,221 @@
...
+            .map(|(k, v)| ((self.map_owned)(k), v))
+            .collect()
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use std::vec;
+
+    use crate::{
```
Kixiron, line 142:
```suggestion
    use std::vec;
    use crate::{
```
I'd also just straight up import `vec::IntoIter` here since it's a test and doesn't clash with anything

src/ref_pair.rs
```diff
@@ -0,0 +1,23 @@
+//!
+
+/// Trait for types that can be converted into a pair of references
```
Kixiron, line 3:
```suggestion
/// Trait for types that can be converted into a pair of references
```

Kixiron (changes requested):

PR #52, opened 2022-03-16
Initial implementation of the Galen benchmark.

ryzhyk:
Galen benchmark and some new infrastructure to support it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,9 +4,9 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-
 default = ["with-serde"]
 with-serde = ["serde"]
+with-csv = ["csv"]
 
 [dependencies]
 num = "0.4.0"
@@ -17,5 +17,10 @@ crossbeam-utils = "0.8.6"
 once_cell = "1.9.0"
 priority-queue = "1.2.1"
 hashbrown = "0.12.0"
-csv = "1.1"
+csv = { version = "1.1", optional = true }
 serde = { version = "1.0", optional = true }
+
+[[bench]]
+name = "galen"
+harness = false
+required-features = ["with-csv"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,10 @@ name = "dbsp"
 version = "0.1.0"
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+
+default = ["with-serde"]
+with-serde = ["serde"]
 
 [dependencies]
 num = "0.4.0"
@@ -14,3 +17,5 @@ crossbeam-utils = "0.8.6"
 once_cell = "1.9.0"
 priority-queue = "1.2.1"
 hashbrown = "0.12.0"
+csv = "1.1"
+serde = { version = "1.0", optional = true }
--- a/benches/galen.rs
+++ b/benches/galen.rs
@@ -1,0 +1,178 @@
+//! Galen benchmark from
+//! `https://github.com/frankmcsherry/dynamic-datalog/tree/master/problems/galen`
+
+use csv::{Reader as CsvReader, ReaderBuilder};
+use dbsp::{
+    algebra::{FiniteMap, HasZero, ZSetHashMap},
+    circuit::{Root, Runtime, Stream},
+    operator::{CsvSource, DelayedFeedback},
+};
+use std::{fs::File, path::PathBuf};
+
+/*
+.decl p(X: Number, Z: Number)
+.decl q(X: Number, Y: Number, Z: Number)
+.decl r(R: Number, P: Number, E: Number)
+.decl c(Y: Number, Z: Number, W: Number)
+.decl u(R: Number, Z: Number, W: Number)
+.decl s(R: Number, P: Number)
+
+.input p(IO="file", filename="p.txt", delimiter=",")
+.input q(IO="file", filename="q.txt", delimiter=",")
+.input r(IO="file", filename="r.txt", delimiter=",")
+.input c(IO="file", filename="c.txt", delimiter=",")
+.input u(IO="file", filename="u.txt", delimiter=",")
+.input s(IO="file", filename="s.txt", delimiter=",")
+
+p(?x,?z) :- p(?x,?y), p(?y,?z).
+p(?x,?z) :- p(?y,?w), u(?w,?r,?z), q(?x,?r,?y).
+p(?x,?z) :- c(?y,?w,?z),p(?x,?w), p(?x,?y).
+q(?x,?r,?z) :- p(?x,?y), q(?y,?r,?z).
+q(?x,?q,?z) :- q(?x,?r,?z),s(?r,?q).
+q(?x,?e,?o) :- q(?x,?y,?z),r(?y,?u,?e),q(?z,?u,?o).
+*/
+
+type Number = u32;
+type Weight = isize;
+
+fn csv_source<T>(file: &str) -> CsvSource<File, ZSetHashMap<T, Weight>, T>
+where
+    T: Clone,
+{
+    let path: PathBuf = ["benches", "galen_data", file].iter().collect();
+
+    let reader: CsvReader<File> = ReaderBuilder::new()
+        .delimiter(b',')
+        .has_headers(false)
+        .from_path(path)
+        .unwrap();
+    CsvSource::from_csv_reader(reader)
+}
+
+fn main() {
+    let hruntime = Runtime::run(1, |_runtime, _index| {
+        println!("cwd: {:?}", std::env::current_dir().unwrap());
+
+        let root = Root::build(move |circuit| {
+            let p_source = csv_source::<(Number, Number)>("p.txt");
+            let q_source = csv_source::<(Number, Number, Number)>("q.txt");
+            let r_source = csv_source::<(Number, Number, Number)>("r.txt");
+            let c_source = csv_source::<(Number, Number, Number)>("c.txt");
+            let u_source = csv_source::<(Number, Number, Number)>("u.txt");
+            let s_source = csv_source::<(Number, Number)>("s.txt");
+
+            let p: Stream<_, ZSetHashMap<_, Weight>> = circuit.add_source(p_source);
+            let q: Stream<_, ZSetHashMap<_, Weight>> = circuit.add_source(q_source);
+            let r: Stream<_, ZSetHashMap<_, Weight>> = circuit.add_source(r_source);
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -3,7 +3,10 @@ name = "dbsp"
 version = "0.1.0"
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+
```
Kixiron, line 7:
```suggestion
```

Cargo.toml
```diff
@@ -14,3 +17,10 @@ crossbeam-utils = "0.8.6"
 once_cell = "1.9.0"
 priority-queue = "1.2.1"
 hashbrown = "0.12.0"
+csv = "1.1"
+serde = { version = "1.0", optional = true }
+
+[[bench]]
+
+name = "galen"
```
Kixiron, line 25:
```suggestion
[[bench]]
name = "galen"
```

benches/galen.rs
```diff
@@ -0,0 +1,177 @@
+use dbsp::{
+    algebra::{FiniteMap, HasZero, ZSetHashMap},
+    circuit::{Root, Runtime, Stream},
+    operator::{CsvSource, DelayedFeedback},
+};
+
+use csv::{Reader as CsvReader, ReaderBuilder};
+
+use std::{fs::File, path::PathBuf};
```
Kixiron, line 9:
```suggestion
};
use csv::{Reader as CsvReader, ReaderBuilder};
use std::{fs::File, path::PathBuf};
```

benches/galen.rs
```diff
@@ -0,0 +1,177 @@
...
+p(?x,?z) :- c(?y,?w,?z),p(?x,?w), p(?x,?y).
+q(?x,?r,?z) :- p(?x,?y), q(?y,?r,?z).
+q(?x,?q,?z) :- q(?x,?r,?z),s(?r,?q).
+q(?x,?e,?o) :- q(?x,?y,?z),r(?y,?u,?e),q(?z,?u,?o).
+*/
+
+type Number = u32;
+type Weight = isize;
+
+fn mk_csv_source<T>(fname: &str) -> CsvSource<File, ZSetHashMap<T, Weight>, T>
```
Kixiron, line 37:
```suggestion
fn csv_source<T>(fname: &str) -> CsvSource<File, ZSetHashMap<T, Weight>, T>
```

benches/galen.rs
```diff
@@ -0,0 +1,177 @@
...
+where
+    T: Clone,
+{
+    let path: PathBuf = ["benches", "galen_data", fname].iter().collect();
```
Kixiron, line 41:
```suggestion
fn mk_csv_source<T>(file: &str) -> CsvSource<File, ZSetHashMap<T, Weight>, T>
where
    T: Clone,
{
    let path: PathBuf = ["benches", "galen_data", file].iter().collect();
```

src/algebra/zset/mod.rs
```diff
@@ -82,6 +82,21 @@ where
 /// An implementation of ZSets using [`FiniteHashMap`]s
 pub type ZSetHashMap<Data, Weight> = FiniteHashMap<Data, Weight>;
 
+/// Build a Z-set from an iterator over values by adding each value with
+/// weight 1.
```
Kixiron, line 86:
```suggestion
/// Build a Z-set from an iterator, giving each item a weight of 1
```

src/algebra/zset/mod.rs
```diff
@@ -82,6 +82,21 @@ where
...
+impl<Data, Weight> FromIterator<Data> for ZSetHashMap<Data, Weight>
+where
+    Data: KeyProperties,
+    Weight: ZRingValue,
+{
+    fn from_iter<T>(iter: T) -> Self
+    where
+        T: IntoIterator<Item = Data>,
+    {
+        Self::from_iter(iter.into_iter().map(|d| (d, Weight::one())))
```
Kixiron, line 96: Is this infinitely recursive? It looks like it calls itself again

src/operator/csv.rs
```diff
@@ -0,0 +1,110 @@
+//! Source operator that reads data from a CSV file.
+
+// TODO:
+// - Error handling (currently we just panic on reader error or deserialization error).
+// - Batching (don't read the whole file in one clock cycle)
+// - Async implementation (wait for data to become available in the reader)
+// - Sharded implementation.
+
+#![cfg(feature = "serde")]
```
Kixiron, line 9:
```suggestion
//! Source operator that reads data from a CSV file.
#![cfg(feature = "serde")]

// TODO:
// - Error handling (currently we just panic on reader error or deserialization error).
// - Batching (don't read the whole file in one clock cycle)
// - Async implementation (wait for data to become available in the reader)
// - Sharded implementation.

```

src/operator/csv.rs
```diff
@@ -0,0 +1,110 @@
+//! Source operator that reads data from a CSV file.
```
Kixiron, line 1: Does this really need to be part of the library itself? At very least it should be feature gated, most users won't need it and so don't need the dependencies it pulls in

Kixiron (changes requested):

benches/galen.rs
```diff
@@ -0,0 +1,177 @@
...
+    let reader: CsvReader<File> = ReaderBuilder::new()
+        .delimiter(b',')
+        .has_headers(false)
+        .from_path(path)
+        .unwrap();
+    CsvSource::from_csv_reader(reader)
+}
+
+fn main() {
+    let hruntime = Runtime::run(1, |_runtime, _index| {
```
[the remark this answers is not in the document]
ryzhyk: I don't think this is a good idea.  Users should be able to change the number of workers without changing anything else in their code.  It is still possible to instantiate the circuit without a runtime though, in which case it will run on the caller's thread.

src/algebra/zset/mod.rs
ryzhyk: It calls a different `FromIterator` implementation, which takes an iterator over (value, weight) pairs.

src/operator/csv.rs
ryzhyk: That's a good point.  It makes sense to keep non-core operators in separate crates.  Which by the way reminds me that we should move to crates.io now that the project is open source.

This particular operator we are using it in benchmarks that live in the same crate, so I don't think it can be easily separated. I am also not sure how to feature gate it so that benchmarks can still use it.  Any ideas?

benches/galen.rs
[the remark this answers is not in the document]
Kixiron: This wouldn't really change anything else within their code, the first worker would always be spawned on the current thread so if the user always wanted the runtime to occur on another thread they'd just wrap it in a thread no matter what

benches/galen.rs
```diff
@@ -0,0 +1,177 @@
...
+                        qvar.stream().index_with(|&(x, y, z)| ((y, z), x));
+                    let c_by_2: Stream<_, ZSetHashMap<_, _>> =
+                        c.delta0(child).index_with(|&(x, y, z)| (y, (x, z)));
+                    let r_by_1: Stream<_, ZSetHashMap<_, _>> =
+                        r.delta0(child).index_with(|&(x, y, z)| (x, (y, z)));
+                    let s_by_1: Stream<_, ZSetHashMap<_, _>> = s.delta0(child).index();
+
+                    // IR1: p(x,z) :- p(x,y), p(y,z).
+                    let ir1 = p_by_2.join_incremental_nested(&p_by_1, |&_y, &x, &z| (x, z));
+                    ir1.inspect(|zs: &ZSetHashMap<_, _>| println!("ir1: {}", zs.support_size()));
```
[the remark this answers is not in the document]
ryzhyk: We don't.  This is really wip code.  It is also missing time/memory measurements and result validation.  I wanted to commit it though before working on circuit visualization and profiling.

benches/galen.rs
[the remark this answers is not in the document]
ryzhyk: But what's the advantage?  Rust threads are light-weight enough that creating an extra one doesn't matter.

src/operator/csv.rs
Kixiron: You'd pretty much just feature gate the relevant code, add `optional = true` to the `csv` dependency and then add `required-features = ["csv"]` to the benches that need it

mbudiu-vmw (approved):

benches/galen.rs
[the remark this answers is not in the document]
Kixiron: Control, mostly. As an example, WASM doesn't support threads, so this would enable users on WASM to actually run dbsp (either in the browser or as part of a sandboxed plugin) which I think is something we should enable without forcing them to copy+paste the entire runtime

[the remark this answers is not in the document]
ryzhyk: WASM support is a valid argument. I'll create an issue for this.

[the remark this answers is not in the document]
ryzhyk: On a second thought, one shouldn't use `Runtime` if they only want one thread.  `Runtime` is only meant to provide services for multithreaded execution.  The circuit API is fully functional in a single-threaded context without a `Runtime`.

src/operator/join.rs
```diff
@@ -132,14 +132,14 @@ where
         Z: ZSet<V, W>,
     {
         let join1: Stream<_, Z> = self
-            .integrate()
             .integrate_nested()
+            .integrate()
```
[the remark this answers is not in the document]
ryzhyk: It should be more memory-efficient.

benches/galen.rs
```diff
@@ -0,0 +1,175 @@
...
+
+                    // IR2: q(x,r,z) := p(x,y), q(y,r,z)
+                    let ir2 = p_by_2.join_incremental_nested(&q_by_1, |&_y, &x, &(r, z)| (x, r, z));
+                    ir2.inspect(|zs: &ZSetHashMap<_, _>| println!("ir2: {}", zs.support_size()));
+
+                    // IR3: p(x,z) := p(y,w), u(w,r,z), q(x,r,y)
+                    let ir3 = p_by_2
+                        .join_incremental_nested::<_, _, _, _, _, _, _, ZSetHashMap<_, _>>(
```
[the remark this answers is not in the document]
ryzhyk: DD only has one join implementation that works for any scope.  We can do something similar as a convenience API.

src/operator/csv.rs
```diff
@@ -0,0 +1,110 @@
...
+#![cfg(feature = "with-serde")]
```
Kixiron, line 2: This should be `with-csv`, otherwise it's broken if only `with-serde` is enabled

PR #58, opened 2022-03-23
Rudimentary circuit visualization and profiling infrastructure.

ryzhyk:


src/num_entries.rs
```diff
@@ -0,0 +1,93 @@
...
+use impl_trait_for_tuples::impl_for_tuples;
+use std::rc::Rc;
+
+/// Trait to report object size as the number of entries.
+///
+/// Scalars have size 1.  Container (e.g., vector, map) size is
+/// the sum of sizes of its elements.  If elements have constant size,
+/// container size can be efficiently measured by multiplying the
+/// number of elements by the size of each element.
+pub trait NumEntries {
```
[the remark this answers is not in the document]
mbudiu-vmw: Perhaps something like SizeComplexity is a better name? Complexity because it is really like the O notation - it abstracts constant factors away.

src/num_entries.rs
```diff
@@ -0,0 +1,93 @@
...
+num_entries_scalar!(String);
+num_entries_scalar!(&'static str);
+
+#[impl_for_tuples(12)]
+impl NumEntries for Tuple {
+    fn num_entries(&self) -> usize {
+        1
+    }
+    fn const_num_entries() -> Option<usize> {
+        Some(1)
```
mbudiu-vmw, line 58: Are these tuples of scalars?
I could argue that the num_entries should really be the sum of the fields.

src/num_entries.rs
```diff
@@ -0,0 +1,93 @@
...
+/// Scalars have size 1.  Container (e.g., vector, map) size is
+/// the sum of sizes of its elements.  If elements have constant size,
+/// container size can be efficiently measured by multiplying the
+/// number of elements by the size of each element.
+pub trait NumEntries {
+    /// Returns the number of entries in `self`.
+    fn num_entries(&self) -> usize;
+
+    /// Returns `Some(n)` if `Self` has constant size or `None` otherwise.
+    fn const_num_entries() -> Option<usize>;
```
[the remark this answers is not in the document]
mbudiu-vmw: Perhaps this should be two different traits.

src/circuit/circuit_builder.rs
```diff
@@ -952,11 +979,28 @@ where
...
 
-        circuit.log_scheduler_event(&SchedulerEvent::eval_end(id));
+        circuit.log_scheduler_event(&SchedulerEvent::eval_end(circuit.nodes[id.0].as_ref()));
 
         Ok(())
     }
 
+    /// Evaluate closure `f` inside a new circuit region.
+    ///
+    /// A region is a logical grouping of circuit nodes.  Regions are used
```
mbudiu-vmw, line 989: This is not obvious to me - you are assuming that the closure behaves in a certain way, e.g., adds new nodes that are somehow connected to each other. If it doesn't then the region has nothing to do with a group of nodes.
Another choice may be to invert this API, where node insertion could take some additional logging code to invoke before and after.

src/circuit/circuit_builder.rs
```diff
@@ -1717,8 +1776,12 @@ where
     O: Clone,
     Op: UnaryOperator<I, O>,
 {
-    fn id(&self) -> NodeId {
-        self.id
+    fn local_id(&self) -> NodeId {
```
mbudiu-vmw, line 1779: could these be default implementations?

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
...
+    collections::{hash_map::Entry, HashMap},
+    slice,
+};
+
+use super::visual_graph::{
+    ClusterNode, Edge as VisEdge, Graph as VisGraph, Node as VisNode, SimpleNode,
+};
+use crate::circuit::{trace::EdgeKind, GlobalNodeId, NodeId};
+
+/// A region is a named grouping of operators in a circuit.
```
mbudiu-vmw, line 12: In the Java code I wrote regions were nothing but circuits.

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
...
+    pub(super) fn add_region(&mut self, parent: &RegionId, name: Cow<'static, str>) -> RegionId {
+        debug_assert_eq!(self.id, RegionId::root());
+
+        self.do_add_region(parent.0.as_slice(), name)
+    }
+
+    fn do_get_region(&mut self, path: &[usize]) -> &mut Region {
+        match path.split_first() {
+            None => self,
+            Some((id, ids)) => self.children[*id].do_get_region(ids),
```
mbudiu-vmw, line 128: I guess this can panic for out of bounds

src/circuit/operator_traits.rs
```diff
@@ -133,6 +133,10 @@ pub trait Operator: 'static {
         F: Fn() + Send + Sync + 'static,
     {
     }
+
+    fn summary(&self, output: &mut String) {
```
mbudiu-vmw, line 137: This deserves some documentation

src/operator/differentiate.rs
```diff
@@ -13,7 +14,7 @@ circuit_cache_key!(NestedDifferentiateId<C, D>(NodeId => Stream<C, D>));
 impl<P, D> Stream<Circuit<P>, D>
 where
     P: Clone + 'static,
-    D: GroupValue,
+    D: NumEntries + GroupValue,
```
mbudiu-vmw, line 17: Maybe this should be a new Trait "SizedGroupValue" or something similar.

src/monitor/visual_graph.rs
```diff
@@ -0,0 +1,126 @@
...
+    label: String,
+    nodes: Vec<Node>,
+}
+
+impl ClusterNode {
+    pub(super) fn new(id: Id, label: String, nodes: Vec<Node>) -> Self {
+        Self { id, label, nodes }
+    }
+
+    // TODO: We add a pair of enter/exit nodes to each cluster and connect all
```
mbudiu-vmw, line 61: graphviz supports edges to nodes in a cluster.

mbudiu-vmw: https://graphviz.org/Gallery/directed/cluster.html

src/monitor/mod.rs
```diff
@@ -0,0 +1,570 @@
+///! Trace monitor that validates event traces from a circuit.  It is
```
mbudiu-vmw, line 1: this seems similar to the DDlog 1.0 monitors

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
+use std::{
```
mbudiu-vmw, line 1: I would document this file

mbudiu-vmw: People may be confused by the fact that you have two graphs: one is the DBSP circuit, and the other one is this topological representation. In Java I had the circuit itself have methods to dump the topology.
I am still not sure why you chose to separate them.

mbudiu-vmw (approved):
Where did you use the tracing to find out the problems with the size of the data?

src/monitor/visual_graph.rs
ryzhyk: Not really. An edge must point to a node in the cluster.

src/num_entries.rs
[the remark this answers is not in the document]
ryzhyk: I don't understand the problem. `const_num_entries` is a static method (per type).

The trait returns the number of entries in an object, hence the name. For a finite map this is the size of support.  For an indexed z-set this is the sum of the number of entries in all partitions.

src/num_entries.rs
ryzhyk: yeah, this is kind of specialized for the two important use cases: finite maps and indexed z-sets. Everything else is a const.

src/num_entries.rs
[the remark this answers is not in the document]
mbudiu-vmw: It should only return the number of entries for a finite map if the element in the map has itself a constant size.
Perhaps it should be called "num_entries_if_all_same_size()"?
Will the number returned be the support_size() * size_of_an_element?

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    /// Returns a reference to the root region of `self` if
+    /// `self` is a circuit node.
+    pub(super) fn region(&self) -> Option<&Region> {
+        if let NodeKind::Circuit { region, .. } = &self.kind {
+            Some(region)
+        } else {
+            None
+        }
+    }
```
Kixiron, line 241: Convention is to name these methods as `.as_*`, e.g. `.as_children()` or `.as_mut_region()`

src/num_entries.rs
Kixiron, line 17: If you want to make this an actual const you can just make it an associated constant,
```suggestion
    /// Is `Some(n)` if `Self` has constant size and `None` otherwise.
    const STATIC_ENTRIES: Option<usize> = None;
```

src/num_entries.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+impl<T> NumEntries for Rc<T>
+where
+    T: NumEntries,
+{
+    fn num_entries(&self) -> usize {
+        self.as_ref().num_entries()
+    }
+    fn const_num_entries() -> Option<usize> {
```
Kixiron, line 90:
```suggestion
    }

    fn const_num_entries() -> Option<usize> {
```

src/num_entries.rs
```diff
@@ -0,0 +1,93 @@
...
+                let mut res = 0;
+                for x in self.iter() {
+                    res += x.num_entries();
+                }
+                res
+            }
+            Some(n) => n * self.len(),
+        }
+    }
+    fn const_num_entries() -> Option<usize> {
```
Kixiron, line 78:
```suggestion
    }

    fn const_num_entries() -> Option<usize> {
```

src/algebra/finite_map/mod.rs
```diff
@@ -189,6 +189,28 @@ pub struct FiniteHashMap<Key, Value> {
...
+{
+    fn num_entries(&self) -> usize {
+        match Value::const_num_entries() {
+            None => {
+                let mut res = 0;
+                for (_, v) in self.into_iter() {
+                    res += v.num_entries();
+                }
+                res
+            }
```
Kixiron, line 205:
```suggestion
            None => self.into_iter().map(NumEntries::num_entries).sum(),
```
Also, couldn't/shouldn't that just be an `.iter()` call instead of an `.into_iter()` one?

src/monitor/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+    pub fn visualize_circuit(&self) -> VisGraph {
+        self.visualize_circuit_annotate(&|_| "".to_string())
+    }
+
+    pub fn visualize_circuit_annotate<F>(&self, f: &F) -> VisGraph
+    where
+        F: Fn(&GlobalNodeId) -> String,
+    {
+        self.0.lock().unwrap().circuit.visualize(f)
+    }
```
Kixiron, line 161: This shouldn't take a reference to a closure and should take a `FnMut` to give greater flexibility. I'd also like to see a better name than `f` for the argument and some docs on what the closure is for
```suggestion
    pub fn visualize_circuit_annotate<F>(&self, f: F) -> VisGraph
    where
        F: FnMut(&GlobalNodeId) -> String,
    {
        self.0.lock().unwrap().circuit.visualize(f)
    }
```

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
...
+                region_ident.push('_');
+            }
+        }
+        region_ident
+    }
+
+    /// Output region as a cluster in a visual graph.
+    fn visualize<F>(&self, scope: &Node, annotate: &F) -> ClusterNode
+    where
+        F: Fn(&GlobalNodeId) -> String,
```
Kixiron, line 79: Don't take a reference to a closure and take an `FnMut`
```suggestion
    fn visualize<F>(&self, scope: &Node, mut annotate: F) -> ClusterNode
    where
        F: FnMut(&GlobalNodeId) -> String,
```

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
...
+    {
+        let mut nodes = Vec::new();
```
Kixiron, line 81:
```suggestion
        let mut nodes = Vec::with_capacity(self.nodes.len() + self.children.len());
```

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
...
+    /// Lookup node in the subtree with the root in `self` by path.
+    fn node_mut<'a>(&mut self, mut path: slice::Iter<'a, NodeId>) -> Option<&mut Node> {
+        match path.next() {
+            None => Some(self),
+            Some(node_id) => match &mut self.kind {
+                NodeKind::Circuit { children, .. } => children.get_mut(node_id)?.node_mut(path),
+                _ => None,
+            },
+        }
+    }
```
Kixiron, line 195:
```suggestion
    fn get_node<'a>(&self, mut path: slice::Iter<'a, NodeId>) -> Option<&Node> {
        match path.next() {
            None => Some(self),
            Some(node_id) => match &self.kind {
                NodeKind::Circuit { children, .. } => children.get(node_id)?.get_node(path),
                _ => None,
            },
        }
    }

    /// Lookup node in the subtree with the root in `self` by path.
    fn get_mut_node<'a>(&mut self, mut path: slice::Iter<'a, NodeId>) -> Option<&mut Node> {
        match path.next() {
            None => Some(self),
            Some(node_id) => match &mut self.kind {
                NodeKind::Circuit { children, .. } => children.get_mut(node_id)?.get_mut_node(path),
                _ => None,
            },
        }
    }
```

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    /// Locate node by its global id.
+    pub(super) fn node_ref(&self, id: &GlobalNodeId) -> Option<&Node> {
+        self.nodes.node_ref(id.path().iter())
+    }
+
+    /// Locate node by its global id.
+    pub(super) fn node_mut(&mut self, id: &GlobalNodeId) -> Option<&mut Node> {
+        self.nodes.node_mut(id.path().iter())
+    }
```
Kixiron, line 331:
```suggestion
    pub(super) fn get_node(&self, id: &GlobalNodeId) -> Option<&Node> {
        self.nodes.get_node(id.path().iter())
    }

    /// Locate node by its global id.
    pub(super) fn get_mut_node(&mut self, id: &GlobalNodeId) -> Option<&mut Node> {
        self.nodes.get_mut_node(id.path().iter())
    }
```

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    pub(super) fn add_edge(&mut self, from: &GlobalNodeId, to: &GlobalNodeId, kind: &EdgeKind) {
+        match self.edges.entry(from.clone()) {
+            Entry::Occupied(mut oe) => {
+                oe.get_mut().push((to.clone(), kind.clone()));
+            }
+            Entry::Vacant(ve) => {
+                ve.insert(vec![(to.clone(), kind.clone())]);
+            }
```
Kixiron, line 340:
```suggestion
        self.edges
            .entry(from.clone())
            .or_insert_with(Vec::with_capacity(1))
            .push((to.clone(), kind.clone());
```

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
...
+    pub(super) fn node_ref(&self, id: &GlobalNodeId) -> Option<&Node> {
+        self.nodes.node_ref(id.path().iter())
+    }
+
+    /// Locate node by its global id.
+    pub(super) fn node_mut(&mut self, id: &GlobalNodeId) -> Option<&mut Node> {
+        self.nodes.node_mut(id.path().iter())
+    }
+
+    pub(super) fn add_edge(&mut self, from: &GlobalNodeId, to: &GlobalNodeId, kind: &EdgeKind) {
```
Kixiron, line 333: This method should take ownership of these id vecs since it immediately clones them all

src/monitor/circuit_graph.rs
```diff
@@ -0,0 +1,379 @@
...
+            Entry::Vacant(ve) => {
+                ve.insert(vec![(to.clone(), kind.clone())]);
+            }
+        }
+    }
+
+    /// Output circuit graph as visual graph.
+    pub(super) fn visualize<F>(&self, annotate: &F) -> VisGraph
+    where
+        F: Fn(&GlobalNodeId) -> String,
```
Kixiron, line 347:
```suggestion
    pub(super) fn visualize<F>(&self, mut annotate: F) -> VisGraph
    where
        F: FnMut(&GlobalNodeId) -> String,
```

src/monitor/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+    /// Attach trace monitor to a circuit.  The monitor will register for
+    /// both `CircuitEvent`s and `SchedulerEvent`s.
+    pub fn attach(&self, circuit: &Circuit<()>, handler_name: &str) {
+        TraceMonitorInternal::attach(self.0.clone(), circuit, handler_name);
+    }
+
+    pub fn new<CE, SE>(circuit_error_handler: CE, scheduler_error_handler: SE) -> Self
+    where
+        CE: Fn(&CircuitEvent, &TraceError) + 'static,
+        SE: Fn(&SchedulerEvent, &TraceError) + 'static,
```
Kixiron, line 138: These should be `FnMut`s

src/monitor/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+                });
+            };
+        });
+    }
+
+    /// Create a new trace monitor with user-provided event handlers.
+    fn new<CE, SE>(circuit_error_handler: CE, scheduler_error_handler: SE) -> Self
+    where
+        CE: Fn(&CircuitEvent, &TraceError) + 'static,
+        SE: Fn(&SchedulerEvent, &TraceError) + 'static,
```
Kixiron, line 208: `FnMut`

src/monitor/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+    fn set_current_region(&mut self, region: RegionId) {
+        *self.region_stack.last_mut().unwrap() = region;
+    }
+
+    fn circuit_event(&mut self, event: &CircuitEvent) -> Result<(), TraceError> {
+        //println!("event: {}", event);
+        if event.is_node_event() {
+            let node_id = event.node_id().unwrap();
+            let local_node_id = node_id.local_node_id().ok_or(TraceError::EmptyPath)?;
+            let parent_id = node_id.parent_id().unwrap();
```
Kixiron, line 262: Use pattern matching to remove all of the unwraps

src/monitor/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+                    return Err(TraceError::UnexpectedNodeId);
+                }
+                self.current_scope = parent_id;
+                self.region_stack.pop();
+                Ok(())
+            }
+        } else if event.is_edge_event() {
+            let from = event.from().unwrap();
+            let to = event.to().unwrap();
+            let kind = event.edge_kind().unwrap();
```
Kixiron, line 338: Ditto, use pattern matching to remove all of the unwraps

src/monitor/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    fn scheduler_event(&mut self, event: &SchedulerEvent) -> Result<(), TraceError> {
+        // eprintln!("scheduler event: {}", event);
+        if !self.running() {
+            return Err(TraceError::InvalidEvent(Cow::from(
+                "scheduler event received after circuit execution has terminated",
+            )));
+        }
+        let current_node_id = self.current_node_id();
```
Kixiron, line 366:
```suggestion
        }

        let current_node_id = self.current_node_id();
```

src/monitor/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+                    .unwrap()
+                    .is_iterative()
+                    && !self.current_state().is_running()
+                {
+                    return Err(TraceError::InvalidEvent(Cow::from(format!(
+                        "received 'StepStart' event in state {} of an iterative circuit",
+                        self.current_state().name()
+                    ))));
+                }
+                if !self
```
Kixiron, line 414:
```suggestion
                } else if !self
```

src/monitor/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+                    .circuit
+                    .node_ref(&current_node_id)
+                    .unwrap()
+                    .is_iterative()
+                    && !self.current_state().is_idle()
+                {
```
Kixiron, line 420: Pull this out of the conditions
```suggestion
                let current_node = self
                    .circuit
                    .node_ref(&current_node_id)
                    .unwrap();

                if current_node.is_iterative() && !self.current_state().is_running() {
                    return Err(TraceError::InvalidEvent(Cow::from(format!(
                        "received 'StepStart' event in state {} of an iterative circuit",
                        self.current_state().name()
                    ))));

                } else if !(current_node.is_iterative() && self.current_state().is_idle()) {
```

src/monitor/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+                match self.current_state() {
+                    CircuitState::Step(visited_nodes) => {
+                        let expected_len = self
+                            .circuit
+                            .node_ref(&current_node_id)
+                            .unwrap()
+                            .children()
+                            .unwrap()
+                            .len();
+                        if visited_nodes.len() != expected_len {
```
Kixiron, line 440:
```suggestion
                            .len();

                        if visited_nodes.len() != expected_len {
```

src/monitor/mod.rs
Kixiron, line 366: You use the current node in pretty much every arm, hoist the expression up to here and reuse it
```suggestion
        let current_node_id = self.current_node_id();
        let current_node = self.circuit.get_node(&current_node_id).unwrap();
```

Kixiron (changes requested):

ryzhyk:
@mbudiu-vmw , @Kixiron , I will incorporate your feedback in a separate PR, following the data structure revamp, to avoid rebasing hell.

PR #62, opened 2022-04-21
Cpu profiler

ryzhyk:
Cleanup in preparation for Z-set refactoring; initial CPU profiler implementation.


src/algebra/zset/tests.rs
```diff
@@ -254,27 +254,9 @@ fn zset_tuple_tests() {
...
-    assert_eq!(
-        finite_map! {
-            (-1, 0) => CheckedI64::new(9),
-            (0, 0) => CheckedI64::new(9),
-        },
-        j,
-    );
 }
 
+/*
```
mbudiu-vmw, line 259: why not delete it altogether?

src/profile/cpu.rs
```diff
@@ -0,0 +1,101 @@
...
+#[derive(Clone, Default)]
+pub struct CPUProfiler(Rc<RefCell<CPUProfilerInner>>);
+
+impl CPUProfiler {
+    /// Create a new CPU profiler instance.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Attach CPU profiler to a circuit.  The profiler will start circuit's CPU
```
mbudiu-vmw, line 29: start -> measure?

src/profile/cpu.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub fn operator_profile(&self, node: &GlobalNodeId) -> Option<OperatorCPUProfile> {
+        if let Ok(this) = self.0.try_borrow() {
+            this.operators.get(node).cloned()
+        } else {
+            None
+        }
+    }
+}
+
+#[derive(Default)]
```
mbudiu-vmw, line 53: I think this should be before the impl for readability.

src/profile/cpu.rs
```diff
@@ -0,0 +1,101 @@
...
+                    op_profile.total_time += duration;
+                };
+            }
+            _ => (),
+        }
+    }
+}
+
+/// Per-operator CPU profile.
+#[derive(Clone, Default)]
```
mbudiu-vmw, line 83: For code readability I prefer the definition before the first use.

src/profile/cpu.rs
```diff
@@ -0,0 +1,101 @@
...
+            SchedulerEvent::EvalEnd { node } => {
+                if let Some(start_time) = self.start_times.remove(node.global_id()) {
+                    let duration = Instant::now().duration_since(start_time);
+                    let op_profile = self
+                        .operators
+                        .entry(node.global_id().clone())
+                        .or_insert_with(Default::default);
+                    op_profile.invocations += 1;
+                    op_profile.total_time += duration;
+                };
```
mbudiu-vmw, line 75: So the start time can be missing if you toggle profiling while running?

mbudiu-vmw: But in that case the results may be wrong if you start/stop profiling - you will use perhaps an old start time.

mbudiu-vmw (changes requested):

src/profile/cpu.rs
ryzhyk: If we add the ability to stop and re-start profiling, start times will need to be cleared for this reason.

PR #63, opened 2022-05-20
Sql compiler

mbudiu-vmw:
This is just a skeleton that puts together some big pieces.
It generates some compilable rust from a simple SQL query with a project operator.


mbudiu-vmw:
Moving to a separate repository.

PR #64, opened 2022-05-31
Z-set revamp

ryzhyk:
Major re-work of the Z-set implementation, API and operators based on it.  There are many todos, which I will document as issues. I broke it up into commits to simplify reviewing.


mbudiu-vmw (commented):
I will continue reviewing later, did 1/3 of the files so far.

src/operator/csv.rs
```diff
@@ -19,13 +22,14 @@ use std::{borrow::Cow, io::Read, marker::PhantomData};
 /// A source operator that reads records of type `T` from a CSV file.
 ///
 /// The operator reads the entire file and yields its contents
-/// in the first clock cycle.
-pub struct CsvSource<R, C, T> {
+/// in the first clock cycle as a Z-set with unit weights.
+pub struct CsvSource<R, T, W, C> {
```
mbudiu-vmw, line 26: I am wondering whether it isn't better to have this actually return tuples and use the makezset operator afterwards. This is one weakness of DD, that everything is a zset, but it doesn't have to be in DBSP.
If that's inefficient you can have separate methods to yield tuples and zsets.

src/operator/csv.rs
```diff
@@ -40,52 +44,74 @@ where
...
+            let input_iter = self.reader.deserialize().map(CsvResult::unwrap).into_iter();
+
+            let mut data = Vec::with_capacity(input_iter.size_hint().0);
+
+            for x in input_iter {
+                data.push(((x, ()), W::one()));
+            }
+
+            self.time += 1;
+            C::from_tuples((), data)
```
mbudiu-vmw, line 90: isn't there a way to do this without a vector, just using an iterator?

src/operator/filter.rs
```diff
@@ -72,30 +69,43 @@ where
...
-            .filter_map(|(k, v)| {
-                if (self.filter)(k) {
-                    Some((k.clone(), v.clone()))
-                } else {
-                    None
+        let mut cursor = i.cursor();
+
+        // We can use Builder because cursor yields ordered values.  This
+        // is a nice property of the filter operation.
+        let mut builder = CO::Builder::with_capacity((), i.len());
```
mbudiu-vmw, line 90: does this cause waste if the filter throws out many elements?

src/test/mod.rs
```diff
@@ -86,6 +86,7 @@ fn tuple_relation_test() {
 }
 
 // Test a filter on a relation containing tuples
+/*
```
mbudiu-vmw, line 89: delete?

src/time.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// tuple.
+    ///
+    /// ```ignore
+    /// assert_eq!((2, 3).recede(0), (2, 2));
+    /// assert_eq!((2, 3).recede(1), (1, 3));
+    /// ```
+    fn recede(&self, scope: Scope) -> Self;
+}
+
+/*
```
mbudiu-vmw, line 109: delete?

mbudiu-vmw (commented):
I read 1/3 more of the files, but I have the remaining big ones left.

src/algebra/zset/mod.rs
```diff
@@ -1,97 +1,24 @@
...
 
-use crate::algebra::{
-    finite_map::{FiniteHashMap, FiniteMap, KeyProperties, MapBuilder},
-    ZRingValue,
-};
+use crate::{algebra::GroupValue, trace::Batch, NumEntries, SharedRef};
+
+// TODO: allow arbitrary `Time` types?
+/// An indexed Z-set maps arbitrary keys to Z-set values.
+pub trait IndexedZSet:
```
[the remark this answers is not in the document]
ryzhyk: Nope, just a specialization of `Batch`. What do you reckon is missing?

src/algebra/zset/zset_macro.rs
```diff
@@ -0,0 +1,23 @@
+/// Allows easily creating
+/// [`OrdIndexedZSet`](crate::trace::ord::OrdIndexedZSet)s
+#[macro_export]
+macro_rules! indexed_zset {
```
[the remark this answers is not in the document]
ryzhyk: Right. I'll keep the name, but fix the comment to explain that this is used for testing and right now creates and OrdIndexedZSet, which is as good as anything.

src/circuit/circuit_builder.rs
```diff
@@ -1537,6 +1543,60 @@ where
...
+    /// continue ticking.
+    ///
+    /// The fixed point check is implemented by checking the following
+    /// condition:
+    ///
+    /// * All operators in the circuit are in such a state that, if their inputs
+    ///   remain constant (i.e., all future inputs are identical to the last
+    ///   input), then their outputs remain constant too.
+    ///
+    /// This is a necessary and sufficient condition that is also easy to check
```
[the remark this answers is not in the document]
ryzhyk: I = integrate?  integrate is a composition of Plus and Z. Plus is stateless so once the condition holds for Z, we're done.  Of course, to check it precisely for Z, we'd have to store the last 2 values, which is expensive.

Fortunately, we are about to stop using `integrate` operator in all Z-set operator.  Instead, we will use `integrate_trace` (upcoming PR), which implements a precise fixedpoint test.

src/circuit/circuit_builder.rs
```diff
@@ -1537,6 +1543,60 @@ where
...
+    /// [`Z1Nested`](`crate::operator::Z1Nested`)) require storing the last
+    /// two versions of the state instead of one and comparing them at each
+    /// cycle.  Such operators instead implement imprecise conservative checks,
+    /// e.g., check for a _specific_ fixed point, e.g., a fixed point where both
+    /// input and output of the operator is zero (or empty).  As a result, the
+    /// circuit may fail to detect other fixed point and will iterate forever.
+    /// The goal is to evolve the design so that circuits created using the
+    /// high-level API (`Stream::xxx` methods) implement accurate fixed
+    /// point checks.
+    pub fn fixedpoint<F, T>(&self, constructor: F) -> Result<T, SchedulerError>
```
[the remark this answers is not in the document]
ryzhyk: That operator is still around in `src/operator/condition.rs`

src/operator/apply.rs
```diff
@@ -47,6 +47,11 @@ where
 
     fn clock_start(&mut self, _scope: Scope) {}
     fn clock_end(&mut self, _scope: Scope) {}
+    fn fixedpoint(&self) -> bool {
```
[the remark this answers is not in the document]
ryzhyk: `FnMut` means that it can have state, and so may not be lifter.

src/operator/csv.rs
ryzhyk: The current Batcher API will actually reuse the heap allocation behind this vector, so this is deliberate, at least in the DD design.

src/operator/filter.rs
ryzhyk: It does actually, since these buffers can make it all the way to the output batch.  This is probably ok, because the batch will either get freed at the end of the current clock tick or get added to the trace, where it will likely get merged with other batches soon, at which point the waste is gone.

I'll add a comment, and we'll keep an eye on this in profiling results.

src/operator/aggregate.rs
```diff
@@ -1,64 +1,97 @@
...
+        F: Fn(&I, &mut I::Cursor) -> O::Key + Clone + 'static,
+        O: Clone + ZSet + DeepSizeOf + 'static,
+        O::R: ZRingValue,
+    {
+        self.integrate_nested()
+            .aggregate_incremental(f)
+            .differentiate_nested()
     }
 
+    /*
```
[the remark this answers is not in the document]
ryzhyk: This stuff no longer works, but we need to re-implement it at some point.  I'll create an issue.

ryzhyk:
@mbudiu-vmw, thanks for the review!  Are you done reviewing this?

mbudiu-vmw:
> @mbudiu-vmw, thanks for the review! Are you done reviewing this?

Not even close

src/trace/layers/ordered.rs
```diff
@@ -0,0 +1,673 @@
...
+use crate::{
+    algebra::{AddAssignByRef, AddByRef, NegByRef},
+    trace::layers::{advance, Builder, Cursor, MergeBuilder, Trie, TrieSlice, TupleBuilder},
+    NumEntries, SharedRef,
+};
+use deepsize::DeepSizeOf;
+use std::convert::{TryFrom, TryInto};
+use std::fmt::{Debug, Display, Formatter};
+use std::marker::PhantomData;
+use std::ops::{Add, AddAssign, Neg, Sub};
```
Kixiron, line 12: Merge these imports

src/trace/layers/ordered.rs
```diff
@@ -0,0 +1,673 @@
...
+            let key = cursor.key(storage);
+            writeln!(f, "{}:", key)?;
+            let (val_storage, val_cursor) = cursor.values(storage);
+
+            f.write_str(&indent(
+                &TrieSlice(val_storage, val_cursor).to_string(),
+                "    ",
+            ))?;
+            cursor.step(storage);
+        }
```
Kixiron, line 93: Wouldn't it be more fitting to use `Formatter::debug_map()` for this?

src/trace/layers/ordered.rs
```diff
@@ -0,0 +1,673 @@
...
+    K: Ord + Clone,
+    L: Trie,
+    O: OrdOffset,
+    <O as TryFrom<usize>>::Error: Debug,
+    <O as TryInto<usize>>::Error: Debug,
+{
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        if self.keys() == 0 {
```
Kixiron, line 187: Maybe this should be an `.is_empty()` function

src/trace/layers/ordered_leaf.rs
```diff
@@ -0,0 +1,415 @@
...
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        let TrieSlice(storage, cursor) = self;
+        let mut cursor: OrderedLeafCursor = cursor.clone();
+
+        while cursor.valid(storage) {
+            let (key, val) = cursor.key(storage);
+            writeln!(f, "{} -> {}", key, val)?;
+            cursor.step(storage);
+        }
```
Kixiron, line 66: `.debug_map()` thing again

src/trace/layers/ordered_leaf.rs
```diff
@@ -0,0 +1,415 @@
...
+// TODO: by-value merge
+impl<K, R> Add<Self> for OrderedLeaf<K, R>
+where
+    K: Ord + Clone,
+    R: Eq + HasZero + AddAssignByRef + Clone,
+{
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        if self.keys() == 0 {
```
Kixiron, line 81: `.is_empty()` function

src/trace/layers/ordered_leaf.rs
```diff
@@ -0,0 +1,415 @@
...
+    }
+}
+
+impl<K, R> AddAssign<Self> for OrderedLeaf<K, R>
+where
+    K: Ord + Clone,
+    R: Eq + HasZero + AddAssignByRef + Clone,
+{
+    fn add_assign(&mut self, rhs: Self) {
+        if rhs.keys() != 0 {
```
Kixiron, line 97: `!rhs.is_empty()`

src/trace/layers/ordered_leaf.rs
```diff
@@ -0,0 +1,415 @@
...
+    }
+}
+
+/// A builder for unordered values.
+pub struct OrderedLeafBuilder<K, R> {
+    /// Unordered values.
+    pub vals: Vec<(K, R)>,
+}
+
+impl<K: Ord + Clone, R: Eq + HasZero + AddAssignByRef + Clone> Builder
```
Kixiron, line 191: Maybe move all of these trait bounds to a where clause, it's kinda hard to get the gist of the impl block when you have to figure out where the traits end and the types begin

src/trace/layers/ordered_leaf.rs
```diff
@@ -0,0 +1,415 @@
...
+        let upper1 = cursor1.bounds.1;
+        let mut lower2 = cursor2.bounds.0;
+        let upper2 = cursor2.bounds.1;
+
+        self.vals.reserve((upper1 - lower1) + (upper2 - lower2));
+
+        // while both mergees are still active
+        while lower1 < upper1 && lower2 < upper2 {
+            match trie1.vals[lower1].0.cmp(&trie2.vals[lower2].0) {
+                ::std::cmp::Ordering::Less => {
```
Kixiron, line 240: Import `Ordering`

src/trace/layers/ordered_leaf.rs
```diff
@@ -0,0 +1,415 @@
...
+                    // determine how far we can advance lower1 until we reach/pass lower2
+                    let step = 1 + advance(&trie1.vals[(1 + lower1)..upper1], |x| {
+                        x.0 < trie2.vals[lower2].0
+                    });
+                    let step = std::cmp::min(step, 1000);
```
Kixiron, line 245: Import `min`

src/trace/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+        }
+        /// Describes the times of the updates in the batch.
+        fn description(&self) -> &Description<B::Time> {
+            (&**self).description()
+        }
+    }
+
+    /// Wrapper to provide cursor to nested scope.
+    pub struct RcBatchCursor<B: BatchReader> {
+        phantom: ::std::marker::PhantomData<B>,
```
Kixiron, line 321: Import `PhantomData`

tools/pre-push
```diff
@@ -7,10 +7,7 @@ echo "Checking Rust rules prior to push.  To run this check by hand invoke 'tool
...
 
-if [ $# -gt 0 ]; then
-    cargo fmt -- --config imports_granularity=Crate
-else
-    cargo fmt -- --check --config format_code_in_doc_comments=true --config wrap_comments=true
-    cargo clippy -- -D warnings
-fi
+cargo fmt -- --check --config format_code_in_doc_comments=true --config wrap_comments=true
+cargo clippy -- -D warnings
+
```
Kixiron, line 12: We should probably start running miri since we've got a good bit of unsafe code

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,840 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        let mut res = Vec::new();
+        self.map_batches(|batch| {
+            res.push(writeln!(
+                f,
+                "batch:\n{}",
+                indent(&batch.to_string(), "    ")
+            ))
+        });
+        res.into_iter().collect()
```
Kixiron, line 122: `.debug_map()` or `.debug_list()` maybe?

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,840 @@
...
+    fn consolidate(mut self) -> Option<Self::Batch> {
+        // Merge batches until there is nothing left to merge.
+        let mut fuel = isize::max_value();
+        while !self.reduced() {
+            self.exert(&mut fuel);
+        }
+        // Return the sole remaining batch (if one exists).
+        for merging in self.merging.into_iter() {
+            if let MergeState::Single(Some(batch)) = merging {
+                if batch.len() > 0 {
```
Kixiron, line 296:
```suggestion
                if !batch.is_empty() {
```

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,840 @@
...
+    ///
+    /// The units of effort are updates, and the method should be
+    /// thought of as analogous to inserting as many empty updates,
+    /// where the trace is permitted to perform proportionate work.
+    fn exert(&mut self, effort: &mut isize) {
+        // If there is work to be done, ...
+        self.tidy_layers();
+        if !self.reduced() {
+            // If any merges exist, we can directly call `apply_fuel`.
+            if self.merging.iter().any(|b| b.is_double()) {
```
Kixiron, line 271:
```suggestion
            if self.merging.iter().any(B::is_double) {
```

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,840 @@
...
+
+        // Consolidated trace is empty.
+        None
+    }
+
+    // Ideally, this method acts as insertion of `batch`, even if we are not yet
+    // able to begin merging the batch. This means it is a good time to perform
+    // amortized work proportional to the size of batch.
+    fn insert(&mut self, batch: Self::Batch) {
+        assert!(batch.lower() != batch.upper());
```
Kixiron, line 310:
```suggestion
        assert_ne!(batch.lower(), batch.upper());
```

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,840 @@
...
+            if self.merging[index].is_double() {
+                return false;
+            }
+            if self.merging[index].len() > 0 {
+                non_empty += 1;
+            }
+            if non_empty > 1 {
+                return false;
+            }
+        }
```
Kixiron, line 366: This should iterate over `&self.merging` directly instead of indexing

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,840 @@
...
+    /// The `effort` parameter is that multiplier. This value should be at
+    /// least one for the merging to happen; a value of zero is not helpful.
+    pub fn with_effort(
+        mut effort: usize,
+        activator: Option<timely::scheduling::activate::Activator>,
+    ) -> Self {
+        // Zero effort is .. not smart.
+        if effort == 0 {
+            effort = 1;
+        }
```
Kixiron, line 398: Maybe we should panic if given an effort of zero, that seems like a programmer error.
Alternatively we could take a `NonZeroUsize` instead of a `usize`, I like the type-safety route better

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,840 @@
...
+    }
+
+    /// True iff the layer is a complete merge, ready for extraction.
+    fn is_complete(&self) -> bool {
+        matches!(self, MergeState::Double(MergeVariant::Complete(_)))
+    }
+
+    /// True iff the layer is an in-progress merge.
+    fn is_inprogress(&self) -> bool {
+        matches!(self, MergeState::Double(MergeVariant::InProgress(_, _, _)))
```
Kixiron, line 751:
```suggestion
        matches!(self, MergeState::Double(MergeVariant::InProgress(...)))
```

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,840 @@
...
+            panic!("Failed to complete a merge!");
+        }
+    }
+
+    /// Applies some amount of work, potentially completing the merge.
+    ///
+    /// In case the work completes, the source batches are returned.
+    /// This allows the caller to manage the released resources.
+    fn work(&mut self, fuel: &mut isize) {
+        let variant = std::mem::replace(self, MergeVariant::Complete(None));
```
Kixiron, line 828: Maybe just import `replace()` to the file, it's used a lot

src/trace/ord/val_batch.rs
```diff
@@ -0,0 +1,487 @@
...
+    <O as TryFrom<usize>>::Error: Debug,
+    <O as TryInto<usize>>::Error: Debug,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        writeln!(
+            f,
+            "description: {:?}\nlayer:\n{}",
+            self.desc,
+            textwrap::indent(&self.layer.to_string(), "    ")
+        )
```
Kixiron, line 63: `.debug_map()`?

src/trace/ord/merge_batcher.rs
```diff
@@ -0,0 +1,405 @@
...
+    fn deep_size_of_children(&self, _context: &mut deepsize::Context) -> usize {
+        self.queue.deep_size_of() + self.stash.deep_size_of()
+    }
+}
+
+impl<D: Ord, R: MonoidValue> MergeSorter<D, R> {
+    const BUFFER_SIZE_BYTES: usize = 1 << 13;
+
+    fn buffer_size() -> usize {
+        let size = ::std::mem::size_of::<(D, R)>();
```
Kixiron, line 174: Just import `size_of()`

src/trace/ord/merge_batcher.rs
```diff
@@ -0,0 +1,405 @@
...
+    pub fn finish_into(&mut self, target: &mut Vec<Vec<(D, R)>>) {
+        while self.queue.len() > 1 {
+            let list1 = self.queue.pop().unwrap();
+            let list2 = self.queue.pop().unwrap();
+            let merged = self.merge_by(list1, list2);
+            self.queue.push(merged);
+        }
+
+        if let Some(mut last) = self.queue.pop() {
+            ::std::mem::swap(&mut last, target);
```
Kixiron, line 243: Import `swap()`

src/trace/ord/merge_batcher.rs
```diff
@@ -0,0 +1,405 @@
...
+            output.push(result);
+        } else if result.capacity() > 0 {
+            self.stash.push(result);
+        }
+
+        if !head1.is_empty() {
+            let mut result = self.empty();
+            for _ in 0..head1.len() {
+                result.push(head1.pop());
+            }
```
Kixiron, line 337: Shouldn't this be an `.extend()` call or something? Plus we can pre-allocate `result`, we know its exact size

src/trace/ord/merge_batcher.rs
```diff
@@ -0,0 +1,405 @@
...
+        if !head2.is_empty() {
+            let mut result = self.empty();
+            for _ in 0..head2.len() {
+                result.push(head2.pop());
+            }
+            output.push(result);
+        }
+        while !list2.is_empty() {
+            output.push(list2.pop());
+        }
```
Kixiron, line 353: More extend, more pre-alloc

src/trace/ord/merge_batcher.rs
```diff
@@ -0,0 +1,405 @@
...
+        let mut result = 0;
+
+        for alloc in self.queue.iter() {
+            for v in alloc.iter() {
+                result += v.len();
+            }
+        }
+        for v in self.stash.iter() {
+            result += v.len();
+        }
```
Kixiron, line 368: Can be `.sum()` calls on iterators

src/trace/ord/key_batch.rs
```diff
@@ -0,0 +1,432 @@
...
+    <O as TryFrom<usize>>::Error: Debug,
+    <O as TryInto<usize>>::Error: Debug,
+{
+    type Storage = OrdKeyBatch<K, T, R, O>;
+
+    fn key<'a>(&self, storage: &'a Self::Storage) -> &'a K {
+        self.cursor.key(&storage.layer)
+    }
+    fn val<'a>(&self, _storage: &'a Self::Storage) -> &'a () {
+        unsafe { ::std::mem::transmute(&self.empty) }
```
Kixiron, line 330: This is UB, replace it with a reference to a unit literal (it's a literal so it gets static promotion)
```suggestion
        &()
```

Kixiron (changes requested):
Looks mostly ok aside from some mostly mundane stuff. There's a few big soundness issues I want addressed though, those are pretty concerning

Cargo.toml
```diff
@@ -20,6 +20,13 @@ hashbrown = "0.12.0"
 csv = { version = "1.1", optional = true }
 serde = { version = "1.0", optional = true }
 impl-trait-for-tuples = "0.2"
+deepsize = "0.2.0"
+deepsize_derive = "0.1.2"
+textwrap = "0.15.0"
```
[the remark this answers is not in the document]
ryzhyk: Agreed, we have an issue for that :)  #72

src/lattice.rs
```diff
@@ -0,0 +1,228 @@
...
+implement_lattice!(u8, 0);
+implement_lattice!(isize, 0);
+implement_lattice!(i128, 0);
+implement_lattice!(i64, 0);
+implement_lattice!(i32, 0);
+implement_lattice!(i16, 0);
+implement_lattice!(i8, 0);
+implement_lattice!((), ());
+
+impl<T: Lattice + Clone> Lattice for Antichain<T> {
```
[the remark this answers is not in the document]
ryzhyk: I stole this entire module from DD. If they didn't bother with these optimizations (or didn't want them for some reason), neither will I :)

src/operator/join.rs
```diff
@@ -152,7 +121,186 @@ where
...
+        let left =
+            self.circuit()
+                .add_binary_operator(JoinTrace::new(join_func), self, &other_trace);
+
+        fn flip_args<F, K, V1, V2, V>(f: F) -> F
+        where
+            F: Fn(&K, &V1, &V2) -> V,
+        {
+            f
+        }
```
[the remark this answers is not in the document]
ryzhyk: It flips the arguments of the join function, so that we can use the same function to do `a.join(trace(b))` and `b.join(trace(a))`.

src/operator/join.rs
```diff
@@ -166,150 +314,354 @@ where
...
+            }
+        });
+
+        //println!("new_len: {}", new_len);
+        self.output_batchers
+            .resize_with(new_len as usize, || Z::Batcher::new(()));
+        // TODO: keep output batches in `self` for allocation reuse across iterations
+        // (but potentially using more memory)..
+        let mut output_batches = Vec::new();
+        output_batches.resize_with((new_len - self.time) as usize, Vec::new);
```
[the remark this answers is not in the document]
ryzhyk: Makes sense, but I won't make this change, as this code will change completely two PRs later.

src/operator/map.rs
```diff
@@ -5,74 +5,56 @@ use crate::{
...
 where
-    SR: Clone,
+    B: Clone,
     P: Clone + 'static,
 {
     /// Apply [`MapKeys`] operator to `self`.
     ///
     /// The `map` closure takes input keys by reference.
-    pub fn map_keys<K1, K2, V, CO, F>(&self, map: F) -> Stream<Circuit<P>, CO>
+    pub fn map_keys<CO, F>(&self, map: F) -> Stream<Circuit<P>, CO>
```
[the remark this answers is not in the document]
ryzhyk: Makes sense, but let's leave it for future work.  Can you create an issue?

src/operator/map.rs
```diff
@@ -127,33 +106,39 @@ where
...
+    FO: Fn(CI::Key) -> CO::Key + 'static,
 {
     fn eval(&mut self, i: &CI) -> CO {
-        i.into_iter()
-            .map(|pair| {
-                let (k, v) = pair.into_refs();
-                ((self.map_borrowed)(k), v.clone())
-            })
-            .collect()
+        let mut batch = Vec::with_capacity(i.len());
```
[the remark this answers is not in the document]
ryzhyk: I really think that one malloc per clock cycle is not a problem. Let's worry about it if it pops up in the profile.

src/operator/trace.rs
```diff
@@ -0,0 +1,427 @@
...
+    BI::Key: Clone,
+    BI::Val: Clone,
+    BO: Batch<Key = BI::Key, Val = BI::Val, Time = TS, R = BI::R>,
+{
+    let mut builder = BO::Builder::with_capacity(timestamp.clone(), batch.len());
+    let mut cursor = batch.cursor();
+    while cursor.key_valid(batch) {
+        while cursor.val_valid(batch) {
+            let val = cursor.val(batch);
+            let key = cursor.key(batch);
```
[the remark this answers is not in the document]
ryzhyk: good catch!

src/operator/trace.rs
```diff
@@ -0,0 +1,427 @@
...
+    B: 'static,
+{
+    fn name(&self) -> Cow<'static, str> {
+        Cow::from("UntimedTraceAppend")
+    }
+    fn clock_start(&mut self, _scope: Scope) {}
+    fn clock_end(&mut self, _scope: Scope) {}
+    fn fixedpoint(&self) -> bool {
+        true
+    }
```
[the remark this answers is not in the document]
ryzhyk: This is on purpose.  I want the programmer to make an explicit decision here.

`clock_start` and `clock_end`, otoh, can safely default to empty scopes.

src/operator/trace.rs
```diff
@@ -0,0 +1,427 @@
...
+}
+
+impl<T, B> BinaryOperator<T, B, T> for UntimedTraceAppend<T, B>
+where
+    B: Batch + Clone + 'static,
+    T: Trace<Batch = Rc<B>> + 'static,
+{
+    fn eval(&mut self, _trace: &T, _batch: &B) -> T {
+        // Refuse to accept trace by reference.  This should not happen in a correctly
+        // constructed circuit.
```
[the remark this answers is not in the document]
ryzhyk: The use of `STRONGLY_PREFER_OWNED` to connect the input of this operator guarantees that we will either get an owned value or circuit construction will fail.  The latter won't happen in practice, because this operator is only instantiated by methods in this module.

tools/pre-push
ryzhyk: Any chance you could create a PR for this?

src/trace/spine_fueled.rs
ryzhyk: This requires `B::Time: Debug`

src/trace/ord/val_batch.rs
```diff
@@ -0,0 +1,487 @@
...
+        }
+        self.layer.vals.keys.truncate(write_position);
+        self.layer.vals.offs.truncate(write_position + 1);
+        self.layer.offs[self.layer.keys.len()] = O::try_from(write_position).unwrap();
+
+        // 4. Remove empty keys.
+        let mut write_position = 0;
+        for i in 0..self.layer.keys.len() {
+            let lower: usize = self.layer.offs[i].try_into().unwrap();
+            let upper: usize = self.layer.offs[i + 1].try_into().unwrap();
```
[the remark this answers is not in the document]
ryzhyk: There's always an extra element in the layer for this purpose. Overall, this is well tested code.  I wish it came with more tests though.

src/trace/ord/val_batch.rs
```diff
@@ -0,0 +1,487 @@
...
+                        to_copy = self.upper2 - self.lower2;
+                    }
+                    self.result
+                        .copy_range(&source2.layer, self.lower2, self.lower2 + to_copy);
+                    self.lower2 += to_copy;
+                }
+            }
+        }
+
+        effort = (self.result.vals.vals.vals.len() - starting_updates) as isize;
```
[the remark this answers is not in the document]
ryzhyk: turtles all the way down.

src/trace/ord/merge_batcher.rs
```diff
@@ -0,0 +1,405 @@
...
+    }
+}
+
+#[inline]
+unsafe fn push_unchecked<T>(vec: &mut Vec<T>, element: T) {
+    debug_assert!(vec.len() < vec.capacity());
+    let len = vec.len();
+    ::std::ptr::write(vec.get_unchecked_mut(len), element);
+    vec.set_len(len + 1);
+}
```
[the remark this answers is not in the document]
ryzhyk: I suspect the reason DD does it this way is to avoid bounds checking, which will still be required if we use `spare_capacity_mut`.

src/circuit/circuit_builder.rs
```diff
@@ -2149,6 +2263,10 @@ where
     O: Clone,
     Op: StrictUnaryOperator<I, O>,
 {
+    fn name(&self) -> Cow<'static, str> {
+        unsafe { &*self.operator.get() }.name()
```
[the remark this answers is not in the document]
ryzhyk: We should probably just switch to `RefCell`.  I doubt `UnsafeCell` makes any real performance difference.  I'll create an issue.

ryzhyk:
@Kixiron , thanks for the review, I implemented most suggestions.  I skipped some suggestions in modules that I stole from DD, since that code is known to work well, and we don't have a way to thoroughly test and benchmark it.

src/lattice.rs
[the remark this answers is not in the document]
Kixiron: I mean, it's probably just an oversight then, there's no real reason *not* to implement `.*_assign()` here since it allows us to immediately reuse the buffer underlying the `Antichain`

src/operator/join.rs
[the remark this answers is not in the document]
Kixiron: I don't see how it does that, it passes the closure transparently

src/trace/ord/merge_batcher.rs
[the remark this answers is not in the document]
Kixiron: Well then at very least we should use `vec.as_mut_ptr().add(len).write(element);` then because this is undefined behavior as written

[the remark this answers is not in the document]
Kixiron: Actually, `.spare_capacity_mut().get_unchecked_mut(0)` doesn't require bounds checking, either way works just fine though (see [`.spare_capacity_mut()`](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#2083-2087)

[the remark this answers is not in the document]
ryzhyk: Just curious: what makes this undefined behavior given that we don't read the uninitialized memory, but only write to it.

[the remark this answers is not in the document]
ryzhyk: > Actually, `.spare_capacity_mut().get_unchecked_mut(0)` doesn't require bounds checking, either way works just fine though (see [`.spare_capacity_mut()`](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#2083-2087))

Oh, sure, but isn't this UB in exactly the same way as the original code?

ryzhyk:
> If anything that's indication to me that we first need to thoroughly test that code before we integrate it

We built a whole system to test it, it's called DDlog ;)

Nah, just kidding, tests would be great. Let me at least add an issue for this.

src/trace/ord/merge_batcher.rs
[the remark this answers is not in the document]
Kixiron: No, `.spare_capacity_mut()` returns an `&mut [MaybeUninit<T>]` from which the `.get_unchecked_mut()` call gives us an `&mut MaybeUninit<T>` which is completely valid. The direct `.get_unchecked_mut()` call gives us an `&mut T` to uninitialized memory, which is what's ub.
The docs on `.spare_capacity_mut()` do pretty much exactly what we're doing:
```rust
// Allocate vector big enough for 10 elements.
let mut v = Vec::with_capacity(10);

// Fill in the first 3 elements.
let uninit = v.spare_capacity_mut();
uninit[0].write(0);
uninit[1].write(1);
uninit[2].write(2);

// Mark the first 3 elements of the vector as being initialized.
unsafe {
    v.set_len(3);
}

assert_eq!(&v, &[0, 1, 2]);
```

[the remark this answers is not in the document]
ryzhyk: I see, that's clever.  I'll implement the fix you suggested.

src/operator/join.rs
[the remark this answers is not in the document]
ryzhyk: Oh, now I remember, it does pass the closure transparently, but it needs to be declared as a function with explicit signature to convince the type checker.  It doesn't work in any other way.

Kixiron:
Btw the paths benchmark is broken

mbudiu-vmw (commented):
I read 10 more files and left some comments.
I still have 12 more to go.

src/operator/distinct.rs
```diff
@@ -125,86 +160,402 @@ where
...
+}
+
+impl<Z, T> DistinctTrace<Z, T>
+where
+    Z: ZSet,
+    Z::Key: Clone + Ord + PartialEq,
+    Z::R: ZRingValue,
+    T: TraceReader<Key = Z::Key, Val = (), Time = NestedTimestamp32, R = Z::R> + 'static,
+{
+    // Evaluate nested incremental distinct for a single value.
```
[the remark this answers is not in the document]
ryzhyk: I'll create an issue for this.  I also have a new implementation of `join` (next PR) that works for arbitrarily nested circuits, or so I think.  We should document and prove its correctness.

Also, for some reason, I cannot respond inline to this comment, so answering here:

> Maybe we should have two different versions of Apply then.

I agree, but prefer to wait for a use case.

src/operator/distinct.rs
```diff
@@ -125,86 +160,402 @@ where
...
+        //eprintln!("distinct_trace {}", self.time);
+
+        self.empty_input = delta.is_zero();
+
+        // Make sure we have enough room in `future_updates` to
+        // accommodate the largest timestamp in the trace, so we don't
+        // need to worry about growing `future_updates` later on.
+        let mut new_len: u32 = self.time + 1;
+        trace.map_batches(|batch| {
+            for ts in batch.upper().elements().iter() {
```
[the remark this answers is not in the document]
ryzhyk: It's not really incremental.  I just scan all the batches searching for the largest timestamp and make sure I have enough room in the array.  Since we generally don't have an upper bound of the number of iterations, we can't allocate a big enough array at initialization time.

src/operator/distinct.rs
```diff
@@ -213,28 +564,85 @@ mod test {
...
         operator::{Apply2, GeneratorNested},
+        trace::ord::OrdZSet,
+        zset,
     };
 
     #[test]
     fn distinct_incremental_nested_test() {
         let root = Root::build(move |circuit| {
             let mut inputs = vec![
+                vec![zset! { 1 => 1, 2 => 1 }, zset! { 2 => -1, 3 => 2, 4 => 2 }],
```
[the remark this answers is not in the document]
ryzhyk: Nope, the order shouldn't matter.

src/operator/join.rs
```diff
@@ -1,62 +1,43 @@
...
 {
     /// Apply [`Join`] operator to `self` and `other`.
     ///
     /// See [`Join`] operator for more info.
-    pub fn join<K, V1, V2, V, W, F, SR2, Z>(
-        &self,
-        other: &Stream<Circuit<P>, SR2>,
-        f: F,
-    ) -> Stream<Circuit<P>, Z>
+    pub fn join<F, IZ2, Z>(&self, other: &Stream<Circuit<P>, IZ2>, f: F) -> Stream<Circuit<P>, Z>
```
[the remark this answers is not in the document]
ryzhyk: It's `join` because it operates element-wise on indexed z-sets.

src/operator/trace.rs
```diff
@@ -0,0 +1,424 @@
...
+                        .cache_insert(DelayedTraceId::new(trace.local_node_id()), local);
+                    self.circuit()
+                        .cache_insert(ExportId::new(trace.local_node_id()), export);
+                    trace
+                })
+            })
+            .clone()
+    }
+
+    // TODO: this method should replace `Stream::integrate()`.
```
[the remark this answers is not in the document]
ryzhyk: Yes, but when working with z-sets the general `integrate` method is inefficient; we want to integrate them into a trace.  I'm not suggesting we remove `integrate`, but we will not use it with z-sets.

src/trace/cursor/mod.rs
```diff
@@ -0,0 +1,114 @@
+//! Traits and types for navigating order sequences of update tuples.
+//!
+//! The `Cursor` trait contains several methods for efficiently navigating
+//! ordered collections of tuples of the form `(key, val, time, diff)`. The
```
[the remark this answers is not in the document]
ryzhyk: They _are_ consolidated within a batch.  But yes, pairs and lists are not consolidated.

src/trace/cursor/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+
+/// A cursor for navigating ordered `(key, val, time, diff)` updates.
+pub trait Cursor<K, V, T, R> {
+    /// Type the cursor addresses data in.
+    type Storage;
+
+    /// Indicates if the current key is valid.
+    ///
+    /// A value of `false` indicates that the cursor has exhausted all keys.
+    fn key_valid(&self, storage: &Self::Storage) -> bool;
```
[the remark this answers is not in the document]
ryzhyk: That wouldn't be the same.  `key_valid = false` means that the _current_ key does not exist.

src/trace/cursor/cursor_pair.rs
```diff
@@ -0,0 +1,189 @@
...
+    fn map_times<L: FnMut(&T, &R)>(&mut self, storage: &Self::Storage, mut logic: L) {
+        if self.key_order == Ordering::Less
+            || (self.key_order == Ordering::Equal && self.val_order != Ordering::Greater)
+        {
+            self.cursor1.map_times(&storage.0, |t, d| logic(t, d));
+        }
+        if self.key_order == Ordering::Greater
+            || (self.key_order == Ordering::Equal && self.val_order != Ordering::Less)
+        {
+            self.cursor2.map_times(&storage.1, |t, d| logic(t, d));
```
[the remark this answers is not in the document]
ryzhyk: yes, that's what we want here

src/trace/cursor/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Returns a reference to the current value, if valid.
+    fn get_val<'a>(&self, storage: &'a Self::Storage) -> Option<&'a V> {
+        if self.val_valid(storage) {
+            Some(self.val(storage))
+        } else {
+            None
+        }
+    }
+
+    /// Applies `logic` to each pair of time and difference. Intended for
```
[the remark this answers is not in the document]
ryzhyk: > Is this done for all keys and values or just for the current key, etc?

It maps the closure over all (time, weight) pairs for the current key and value

mbudiu-vmw (commented):
I have 5 more files to read.

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,841 @@
...
+    /// to correctly install the result.
+    fn work(&mut self, fuel: &mut isize) {
+        // We only perform work for merges in progress.
+        if let MergeState::Double(layer) = self {
+            layer.work(fuel)
+        }
+    }
+
+    /// Extract the merge state, typically temporarily.
+    fn take(&mut self) -> Self {
```
mbudiu-vmw, line 768: this function has a strange comment and name

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,841 @@
...
+    ///
+    /// Either batch may be `None` which corresponds to a structurally
+    /// empty batch whose upper and lower froniers are equal. This
+    /// option exists purely for bookkeeping purposes, and no computation
+    /// is performed to merge the two batches.
+    fn begin_merge(batch1: Option<B>, batch2: Option<B>) -> MergeState<B> {
+        let variant = match (batch1, batch2) {
+            (Some(batch1), Some(batch2)) => {
+                // Leonid: we do not require batch bounds to grow monotonically.
+                //assert!(batch1.upper() == batch2.lower());
```
mbudiu-vmw, line 787: so this does not break anything?

src/trace/spine_fueled.rs
```diff
@@ -0,0 +1,841 @@
+//! An append-only collection of update batches.
```
mbudiu-vmw, line 1: so does dbsp also this delayed work?

src/trace/ord/zset_batch.rs
```diff
@@ -0,0 +1,391 @@
...
+
+/// An immutable collection of `(key, weight)` pairs without timing information.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct OrdZSet<K, R>
+where
+    K: Ord,
+{
+    /// Where all the dataz is.
+    pub layer: OrderedLeaf<K, R>,
+    pub desc: Description<()>,
```
mbudiu-vmw, line 36: this is necessary to fulfill some traits?

src/trace/ord/indexed_zset_batch.rs
```diff
@@ -0,0 +1,490 @@
...
+where
+    K: Ord,
+    V: Ord,
+    R: Clone,
+    O: OrdOffset,
+    <O as TryFrom<usize>>::Error: Debug,
+    <O as TryInto<usize>>::Error: Debug,
+{
+    /// Where all the dataz is.
+    pub layer: OrderedLayer<K, OrderedLeaf<V, R>, O>,
```
mbudiu-vmw, line 42: is this logically OrederedLayer<K, OrderedZSet<V, R>, O>?

mbudiu-vmw (commented):
Now I am done.

Issue #68, opened 2022-05-31
Implement `BatchReader` trait for traces.


ryzhyk:
### Is your feature request related to a problem? Please describe.

Similar to `BatchReader`, the `TraceReader` API exposes an iterator over all elements in the trace. The two APIs can be reconciled with a bit of work.  This way, any operator that works on batches would also work on traces, e.g., `Join` currently takes a batch as input.  When using `Join` as part of `join_incremental` computation, one of the input batches is an integral of the input stream.  With this change we will be able to switch to integrating streams of (indexed) Z-sets into a trace instead of a batch and everything would still work.

### Describe the solution you'd like

-

PR #75, opened 2022-06-01
Unify BatchReader and TraceReader traits.

ryzhyk:
This PR resolves issue #68.


src/operator/aggregate.rs
```diff
@@ -178,19 +183,32 @@ where
...
+    Z: IndexedZSet<R = O::R> + 'static,
+    F: Fn(&Z::Key, &mut [(&Z::Val, Z::R)]) -> O::Key + 'static,
     O: Clone + ZSet + 'static,
     O::R: ZRingValue,
 {
-    fn eval(&mut self, i: &I) -> O {
+    fn eval(&mut self, i: &Z) -> O {
         let mut elements = Vec::with_capacity(i.len());
         let mut cursor = i.cursor();
+        let mut vals: Vec<(&Z::Val, Z::R)> = Vec::new();
```
Kixiron, line 196: Should we pre-allocate this with the number of keys or values?

src/operator/aggregate.rs
```diff
@@ -314,18 +341,13 @@ mod test {
...
                         let mut result: isize = 0;
 
-                        while cursor.val_valid(storage) {
-                            let v = cursor.val(storage);
-                            let w = cursor.weight(storage);
-                            result += (*v as isize) * w;
-                            cursor.step_val(storage);
+                        for (v, w) in vals.into_iter() {
+                            result += (**v as isize) * (*w);
                         }
```
Kixiron, line 349: Github won't let me make a suggestion, but here it is
```rust
let result = vals.iter().map(|&(&v, w)| v * w).sum();
```

src/operator/aggregate.rs
```diff
@@ -382,20 +404,16 @@ mod test {
...
-                     -> (usize, usize) {
+                    let min = |key: &usize, vals: &mut [(&usize, isize)]| -> (usize, usize) {
                         let mut result = usize::MAX;
 
-                        while cursor.val_valid(storage) {
-                            let v = cursor.key(storage);
-                            if v < &result {
+                        for (v, _) in vals.into_iter() {
+                            if *v < &result {
                                 result = v.clone();
```
Kixiron, line 412:
```suggestion
                        for &(&v, _) in vals.iter() {
                            if v < result {
                                result = v;
```

src/trace/ord/val_batch.rs
```diff
@@ -393,12 +393,17 @@ where
             self.cursor.child.child.step(&storage.layer.vals.vals);
         }
     }
-    fn weight<'a>(&self, storage: &'a Self::Storage) -> &'a R
+    fn weight(&mut self, storage: &Self::Storage) -> R
```
Kixiron, line 396: Looks like we clone within pretty much every `.weight()` call, maybe it should return a reference and allow the caller to decide whether or not it should be cloned

src/operator/aggregate.rs
```diff
@@ -178,19 +183,32 @@ where
...
+                let w = cursor.weight(i);
+                // Skip values with weight zero.
+                if !w.is_zero() {
+                    vals.push((cursor.val(i), w));
+                }
+                cursor.step_val(i);
+            }
+            // Skip keys that only contain values with weight zero.
+            if !vals.is_empty() {
+                elements.push((((self.agg_func)(cursor.key(i), &mut vals), ()), Z::R::one()));
```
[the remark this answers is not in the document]
ryzhyk: That might actually work, thanks for the idea.

src/operator/aggregate.rs
ryzhyk: At this point we have no insight about the average/max size of a group. But I guess there's not much harm in preallocating `delta.len()` elements.

src/trace/ord/val_batch.rs
ryzhyk: Problem is, the result sometimes needs to be computed, but I can probably make it work...

ryzhyk: Actually, no, not in a nice way.

PR #81, opened 2022-06-13
Start building infra for arbitrarily nested circuits.

ryzhyk:
This PR makes progress towards building circuits with more than one level of nesting.  These don't occur very often, but for example efficient implementation of some graph [algorithms](https://github.com/TimelyDataflow/differential-dataflow/tree/master/src/algorithms/graphs) like strongly connected components requires an extra level of nested circuits.

Specifically, we add a `Product` type that allows creating arbitrarily nested clocks.  We also generalize the `join_trace` operator to work for any nesting depth.

Future PRs will generalize `distinct` and `aggregate`.

### DIFF
--- a/src/circuit/operator_traits.rs
+++ b/src/circuit/operator_traits.rs
@@ -138,14 +138,33 @@ pub trait Operator: 'static {
     ///
     /// This method is invoked as part of checking if the circuit has reached a
     /// fixed point state, i.e., a state where the outputs of all operators will
-    /// remain constant forever
+    /// remain constant until the end of the current clock epoch
     /// (see [`Circuit::fixedpoint`](`crate::circuit::Circuit::fixedpoint`)).
     ///
-    /// The method is invoked at the end of a clock cycle, after
-    /// the operator has been evaluated.  It returns `true` if the operator's
-    /// output is guaranteed to remain constant (i.e., all future outputs
-    /// will be equal to the last output) as long as its inputs remain
-    /// constant.
+    /// It returns `true` if the operator's output is guaranteed to remain
+    /// constant (i.e., all future outputs will be equal to the last output) as
+    /// long as its inputs remain constant.
+    ///
+    /// The exact semantics depends on the value of the `scope` argument, which
+    /// identifies the circuit whose fixed point state is being checked.
+    /// Scope 0 is the local circuit.  The method is invoked with `scope=0`
+    /// at the end of a clock cycle, and should return `true` if, assuming that
+    /// it will see inputs identical to the last input during all future clock
+    /// cycles in the current clock epoch, it will keep producing the same
+    /// outputs.
+    ///
+    /// Scope 1 represents the parent of the local circuit.  The method is
+    /// invoked with `scope=1` at the end of a clock _epoch_, and should
+    /// return `true` if, assuming that it will see a sequence of inputs
+    /// (aka the input stream) identical to the last epoch during all future
+    /// epochs, it will keep producing the same output streams.
+    ///
+    /// Scope 2 represents the grandparent of the local circuit.  The method is
+    /// invoked with `scope=2` at the end of the parent clock _epoch_, and
+    /// checks that the operator's output will remain stable wrt to the
+    /// nested input stream (i.e., stream of streams).
+    ///
+    /// And so on.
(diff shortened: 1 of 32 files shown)


src/operator/join.rs
```diff
@@ -793,107 +653,217 @@ mod test {
...
+    impl Display for Label {
+        fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+            write!(f, "L({},{})", self.0, self.1)
+        }
+    }
+
+    #[derive(Clone, Debug, Default, Ord, PartialOrd, Eq, PartialEq, DeepSizeOf)]
+    struct Edge(pub usize, pub usize);
+
+    impl Display for Edge {
```
mbudiu-vmw, line 688: Maybe these should be library functions, I expect we'll have more graph tests.

mbudiu-vmw (approved):
This looks very exciting if it works.

src/circuit/operator_traits.rs
```diff
@@ -138,14 +138,33 @@ pub trait Operator: 'static {
     ///
     /// This method is invoked as part of checking if the circuit has reached a
     /// fixed point state, i.e., a state where the outputs of all operators will
-    /// remain constant forever
+    /// remain constant until the end of the current clock epoch
```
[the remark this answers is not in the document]
ryzhyk: that's right

src/operator/join.rs
ryzhyk: Yes, we'll have a graph library, but this implementation of `propagate` isn't actually correct because it doesn't apply `distinct` on each iteration.  This is done deliberately for testing purposes, so it won't be reusable as a library function.  As for the types, they'd have to be generic to be reusable.

src/time/mod.rs
```diff
@@ -104,122 +124,36 @@ pub trait Timestamp: PartialOrder + Clone + Ord + PartialEq + Eq + 'static {
...
-                panic!("NestedTimestamp32::recede timestamp underflow");
-            }
-            Self(self.0 - 1)
-        } else if scope == 1 {
-            Self(self.0 & 0x7fffffff)
-        } else {
-            self.clone()
-        }
+    fn recede(&self, _scope: Scope) -> Self {
+        self - 1
```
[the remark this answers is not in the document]
ryzhyk: It will crash on overflow. But I have a TODO to have this return `Option`.  This will be done as part of moving time management into the circuit.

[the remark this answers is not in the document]
Kixiron: Over/underflow checks only happen when `overflow-checks = true`, which is by default just debug mode so in release mode this will just underflow

src/time/nested_ts32.rs
```diff
@@ -0,0 +1,123 @@
...
+/// distinguishes the latest parent clock cycle, or "epoch", (higher-order bit
+/// set to `1`) from all previous epochs (higher order bit is `0`).
+#[derive(Clone, DeepSizeOf, Default, Eq, PartialEq, Debug, Hash, PartialOrd, Ord)]
+pub struct NestedTimestamp32(u32);
+
+impl Display for NestedTimestamp32 {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(
+            f,
+            "({},{})",
```
Kixiron, line 19:
```suggestion
            "({}, {})",
```

src/time/nested_ts32.rs
```diff
@@ -0,0 +1,123 @@
...
+            if self.epoch() { "new" } else { "old" },
+            self.inner()
```
Kixiron, line 21:
```suggestion
            self.inner(),
```

src/time/nested_ts32.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    #[inline]
+    pub fn epoch(&self) -> bool {
+        self.0 >> 31 == 1
+    }
+
+    #[inline]
+    pub fn inner(&self) -> u32 {
+        self.0 & 0x7fffffff
```
Kixiron, line 42: I'd like some named constants for these magic numbers

Kixiron (approved):

PR #82, opened 2022-06-15
Use blanket implementation for HasZero and HasOne.

absoludity:

Note: I don't expect this to be landed (or reviewed for that matter), just exploring the code.

I was browsing the code and started in this module. Noticed that later trait implementations in the module (`AddByRef`, `NegByRef`) were already using blanket implementations, so out of interest, updated `HasZero` and `HasOne` to do so, rather than explicitly doing so for a limited set of integer types (ie. wasn't implemented for `BigInt` or `BigUInt` I don't think? Whereas with the blanket implementation it is).

The one issue was that there was already an explicit implementation of `HasZero` for the `Trie` trait which the compiler detects as a conflict (since `Trie` may be updated in the future to implement `num::traits::zero`). I was going to move that to be a default implementation of the `Trie` trait itself, with something like:

```diff
diff --git a/src/trace/layers/mod.rs b/src/trace/layers/mod.rs
index 92e97c5..04420aa 100644
--- a/src/trace/layers/mod.rs
+++ b/src/trace/layers/mod.rs
@@ -57,6 +57,13 @@ pub trait Trie: Sized {
         merger.push_merge((self, self.cursor()), (other, other.cursor()));
         merger.done()
     }
+
+    fn is_zero(&self) -> bool {
+        self.is_empty()
+    }
+    fn zero() -> Self {
+        <Self as Trie>::TupleBuilder::new().done()
+    }
 }
 
 pub struct TrieSlice<'a, T: Trie>(&'a T, T::Cursor);
```

but as no tests failed when I simply deleted the implementation, I'm not sure if it's actually used, so just left it :)

Other than that, just a couple of typos.

### DIFF
--- a/src/algebra/mod.rs
+++ b/src/algebra/mod.rs
@@ -84,7 +84,8 @@ pub trait AddAssignByRef {
     fn add_assign_by_ref(&mut self, other: &Self);
 }
 
-/// Implementation of AddAssignByRef for types that already have `AddAssign<&T>`.
+/// Implementation of AddAssignByRef for types that already have
+/// `AddAssign<&T>`.
 impl<T> AddAssignByRef for T
 where
     for<'a> T: AddAssign<&'a T>,
--- a/src/algebra/mod.rs
+++ b/src/algebra/mod.rs
@@ -1,9 +1,7 @@
 //! This module contains declarations of abstract algebraic concepts:
 //! monoids, groups, rings, etc.
 
-use std::{
-    ops::{Add, AddAssign, Mul, Neg},
-};
+use std::ops::{Add, AddAssign, Mul, Neg};
 
 #[macro_use]
 mod checked_int;
@@ -207,6 +205,7 @@ where
 #[cfg(test)]
 mod integer_ring_tests {
     use super::*;
+    use num::bigint::BigInt;
 
     #[test]
     fn fixed_integer_tests() {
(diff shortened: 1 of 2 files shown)


ryzhyk (commented):
Thanks @absoludity!

I'm a bit reluctant to implement this change.  It means that `HasZero` is defined for those and only those types that implement `num::traits::Zero` (You discovered this with the `HasZero` implementation for `Trace`).  And so we can just ditch `HasZero` altogether and use `Zero` instead.  Looks like this would actually work right now, so maybe it is the right thing to do. The two reasons we defined our own trait are (1) `Zero` implies `Add`, while `HasZero` doesn't (2) We want to be able to implement `HasZero` for library types like `Rc<>`, for which `Zero` is not defined.  I'm not sure (1) is a real problem.  As for (2), while it looks like we no longer rely on `impl HasZero for Rc`, we might very well need it in the future, and having our own trait will give us the flexibility to do so.

Does this make sense?

Kixiron:
I think orphan rules should allow for us to still implement `HasZero for Rc<T> where T: HasZero` so that shouldn't be a problem I don't think

Kixiron:
Honestly I'm not a huge fan of the `num` dependency at all, so I'd be happy if it got removed entirely.
We are missing some implementations for `&dyn HasZero`, `Box<T>` and `&T` though now that I look at it, we should add those sometime

absoludity:
> Thanks @absoludity!
...
> Does this make sense?

Yep, makes sense. As above, I'd just started reading the code and wondering why there's blanket implementations for `AddByRef` and `NegByRef` but not for `HasZero`. What wasn't immediately clear to me (until I was forced to remove the implementation for `Trace`/`Trie`) was that adding a blanket implementation for all types that already implement `Zero` also means that adding an implementation of `HasZero` to any other type will then be a conflict, even if though that type does not implement `Zero`, because the type may do so in the future.

> Honestly I'm not a huge fan of the `num` dependency at all, so I'd be happy if it got removed entirely.

Yeah, sounds like that's the better direction.

ryzhyk:
@absoludity, yeah, it's an annoying quirk of Rust.  Hopefully one day we will have support for [specialization](https://rust-lang.github.io/rfcs/1210-impl-specialization.html).

PR #84, opened 2022-06-18
Added tests for merge_batcher

Kixiron:


Kixiron:
@ryzhyk I'm really worried about the `RefCell` stuff that's going on in `src/trace/spine_fueled.rs`, that stuff is *super* sketchy

PR #86, opened 2022-06-22
Add more tests to merge_batcher and fix `MergeSorter`

Kixiron:
* Adds property testing to merge_batcher and consolidation
* Fixes some incorrect behavior of `MergeSorter`
  * Holding empty batches
  * Holding unsorted batches
  * Producing unsorted results
* Adds some benchmarks to consolidation
* Adds a *healthy* amount of debug assertions into `MergeSorter`'s code


src/trace/ord/merge_batcher/mod.rs
```diff
@@ -270,51 +312,85 @@ impl<D: Ord, R: MonoidValue> MergeSorter<D, R> {
...
 
-                head2 = list2.pop_front().map_or_else(VecDeque::new, VecDeque::from);
+                // TODO: We could also attempt to pop from list1 here
+                head2 = list2
+                    .pop()
+                    // .or_else(|| list1.pop())
+                    .map_or_else(VecDeque::new, VecDeque::from);
+            }
+
+            // If result is empty we have to start a new result vec to make sure
```
ryzhyk, line 352: I don't think this is right.  The invariant that the original code appears to maintain is that not just each individual batch in `list1` and `list2`, but the concatenation of batches is sorted.  If that invariant holds, then it's totally fine to continue pushing to the same result vector.

src/trace/ord/merge_batcher/mod.rs
```diff
@@ -172,22 +187,30 @@ impl<D: Ord, R: MonoidValue> MergeSorter<D, R> {
...
-                    self.queue.pop().expect("there's at least two batches"),
-                    self.queue.pop().expect("there's at least two batches"),
-                );
-
-                // Merge the two batches together
-                let merged = self.merge_by(left, right);
-                self.queue.push(merged);
+            if !batch.is_empty() {
+                if let Some(last) = self.queue.last_mut() {
+                    last.push(batch);
```
ryzhyk, line 192: This violates the invariant that the entire list of batches must be ordered, not just individual batches. This is why the original code pushes a single batch: `vec![batch]`

ryzhyk (changes requested):
The new testing infra is awesome, but I think there's something wrong with the actual bug fix.  It seems to change how the whole data structure works.

src/trace/ord/merge_batcher/mod.rs
Kixiron: It's not, take this situation:
```
result = [1, 2, 3, 4]
list2 = [10]

// pop a new list1
list1 = [1]
```
The next iteration will push `1` to `result` which results in `[1, 2, 3, 4, 1]` which is incorrect

rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+wrap_comments = true
```
ryzhyk, line 1: Nice! Let's remove these config options from `tools/pre-push` and `.github/workflows/main.yml`.

src/trace/ord/merge_batcher/mod.rs
Kixiron: So, I applied your suggested fix but I don't think it's correct, now the assertion that `debug_assert!(last <= &merged.0);` gets set off

Kixiron:
@ryzhyk Does this all look good to you?

src/trace/ord/merge_batcher/mod.rs
```diff
@@ -196,42 +220,68 @@ impl<D: Ord, R: MonoidValue> MergeSorter<D, R> {
...
-    fn merge_by(&mut self, list1: Vec<Vec<(D, R)>>, list2: Vec<Vec<(D, R)>>) -> Vec<Vec<(D, R)>> {
+    // TODO: Split this into two functions:
+    //       - `fn merge_batch(Vec<(D, R)>, Vec<(D, R)>) -> Vec<(D, R)>`
+    //       - `fn merge_batches(Vec<Vec<(D, R)>>, Vec<Vec<(D, R)>>) -> Vec<Vec<(D,
+    //         R)>>`
+    // TODO: There's some things that need to be evaluated/tested/benchmarked here:
+    //       - When getting a new head1/head2 after the inner merge happens, is it
+    //         beneficial to pop from the opposite list (list1/list2) when one of
+    //         them is exhausted?
+    //       - Sometimes we can keep on merging into the same results vec when one
```
ryzhyk, line 250: Is this bullet point still valid?  I think we keep pushing to result even as we move `head` to the next element in the corresponding list.

src/trace/ord/merge_batcher/mod.rs
```diff
@@ -196,42 +220,68 @@ impl<D: Ord, R: MonoidValue> MergeSorter<D, R> {
...
-            self.queue.push(merged);
+            if !merged.is_empty() {
+                self.queue.push(merged);
+            }
         }
 
         if let Some(mut last) = self.queue.pop() {
-            swap(&mut last, target);
+            // TODO: Reuse the `last` buffer somehow
+            if target.capacity() == 0 {
```
ryzhyk, line 232: Why is this implementation better that the original?

src/trace/ord/merge_batcher/mod.rs
```diff
@@ -268,51 +319,79 @@ impl<D: Ord, R: MonoidValue> MergeSorter<D, R> {
...
             if head2.is_empty() {
-                if head2.capacity() == Self::BUFFER_ELEMENTS {
+                if head2.capacity() >= Self::BUFFER_ELEMENTS {
                     self.stash.push(Vec::from(head2));
                 }
 
                 head2 = list2.pop_front().map_or_else(VecDeque::new, VecDeque::from);
             }
+
+            // // If result is empty we have to start a new result vec to make
```
ryzhyk, line 356: Remove comment?

ryzhyk (approved):
Nice testing infra!

src/trace/ord/merge_batcher/mod.rs
Kixiron: It takes advantage of any capacity the target vec has

Kixiron: Eh nvm, I'll remove it

ryzhyk (approved):
I already approved :)

PR #89, opened 2022-06-26
Added project overview.

ryzhyk:
I want to create a crates.io release to make it easy for people to try DBSP.  In preparation for that I added a project description to the README.  Once the crate has been published, I will add links to crates.io and docs.rs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](https://opensource.org/licenses/MIT)
+[![CI workflow](https://github.com/vmware/database-stream-processor/actions/workflows/main.yml/badge.svg)](https://github.com/vmware/database-stream-processor/actions)
 [![codecov](https://codecov.io/gh/vmware/database-stream-processor/branch/main/graph/badge.svg?token=0wZcmD11gt)](https://codecov.io/gh/vmware/database-stream-processor)
 
 # Database Stream Processor
--- a/README.md
+++ b/README.md
@@ -4,37 +4,10 @@
 
 Streaming and Incremental Computation Framework
 
-## Development tools
-
-Ideally this code should run fine in Linux, MacOs, and Windows.
-The code is written in Rust.  Here are some tools we found useful for development:
-
-See <https://www.twelve21.io/getting-started-with-rust-on-windows-and-visual-studio-code> about
-using Rust in VSCode.
-
-* Rust `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
-* Tools for Rust
-  * tarpaulin for code coverage (Linux only): `cargo install cargo-tarpaulin`
-* Vscode: <https://code.visualstudio.com/Download>
-* Vscode extensions for Rust:
-  * crates <https://marketplace.visualstudio.com/items?itemName=serayuzgur.crates>
-  * Rust <https://marketplace.visualstudio.com/items?itemName=rust-lang.rust>
-  * Even Better TOML <https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml>
-  * Power Header <https://marketplace.visualstudio.com/items?itemName=epivision.vscode-file-header>
-    (Enables you to insert a copyright header in VSCode by pressing CTRL+ALT+H)
-  * Rust Test Explorer <https://marketplace.visualstudio.com/items?itemName=swellaby.vscode-rust-test-adapter>
-  * Rust debugger CodeLLDB <https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb>
-  * Coverage Gutters <https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters>
-
-## Check test coverage
-
-* Run on the command line `cargo tarpaulin --out Lcov` to generate coverage information
-* Use the Coverage Gutters VSCode extension to watch the lcov file (CTRL+SHIFT+P, Coverage Gutters: Watch)
-
-## Set-up git hooks
+## Setting up git hooks
 
 * Execute the following command to make `git commit` check the code before commit:
 ```
 GITDIR=$(git rev-parse --git-dir)
-ln -sf $(pwd)/tools/prepush.sh ${GITDIR}/hooks/pre-push
+ln -sf $(pwd)/tools/pre-push ${GITDIR}/hooks/pre-push
 ```
--- a/README.md
+++ b/README.md
@@ -4,11 +4,87 @@
 
 # Database Stream Processor
 
-Streaming and Incremental Computation Framework
+Database Stream Processor (DBSP) is a framework for computing over data streams
+that aims to be more expressive and performant than existing streaming engines.
+
+## Why DBSP?
+
+Computing over streaming data is hard.  Streaming computations operate over
+changing inputs and must continuously update their outputs as new inputs arrive.
+They must do so in real time, processing new data and producing outputs with a
+bounded delay.
+
+Existing streaming engines have limited expressive power, which is why their use
+in the modern data stack is restricted to data preparation and visualization.
+Complex business intelligence tasks are left to traditional database systems
+where they run as periodic (e.g., daily) batch jobs.  This is inadequate in the
+world that increasingly relies on real-time alerts, recommendations, control, etc.
+
+DBSP intends to change the situation by offering a simple promise: to **execute
+any query workload that can run in batch mode in streaming mode**.
+
+## Overview
+
+DBSP unifies two kinds of streaming data: time series data and change data.
+
+- **Time series data** can be thought of as an infinitely growing log indexed by
+  time.
+
+- **Change data** represents updates (insertions, deletions, modifications) to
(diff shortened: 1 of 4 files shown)


Kixiron:
I actually just encountered a funky error in the wild with cargo trying to fetch the spec submodule but me not having proper creds to get it, leaving me unable to use the repo. To fix that I think we should add [`exclude = ["doc/spec"]`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) to our `Cargo.toml` to prevent publishing that as part of our crate.

Additionally, our exports really need to be refined. When opening the crate with `cargo doc --open` it's really really chaotic and hard to figure out what's relevant. We've got macros that maybe shouldn't be exported, no docs, none of the entry-points into the crate are exported (`Root`, `Circuit`, etc.) and it's just generally hard to figure out what's going on
image

I'd also like to ditch our timely/abomonation dependencies before we publish because otherwise we'll set off every dependency scan in the galaxy because of how cursed abomonation is.

We're also missing a lotta docs on a lotta very critical things like `.index()`, `.join()`, etc. (the docs on `.join_trace()` are also incomplete, they end with `by first assembling traces of both streams:` with nothing after it)

A bunch of names are also fairly unintuitive, `.delta0()` makes more sense as `.delay_one()` or something.

We also just need more examples, without concrete examples or an understanding of the underlying theory there's almost no discernible difference when looking at `.join()` vs. `.join_incremental()`. I still think we need different flavors of scopes to make it easier to use incremental vs. non-incremental operators so that you can only use incremental operators within incremental scopes.

We can also probably write a good, long doc piece in our readme and then use `#![doc(include="../README.md")]` at the top of our `lib.rs` to make those part of our crate docs

gz:
It would be nice to have a tutorial with some more explanations aside from the spec PDF. Reading [this](https://github.com/vmware/database-stream-processor/blob/main/benches/galen.rs#L157) it seems to be fairly complex?

Having a tutorial with a simple program that's incrementally (pun intended) built and explained would be nice as you add complexity, e.g., start with a join/filter/smth else and make it more complex as you explain things...

README.md
```diff
@@ -1,40 +1,93 @@
...
+one record at a time.
+
+1. **Windowing operators** that group time series data into time windows,
+including various forms of tumbling and sliding windows, windows driven by
+watermarks, etc.
+
+1. The complete set of **relational operators**: select, project, join,
+aggregate, etc.
+
+1. **Recursion**: Recursive queries allow for instance expressing graph
```
gz, line 41: Explanation goes right into example (graph queries) rather than giving a general explanation like you do for {window, per-record} operators.

gz (approved):

ryzhyk:
The PR is only for the readme. The public API is currently non existent. I'll work on it next, but there are some challenges there, it's not just a matter of cleaning the docs. I don't think this should stop us from releasing on crates.io, but if people feel strongly about it, we can do it later.

Kixiron:
Gerd does bring up a good point, I didn't think about reserving the namespace

ryzhyk:
> I actually just encountered a funky error in the wild with cargo trying to fetch the spec submodule but me not having proper creds to get it, leaving me unable to use the repo. To fix that I think we should add [`exclude = ["doc/spec"]`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) to our `Cargo.toml` to prevent publishing that as part of our crate.

Seems like `exclude` won't do it: https://github.com/rust-lang/cargo/issues/4247#issuecomment-959912231

ryzhyk:
> > I actually just encountered a funky error in the wild with cargo trying to fetch the spec submodule but me not having proper creds to get it, leaving me unable to use the repo. To fix that I think we should add [`exclude = ["doc/spec"]`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) to our `Cargo.toml` to prevent publishing that as part of our crate.
> 
> Seems like `exclude` won't do it: [rust-lang/cargo#4247 (comment)](https://github.com/rust-lang/cargo/issues/4247#issuecomment-959912231)

@mbudiu-vmw , is there a way the paper could live in a public repo?

mbudiu-vmw:
The pdf is already in this repo.
Apparently overleaf cannot make the underlying repository public, so we have to copy the sources.

ryzhyk:
> The pdf is already in this repo. Apparently overleaf cannot make the underlying repository public, so we have to copy the sources.

I think there is a way overleaf can work with an existing git repo. But I'm also ok with removing the submodule and just keeping the PDF, which is the important part.  Are ok with that?

mbudiu-vmw:
I just made some edits to the sources. We should have the tex sources too.

mbudiu-vmw:
If you can send Val instructions on how to make the repo public perhaps he can do it.

ryzhyk:
> If you can send Val instructions on how to make the repo public perhaps he can do it.

I don't know how to do it.  More importantly, we don't want to depend on Val's repo being available.  If he changes the visibility or deletes the repo in the future (or overleaf deletes or closes it automatically for whatever reason), it will break the crate.

mbudiu-vmw:
Then we should just copy the sources here as well. We need a backup anyway.

README.md
```diff
@@ -1,40 +1,92 @@
...
-Streaming and Incremental Computation Framework
+Database Stream Processor (DBSP) is a framework for computing over data streams
+that aims to be more expressive and performant than existing streaming engines.
 
-## Development tools
+## Why DBSP?
 
-Ideally this code should run fine in Linux, MacOs, and Windows.
-The code is written in Rust.  Here are some tools we found useful for development:
+Computing over streaming data is hard.  Streaming computations operate over
```
mbudiu-vmw, line 12: the word "incremental" does not appear here at all.

README.md
```diff
@@ -1,40 +1,92 @@
...
-## Check test coverage
+## Overview
 
-* Run on the command line `cargo tarpaulin --out Lcov` to generate coverage information
-* Use the Coverage Gutters VSCode extension to watch the lcov file (CTRL+SHIFT+P, Coverage Gutters: Watch)
+DBSP queries are composed of the following classes of operators (in the order of
+increasing complexity):
 
-## Set-up git hooks
+1. **Per-record operators** that parse, validate, filter, transform data streams
```
mbudiu-vmw, line 31: the notion of "record" is not defined. 
I would first introduce the notion of "transaction".
Once we merge this I will add some diagrams.

README.md
```diff
@@ -1,40 +1,92 @@
...
+one record at a time.
+
+1. **Windowing operators** that group time series data into time windows,
+including various forms of tumbling and sliding windows, windows driven by
+watermarks, etc.
+
+1. The complete set of **relational operators**: select, project, join,
```
mbudiu-vmw, line 38: in order of increasing complexity this comes first.

README.md
```diff
@@ -1,40 +1,92 @@
...
+aggregate, etc.
+
+1. **Recursion**: Recursive queries express iterative computations, e.g.,
```
mbudiu-vmw, line 41: the word "incrementally" appears here for the first time.
Perhaps fixpoint is a more accurate description, with the statement that fixpoint computations can be used to implement recursive queries.

README.md
```diff
@@ -1,40 +1,92 @@
...
+partitioning a graph into strongly connected components.  Like all DPSP queries,
+recursive queries update their outputs incrementally as new data arrives.
+
+DBSP unifies two kinds of streaming data: time series data and change data.
```
mbudiu-vmw, line 45: this paragraph should come first.

README.md
```diff
@@ -1,40 +1,92 @@
...
+
+- **Time series data** can be thought of as an infinitely growing log indexed by
+  time.
+
+- **Change data** represents updates (insertions, deletions, modifications) to
+  some state modeled as a table of records.
+
+In DBSP, a time series is just a table where records are only ever added and
```
mbudiu-vmw, line 53: "table" appears here for the first time.

README.md
```diff
@@ -1,40 +1,92 @@
...
+never removed or modified.  As a result, this table can grow unboundedly; hence
+most queries work with subsets of the table within a bounded time window.  DBSP
+does not need to wait for all data within a window to become available before
+evaluating a query (although the user may choose to do so): like all queries,
+time window queries are updated on the fly as new inputs become available.  This
+means that DBSP can work with arbitrarily large windows as long as they fit
+within available storage.  All other operators listed above apply to both time
+series and change data.
```
mbudiu-vmw, line 61: Perhaps I could take a stab at improving this documentation, after all I have already tons of slides I could steal material from. But I understand that this text should be more developer-oriented.

mbudiu-vmw (approved):

README.md
ryzhyk: I assume that most people know what records and tables are.

README.md
ryzhyk: "Recursion" is a more familiar term to database folks than "fixed point".  The goal here is not to write a formal design document, precisely specifying each notion, but to give an idea of the capabilities we are implementing.

ryzhyk:
I think this README is an improvement over what we have now.  I am going to merge it and move on to other stuff. Improvements are welcome.

Issue #92, opened 2022-06-30
RFC: Operator API design


ryzhyk:
## RFC: Operator API design (so far just a list of problems with the current design)

The current API for assembling operators into circuits has poor ergonomics.
This is partially because it is intentionally low-level, designed to support maximal
flexibility as opposed to ease of use, and partially because it was built in
a bottom-up fashion, with little consideration for user experience.
My goal is to cleanup the design where possible and also hide some low-level details
behind a less flexible but more user-friendly higher-level API.

Below is a summary of problems along with some ideas for addressing them, but first
I mention the aspects of the current design that work well and that I'd like to
preserve going forward.

## The good

- The functional style of instantiating operators, similar to DD, LINQ, and many other
  systems is nice compared to manually instantiating operators and connecting their
  outputs (the `stream.map().join().distinct()` stuff).

- The operator caching feature allows creating high-level operators without introducing
  a performance hit, e.g., `stream.distinct_trace()` internally calls `stream.trace()`.
  If another operator in the circuit calls `stream.trace()`, it will get a reference to
  the same trace rather than creating a new one.  

## The bad

1. **API requires type annotations.** Most operators are generic over return type, e.g.,
   happy to return any type that implements `trait ZSet`.  The goal is to support multiple
   ZSet implementations (e.g., in-memory and disk-backed), but the downside is that the
   programmer often needs to explicitly specify the return type using type annotations.
   - [x] Proposed solution: PR #94 

1. **Operator naming.**  Operator names are currently a complete mess.  Names like
   `join_trace` for example make zero sense to the user.  Ignoring the current lame naming
   scheme, a sensible naming convention is not easy to some up with because DBSP
   supports multiple flavors of the same operator.  Linear operators (map, filter, index,
   etc.) are easy, as the same implementation works in incremental and non-incremental cases
   and for arbitrary nesting depth.  `join`, `aggregate`, `distinct`, and other non-linear
   operators are more tricky.  First, they have incremental and non-incremental versions.
   We typically use the former in practice, but there are situations where non-incremental
   operators are useful.  Whether we use an `_incremental` suffix for incremental operators
   of `_non_incremental` (or something to that effect) for non-incremental operators, this
   will be easy to get wrong since the type system currently doesn't distinguish between
   streams of changes and streams that carry complete collections.  An additional complication
   is that some operators have incremental implementations specialized for different nesting
   depths.  The type system does not currently enforce the use of the correct implementation
   for the given scope.
[...]

PR #94, opened 2022-07-01
Give Stream::index() a concrete return type.

ryzhyk:
One of the DBSP API usability problems listed in #92 is that most
operators are currently generic over return type, e.g.,
`Stream::index` can return any `IndexedZSet`, forcing the programmer
to annotate every call to such operators with concrete return type,
usually `OrdIndexedZSet`.

In this commit we prototype a solution to this problem for the `index`
operator.  We split it into two: `index_generic`, which can return
any indexed Z-set type, and `index`, which specializes `index_generic`
to return `OrdIndexedZSet`.  We do the same for `index_with`.

The result is that I was able to remove all type annotations on
`index` and `index_with` calls.

We also eliminate an inefficiency in the implementation of `index_with`,
which used to create an intermediate batch instead of converting input
batch directly to output Z-set.

### DIFF
--- a/src/operator/index.rs
+++ b/src/operator/index.rs
@@ -1,13 +1,12 @@
-//! Index operator.
+//! Operators to convert Z-sets into indexed Z-sets.
 
 use crate::{
-    algebra::{IndexedZSet, ZRingValue, ZSet},
     circuit::{
         operator_traits::{Operator, UnaryOperator},
         Circuit, NodeId, Scope, Stream,
     },
     circuit_cache_key,
-    trace::{cursor::Cursor, ord::OrdZSet, Batch, Builder},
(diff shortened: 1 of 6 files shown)


src/operator/index.rs
```diff
@@ -114,22 +151,92 @@ where
...
+    fn index_test() {
         let root = Root::build(move |circuit| {
             let mut inputs = vec![
                 zset!{ (1, "a") => 1
-                     , (1, "a") => 2
                      , (1, "b") => 1
-                     , (1, "b") => -1
                      , (2, "a") => 1
                      , (2, "c") => 1
+                     , (1, "a") => 2
```
mbudiu-vmw, line 238: why this change?

mbudiu-vmw (approved):

Kixiron (approved):

PR #95, opened 2022-07-01
Macro that generates a zset with all weights 1

mbudiu-vmw:


ryzhyk (approved):
Please fix the rustfmt issue before merging.

mbudiu-vmw:
Shouldn't the prepush script forbid me to push if the format is not right?

PR #101, opened 2022-07-05
Initial Nexmark generator (Person).

absoludity:

(I'm erring on the side of verbosity here to give context for what I'm doing and why, so that we can discuss async where easier).

After reading the Nexmark paper, I had a go at setting up flink (using the flink-kubernetes-operator) as well as looking through the source of the nexmark/nexmark repository, having (mistakenly?) thought that we may be able to use this specific source as an input for the DBSP benchmarking. But, as far as I can tell, the NexmarkGenerator included in that repository is just a specific implementation of a Nexmark data source for running against Flink (only). That is, although the code includes a [generic nextEvent implementation](https://github.com/nexmark/nexmark/blob/54974ef36a0d01ef8ebc0b4ba39cfc50136af0f6/nexmark-flink/src/main/java/com/github/nexmark/flink/generator/NexmarkGenerator.java#L175-L212) that generates a bid, auction or person in the correct proportions, it integrates this generator with [Flink-specific Java functions](https://github.com/nexmark/nexmark/blob/54974ef36a0d01ef8ebc0b4ba39cfc50136af0f6/nexmark-flink/src/main/java/com/github/nexmark/flink/generator/NexmarkGenerator.java#L20) and [extends Flink's RichParallelSourceFunction class](https://github.com/nexmark/nexmark/blob/master/nexmark-flink/src/main/java/com/github/nexmark/flink/source/NexmarkSourceFunction.java#L40) to provide this generated data specifically to Flink. I spent some time looking for ways that we may be able to use this as an input (including the interesting [JAVA for Rust crate](https://crates.io/crates/j4rs)), before looking for any similar Rust implementations.

Looking around I found a couple of similar nexmark benchmarking source data implemntations in Rust, but again, they're specific to the project being benchmarked - though each provides much to learn from. For example, a custom implementation of the [timely dataflow model](https://github.com/TimelyDataflow/timely-dataflow) called Megaphone is benchmarked using NexMark in the [Megaphone repo](https://github.com/strymon-system/megaphone/tree/master/nexmark). This implementation of NexMark is based on someone's (undergrad?) thesis for [running different benchmarks against timely](https://github.com/Shinmera/bsc-thesis/tree/master/benchmarks).

What is apparent in all of these NexMark implementations (flink and rust) is that the generation of the source data isn't difficult (just events, not SQL) and is just done with some configuration options so as to be flexible. The existing rust implementations found don't separate the Nexmark source generation from the (harness-specific) nexmark queries, so there is currently no Nexmark source generation as a rust crate.

So I *think* (and let me know if I'm moving in the wrong direction) that we're going to need a rust implementation of the Nexmark source generators, which doesn't look difficult (fun actually), which I've started in this branch. I'm adding the code directly into our DBSP repo initially (just because it can take a long time to get approval for a separate repo in a large org), but writing it so that the source generation can be moved and published as a separate crate (which has a few benefits).

As noted by the TODOs in the code, ~~I still need to add a configuration module before the Person generator is complete~~ (done), then I'll follow on (in separate PRs) adding the similar generators for Auction and Bid, so that I can then try passing them through a simple (q0) DBSP circuit.

It was pretty interesting seeing the simplicity of the Rust versions of the same Java functions, especially regarding the `rand::Rng` and `rand::seq::SliceRandom` traits.

You can run the unitests in the benchmark directory with:

```bash
cargo test --bench nexmark                        
    Finished test [unoptimized + debuginfo] target(s) in 0.04s
     Running benches/nexmark/main.rs (target/debug/deps/nexmark-4b029db0460e7132)

running 9 tests
test generator::person_generator::tests::test_last_base0_person_id ... ok
test generator::person_generator::tests::test_next_email ... ok
test generator::person_generator::tests::test_next_person ... ok
test generator::person_generator::tests::test_next_credit_card ... ok
test generator::person_generator::tests::test_next_base0_person_id ... ok
test generator::person_generator::tests::test_next_person_name ... ok
test generator::person_generator::tests::test_next_us_city ... ok
test generator::person_generator::tests::test_next_us_state ... ok
test generator::strings_generator::tests::next_string_length ... ok

test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

```


ryzhyk:
Given that each implementation uses its own data generator, is there a way to do an apples-to-apples comparison with, e.g., Flink?

mbudiu-vmw (approved):

mbudiu-vmw:
the dependency audit bot is unhappy

absoludity:
> the dependency audit bot is unhappy

Yeah - looks like the chrono package currently depends on a very old version of the time crate, which has the vulnerability. I'll switch it for something simpler.

benches/nexmark/generator/person_generator.rs
```diff
@@ -0,0 +1,246 @@
...
+        rng.gen_range(0..10_000),
+        rng.gen_range(0..10_000)
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use chrono::NaiveDate;
+    use rand::rngs::mock::StepRng;
```
Kixiron, line 142: More of a note than anything else, but for the real deal (not tests) we probably want to use something like [xoshiro](https://docs.rs/rand_xoshiro/) with a fixed seed for our rng source so that we get reasonably distributed data while still being deterministic

benches/nexmark/generator/person_generator.rs
```diff
@@ -0,0 +1,246 @@
...
+// Return a random credit card number.
+fn next_credit_card<R: Rng + ?Sized>(rng: &mut R) -> String {
+    format!(
+        "{:04} {:04} {:04} {:04}",
+        rng.gen_range(0..10_000),
+        rng.gen_range(0..10_000),
+        rng.gen_range(0..10_000),
+        rng.gen_range(0..10_000)
+    )
+}
```
Kixiron, line 136: This all would be much better suited as a struct that contains the rng for which these are methods on instead of free functions.
This can probably be done via a `Generator`/`DataGenerator`/`NextmarkData`/etc. struct within `generator.rs` and `impl` blocks on it within `person.rs` and `strings.rs`

benches/nexmark/generator.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright 2022 the DBSP contributors.
+// SPDX-License-Identifier: MIT
+
+//! Generators for the models usd in the Nexmark benchmark suite.
+//!
```
Kixiron, line 5:
```suggestion
//! Generators for the models usd in the Nexmark benchmark suite.
//!
```

benches/nexmark/generator/person_generator.rs
```diff
@@ -0,0 +1,246 @@
+// Copyright 2022 the DBSP contributors.
+// SPDX-License-Identifier: MIT
+
+//! Generates people for the Nexmark streaming data.
```
Kixiron, line 4:
```suggestion
//! Generates people for the Nexmark streaming data.
```

benches/nexmark/generator/person_generator.rs
```diff
@@ -0,0 +1,246 @@
...
+//!
+//! API based on the equivalent [Nexmark Flink PersonGenerator API](https://github.com/nexmark/nexmark/blob/v0.2.0/nexmark-flink/src/main/java/com/github/nexmark/flink/generator/model/PersonGenerator.java).
+
+use rand::{seq::SliceRandom, Rng};
+
+use super::strings_generator::next_string;
```
Kixiron, line 10:
```suggestion
use rand::{seq::SliceRandom, Rng};
use super::strings_generator::next_string;
```

benches/nexmark/generator/person_generator.rs
```diff
@@ -0,0 +1,246 @@
...
+        email_address: next_email(rng),
+        credit_card: next_credit_card(rng),
+        city: next_us_city(rng),
+        state: next_us_state(rng),
+        date_time: DateTime::from_timestamp(secs.try_into().unwrap(), nsecs.try_into().unwrap()),
+        extra: String::new(),
+    }
+}
+
+/// Return a random person id (base 0).
```
Kixiron, line 69: Base zero?

benches/nexmark/generator/strings_generator.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright 2022 the DBSP contributors.
+// SPDX-License-Identifier: MIT
+
+//! Generates strings which are used for different field in other model objects.
```
Kixiron, line 4:
```suggestion
//! Generates strings which are used for different field in other model objects.
```

benches/nexmark/generator.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright 2022 the DBSP contributors.
```
Kixiron, line 1: This repo uses the `mod.rs` style of modules, so could you move this into `generator/` and rename it `mod.rs`? The `_generator` is also redundant in `strings_generator.rs` and `person_generator.rs` so they can be renamed to `strings.rs` and `person.rs` (maybe `people.rs`? Seems like it rolls off the tongue better)

Kixiron (changes requested):

absoludity:
> Given that each implementation uses its own data generator, is there a way to do an apples-to-apples comparison with, e.g., Flink?

Hmm, given that the flink nexmark tests recommend running with a minimum cluster of 8 worker nodes, I'm not sure that an apples-to-apples comparison is possible - though I do still intend to see if I can run the nexmark/flink tests with one workernode. The other possibilities I still have in mind to push further are:

 - The nexmark/flink tests ultimately measure throughput per core, so that's still a baseline we can use, regrardless off the data source (we just need to ensure the queries are the same. If we can run the nexmark/flink tests with one workernode, I think we could compare sensibly (though I still want to see how the network plays a part for the flink tests, given the master node has the data source aiui).
 - I can push a bit further to see if I can run the java source generator in isolation (but even if I can, I'd still need to be converting the emitted Java models to the rust ones, so not (yet) sure it'd be useful.
 - Another option is that I saw we can run Flink [Stateful Functions](https://nightlies.apache.org/flink/flink-statefun-docs-stable/) that can be implemented in Rust - so we may be able to (very handwavy - need to dig deeper) run the original Flink/Nexmark tests against a DBSP stateful function with queries (not sure yet). Might need to convert the original flink queries to use stateful functions to do apples-to-apples.
 - I noticed the last option, using flink stateful functions, uses gRPC/Protobuf for communication, which also opens up the possibility of having a single platform-independent source generator that speaks gRPC, so we could perhaps use the same source generator with separate Flink and raw DBSP setups.

All worth investigating further, though I think the rust-based nexmark source would be useful either way assuming longer-term you don't want to be depending on flink for DBSP's own comparison against itself for progress? That is, I assume a goal here is to have a github action that generates metrics for each PR and highlight when things are getting better / worse.

benches/nexmark/generator.rs
absoludity: Ah - ok. I was just going with the newer recommendation to avoid mod.rs - happy to switch.

benches/nexmark/generator.rs
absoludity: So we shouldn't add copyright and SPDX to new files? (or let me know if you want a separate PR and I'll do the whole repo). Either way, I'll remove it from files in this PR.

benches/nexmark/generator/person_generator.rs
absoludity: I've just copied the doc-strings from the original Java implementation. In this case, it's from:

https://github.com/nexmark/nexmark/blob/54974ef36a0d01ef8ebc0b4ba39cfc50136af0f6/nexmark-flink/src/main/java/com/github/nexmark/flink/generator/model/PersonGenerator.java#L69

I'm not (yet) sure what the original author meant, but *assume* it's related to the `epoch * PERSON_PROPORTION + offset` (see `last_base0_person_id` below). I'll need to look more into the original implementation background (don't think it's mentioned in the paper).

benches/nexmark/generator/person_generator.rs
absoludity: Hmm, I was keen to stay as close as possible to the structure of the original nexmark implementation which passes the Rng to the functions, even though they're java classes, as you can see in the equiv function here:

https://github.com/nexmark/nexmark/blob/54974ef36a0d01ef8ebc0b4ba39cfc50136af0f6/nexmark-flink/src/main/java/com/github/nexmark/flink/generator/model/PersonGenerator.java#L120-L130

It seems like the intention is to avoid maintaining state and keep it purely functional where possible (for eg, you'll also see the `next_person` function also takes the current event id as an arg, which could also be state of some `NextmarkData` struct).

Would you mind if I do this initial PR for the person following the functional approach of the java implementation, then I can do a quick follow-up PR switching to a struct-basd generator so it's easier to compare and discuss there?

benches/nexmark/generator/person_generator.rs
absoludity: Great - thanks, I'll take a look (and yes, I'm only using the mock `StepRng` here for deterministic unit tests - not random at all).

ryzhyk:
A Rust generator seems like a good idea, and yes, the goal is to be able to run it in CI. I was just hoping we could design it to generate similar distributions to the Flink generator, so we could also do comparative performance analysis.

benches/nexmark/generator/person_generator.rs
Kixiron: Yep, that's totally fine for the tests since no one wants to have to figure out what values the rng will yield

benches/nexmark/generator/person_generator.rs
Kixiron: Sure, that's fine by me

benches/nexmark/generator.rs
absoludity: Done.

absoludity:
> A Rust generator seems like a good idea, and yes, the goal is to be able to run it in CI. I was just hoping we could design it to generate similar distributions to the Flink generator, so we could also do comparative performance analysis.

Yep, I'm designing it to have exactly the same configuration options so the distributions will be the same (they're deterministic, based on the set ratios of people, bids etc.), and I'll see if we can minimize other differences.

ryzhyk:
Awesome!

benches/nexmark/generator/person_generator.rs
absoludity: Done in https://github.com/absoludity/database-stream-processor/pull/1

benches/nexmark/generator/people.rs
```diff
@@ -0,0 +1,275 @@
...
+    use super::*;
+    use crate::config::Config;
+    use clap::Parser;
+    use rand::rngs::mock::StepRng;
+    use serial_test::serial;
+
+    #[test]
+    #[serial]
+    fn test_next_person() {
+        let conf = Config::parse();
```
Kixiron, line 145: I'm not sure the tests should be getting configuration from the commandline, they should just be testing against concrete inputs

benches/nexmark/generator/people.rs
```diff
@@ -0,0 +1,275 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::config::Config;
+    use clap::Parser;
+    use rand::rngs::mock::StepRng;
+    use serial_test::serial;
+
+    #[test]
+    #[serial]
```
Kixiron, line 143: Why do the tests need to be run in serial, there's no global state that I can see

Kixiron (changes requested):

benches/nexmark/generator/people.rs
```diff
@@ -0,0 +1,277 @@
...
+/// Choose a random person from any of the 'active' people, plus a few 'leads'.
+/// By limiting to 'active' we ensure the density of bids or auctions per person
+/// does not decrease over time for long running jobs.  By choosing a person id
+/// ahead of the last valid person id we will make newPerson and newAuction
+/// events appear to have been swapped in time.
+///
+/// NOTE: The above is the original comment from the Java implementation. The
+/// "base 0" is referring to the fact that the returned Id is not including the
+/// FIRST_PERSON_ID offset, and should really be "offset 0".
+
```
Kixiron, line 72:
```suggestion
```

benches/nexmark/main.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2022 the DBSP contributors.
+// SPDX-License-Identifier: MIT
+
```
Kixiron, line 3:
```suggestion
```

benches/nexmark/main.rs
```diff
@@ -0,0 +1,21 @@
...
+//! which are themselves based on the [Timely Nexmark benchmark implementation](https://github.com/Shinmera/bsc-thesis/tree/master/benchmarks).
+//!
+//! I'm writing this Nexmark data source generator with the intention of
+//! moving it to its own repo and publishing as a re-usable crate, not only
+//! so we can work with the wider community to improve it, but also because
+//! it will allow us later to extend it to support multiple data sources in
+//! parallel when DBSP can be scaled.
+#![feature(test)]
+
+extern crate test;
```
Kixiron, line 17: We shouldn't need the test feature, [`criterion`](https://github.com/bheisler/criterion.rs) is much better

benches/nexmark/model.rs
```diff
@@ -0,0 +1,54 @@
+// Copyright 2022 the DBSP contributors.
+// SPDX-License-Identifier: MIT
+
+//! Model structs for the Nexmark benchmark suite.
```
Kixiron, line 4:
```suggestion
//! Model structs for the Nexmark benchmark suite.
```

Kixiron (changes requested):

benches/nexmark/main.rs
absoludity: woops, I think I had that originally here when I had some example tests when I began. Removed now.

absoludity:
@Kixiron I've merged the other PR into here and addressed the requested changes. Let me know if there's anything else before I squash-n-merge :) (I have perms now, it seems)

PR #105, opened 2022-07-08
More API work

ryzhyk:
See commit message for details.

### DIFF
--- a/src/circuit/circuit_builder.rs
+++ b/src/circuit/circuit_builder.rs
@@ -1418,7 +1418,7 @@ where
     /// ```
     /// # use dbsp::{
     /// #     circuit::Root,
-    /// #     operator::{Z1, Apply2, Generator},
+    /// #     operator::{Z1, Generator},
     /// # };
     /// # let root = Root::build(|circuit| {
     /// // Create a data source.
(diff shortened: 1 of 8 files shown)


src/circuit/circuit_builder.rs
```diff
@@ -1427,11 +1427,7 @@ where
     /// // is a placeholder where we can later plug the input to `z1`.
     /// let (z1_output, z1_feedback) = circuit.add_feedback(Z1::new(0));
     /// // Connect outputs of `source` and `z1` to the plus operator.
-    /// let plus = circuit.add_binary_operator(
-    ///     Apply2::new(|n1: &usize, n2: &usize| n1 + n2),
-    ///     &source,
-    ///     &z1_output,
-    /// );
+    /// let plus = source.zip_with(&z1_output, |n1: &usize, n2: &usize| n1 + n2);
```
mbudiu-vmw, line 1430: this naming is problematic, because you can think of zipping several things: streams (element by element), or zsets (matching keys). In general your naming ignores streams, and refers to zsets, so I think this name can be confusing. Maybe pair_apply or pair_map (or even apply2) is better.

mbudiu-vmw (changes requested):

src/circuit/circuit_builder.rs
ryzhyk: zip for Z-sets is called join.  But I guess there might still be space for confusion. I'll rename it back to apply2.

Kixiron (approved):

PR #107, opened 2022-07-11
Non-incremental range-join operator.

ryzhyk:
This commit introduces our first form of non-equi join, where each key
in the left operand matches a contiguous range of keys in the right
operand.  Consider two indexed Z-sets `z1` and `z2`, a function
`range_func` that maps a key in `z1` to a half-closed interval of keys
`[lower, upper)` in `z2` and another function `join_func` that, given
`(k1, v1)` in `z1` and `(k2, v2)` in `z2`, returns an iterable
collection `c` of output values.  The range-join operator works as
follows:

* For each `k1` in `z1`, find all keys in `z2` within the half-closed
  interval `join_range(k1)`.

* For each `((k1, v1), w1)` in `z1` and `((k2, v2), w2)` in `z2` where `k2 ∈
  join_range(k1)`, add all values in `join_func(k1,v1,k2,v2)` to the output
  batch with weight `w1 * w2`.

So far I implemented a non-incremental version of this operator.

### DIFF
--- a/src/operator/join_range.rs
+++ b/src/operator/join_range.rs
@@ -1,0 +1,303 @@
+//! Range-join operators.
+//!
+//! Range-join is a form of non-equi join where each key in the left operand
+//! matches a contiguous range of keys in the right operand.
+//!
+//! Consider two indexed Z-sets `z1` and `z2`, a function `range_func` that
+//! maps a key in `z1` to a half-closed interval of keys `[lower, upper)` in
+//! `z2` and another function `join_func` that, given `(k1, v1)` in `z1` and
+//! `(k2, v2)` in `z2`, returns an iterable collection `c` of output values.
+//! The range-join operator works as follows:
+//! * For each `k1` in `z1`, find all keys in `z2` within the half-closed
+//!   interval `join_range(k1)`.
+//! * For each `((k1, v1), w1)` in `z1` and `((k2, v2), w2)` in `z2` where `k2 ∈
+//!   join_range(k1)`, add all values in `join_func(k1,v1,k2,v2)` to the output
+//!   batch with weight `w1 * w2`.
+
+use crate::{
+    algebra::MulByRef,
+    circuit::{
+        operator_traits::{BinaryOperator, Operator},
+        Circuit, Scope, Stream,
+    },
+    trace::{cursor::Cursor, Batch, BatchReader},
+    OrdIndexedZSet, OrdZSet,
+};
+use std::{borrow::Cow, marker::PhantomData};
+
+impl<P, I1> Stream<Circuit<P>, I1>
+where
+    P: Clone + 'static,
+{
+    /// Range-join two streams into an `OrdZSet`.
+    ///
+    /// See module documentation for the definition of the range-join operator
+    /// and its arguments.
+    ///
+    /// This operator is non-incremental, i.e., it joins the pair of batches it
+    /// receives at each timestamp ignoring previous inputs.
+    pub fn stream_join_range<RF, JF, It, I2>(
+        &self,
+        other: &Stream<Circuit<P>, I2>,
+        range_func: RF,
+        join_func: JF,
+    ) -> Stream<Circuit<P>, OrdZSet<It::Item, I1::R>>
+    where
+        I1: BatchReader<Time = ()> + Clone + 'static,
+        I1::R: MulByRef,
+        I2: BatchReader<Time = (), R = I1::R> + Clone + 'static,
+        I2::Key: Ord,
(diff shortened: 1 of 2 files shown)


src/operator/join_range.rs
```diff
@@ -0,0 +1,294 @@
...
+
+            i2_cursor.seek_key(&lower);
+            while i2_cursor.key_valid() && i2_cursor.key() < &upper {
+                i1_cursor.rewind_vals();
+                while i1_cursor.val_valid() {
+                    let w1 = i1_cursor.weight();
+                    let k1 = i1_cursor.key();
+                    let v1 = i1_cursor.val();
+                    i2_cursor.rewind_vals();
+                    while i2_cursor.val_valid() {
```
Kixiron, line 174:
```suggestion
                    i2_cursor.rewind_vals();

                    while i2_cursor.val_valid() {
```

src/operator/join_range.rs
```diff
@@ -0,0 +1,294 @@
...
+                        let w2 = i2_cursor.weight();
+                        let w = w1.mul_by_ref(&w2);
+                        for (k, v) in (self.join_func)(k1, v1, i2_cursor.key(), i2_cursor.val()) {
```
Kixiron, line 177:
```suggestion
                        let w = w1.mul_by_ref(&w2);

                        for (k, v) in (self.join_func)(k1, v1, i2_cursor.key(), i2_cursor.val()) {
```

src/operator/join_range.rs
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        O::from_tuples((), tuples)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::{circuit::Root, operator::Generator, zset};
+    use std::vec;
```
Kixiron, line 196:
```suggestion
```

src/operator/join_range.rs
```diff
@@ -0,0 +1,294 @@
...
+    I1: BatchReader<Time = (), R = O::R> + Clone + 'static,
+    I2: BatchReader<Time = (), R = O::R> + Clone + 'static,
+    I2::Key: Ord,
+    O: Batch<Time = ()> + 'static,
+    O::R: MulByRef,
+    RF: Fn(&I1::Key) -> (I2::Key, I2::Key) + 'static,
+    JF: Fn(&I1::Key, &I1::Val, &I2::Key, &I2::Val) -> It + 'static,
+    It: IntoIterator<Item = (O::Key, O::Val)> + 'static,
+{
+    fn eval(&mut self, i1: &I1, i2: &I2) -> O {
```
Kixiron, line 157: Some comments/docs on what's doing on within this would be good for future readers

Kixiron (approved):

ryzhyk:
I'm a bit tempted to call this `join_window` instead of `join_ranfe`.  The idea being that it is similar to the `window` operator, except window boundaries are defined per key in the left operand.

mbudiu-vmw:
"window" is also overloaded. The key feature of the join is that it uses comparisons, so maybe it should be called compare_join.

mbudiu-vmw:
or ordered_join.

Kixiron:
I think `.join_range` is fine

src/operator/join_range.rs
```diff
@@ -0,0 +1,303 @@
+//! Range-join operators.
+//!
+//! Range-join is a form of non-equi join where each key in the left operand
+//! matches a contiguous range of keys in the right operand.
+//!
+//! Consider two indexed Z-sets `z1` and `z2`, a function `range_func` that
+//! maps a key in `z1` to a half-closed interval of keys `[lower, upper)` in
+//! `z2` and another function `join_func` that, given `(k1, v1)` in `z1` and
+//! `(k2, v2)` in `z2`, returns an iterable collection `c` of output values.
+//! The range-join operator works as follows:
```
mbudiu-vmw, line 10: this is a strange definition. are you sure it's bilinear?
also, how is this used to model more common types of joins like JOIN on T1.x < T2.y?

ryzhyk: Why do you think it's strange? I'm pretty sure it is bilinear.

> also, how is this used to model more common types of joins like JOIN on T1.x < T2.y?

It's not meant to be the most general equijoin. It can probably be generalized or we may add another operator for that.

mbudiu-vmw: It is strange because the second function returns a collection. The size of the colection is not related to the size of the inputs

ryzhyk: Usually the collection is `Option<>`, but it can also be, e.g., `Vec<>`. I don't think this breaks bi-linearity, since each element in the collection will be added with the weight that is the product of input weights.

PR #108, opened 2022-07-12
[nexmark] Add initial nexmark auction generation function.

absoludity:

This continues the work that began in #101, this time adding the first function required for the Nexmark auction generation (which in turn requires new functions on the config object(s)), again, based closely on the Java implementation.

As part of this change, I also:
- separated out the general nexmark config from the individual generator config (individual in the sense that the original code is structured such that you can run multiple generators with different configs simultaneously), matching the java implementation (I didn't do this in the last PR as it wasn't yet obvious *why* it was separated, but it is now).
- removed the `DateTime` alias of the `std::time::SystemTime` (I'd done this at first as I wasn't sure which time implementation I'd use and wanted to be able to switch it later, but it seems the std system time is sufficient)

Note: I'm adding unit tests for the functions as I create them (which aren't in the original java implementation) mainly to help me ensure my understanding of the functions is correct (as it's not always obvious from the code, at least to me), though I think they're helpful as concise documentation of what the functions do with example values etc.

Next up: the other auction-related generation functionality before finally moving to the bid model generation.

### DIFF
--- a/benches/nexmark/config.rs
+++ b/benches/nexmark/config.rs
@@ -4,15 +4,8 @@
 //! and the specific [Nexmark Flink Generator config](https://github.com/nexmark/nexmark/blob/v0.2.0/nexmark-flink/src/main/java/com/github/nexmark/flink/generator/GeneratorConfig.java).
 use clap::Parser;
 
-// We start the ids at specific values to help ensure the queries find a match
-// even on small synthesized dataset sizes.
-pub const FIRST_PERSON_ID: usize = 1000;
-
 // Number of yet-to-be-created people and auction ids allowed.
 pub const PERSON_ID_LEAD: usize = 10;
-
-/// Maximum number of people to consider as active for placing auctions or bids.
-pub const NUM_ACTIVE_PEOPLE: usize = 1000;
 
 /// A Nexmark streaming data source generator
 ///
@@ -20,14 +13,6 @@ pub const NUM_ACTIVE_PEOPLE: usize = 1000;
 #[derive(Parser, Debug)]
 #[clap(author, version, about)]
 pub struct Config {
-    #[clap(
-        long = "person-proportion",
-        default_value = "1",
-        env = "NEXMARK_PERSON_PROPORTION",
-        help = "Specify the proportion of events that will be new people"
-    )]
-    pub person_proportion: usize,
-
     #[clap(
         long = "auction-proportion",
         default_value = "3",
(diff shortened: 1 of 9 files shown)


benches/nexmark/config.rs
```diff
@@ -37,16 +22,89 @@ pub struct Config {
     pub auction_proportion: usize,
 
     #[clap(
-        long = "bit-proportion",
+        long = "bid-proportion",
```
Kixiron, line 25:
```suggestion
        long,
```
Clap will rename it automatically, so this can be applied to other members too (e.g. `auction_proportion`)

benches/nexmark/config.rs
```diff
@@ -37,16 +22,89 @@ pub struct Config {
...
+    #[clap(
+        long = "out-of-order-group-size",
+        default_value = "1",
+        env = "NEXMARK_OUT_OF_ORDER_GROUP_SIZE",
+        help = "Number of events in out-of-order groups. 1 implies no out-of-order events. 1000 implies every 1000 events per generator are emitted in pseudo-random order."
+    )]
+    pub out_of_order_group_size: usize,
+
+    #[clap(
+        long = "num-in-flight-auctions",
```
Kixiron, line 49:
```suggestion
        long,
```

benches/nexmark/config.rs
```diff
@@ -37,16 +22,89 @@ pub struct Config {
...
         env = "NEXMARK_BID_PROPORTION",
         help = "Specify the proportion of events that will be new bids"
     )]
     pub bid_proportion: usize,
+
+    #[clap(
+        long = "person-proportion",
+        default_value = "1",
+        env = "NEXMARK_PERSON_PROPORTION",
+        help = "Specify the proportion of events that will be new people"
```
Kixiron, line 36: Help can be given via doc comments, I think that's nicer and it also provides docs with editors that support it
```suggestion
    /// Specify the proportion of events that will be new people
    #[clap(
        long = "person-proportion",
        default_value = "1",
        env = "NEXMARK_PERSON_PROPORTION",
```

benches/nexmark/config.rs
```diff
@@ -37,16 +22,89 @@ pub struct Config {
...
+            auction_proportion: 3,
+            bid_proportion: 46,
+            person_proportion: 1,
+            num_in_flight_auctions: 100,
+            num_active_people: 1000,
+            out_of_order_group_size: 1,
+            first_event_rate: 10_000,
+            num_event_generators: 1,
+        }
+    }
```
Kixiron, line 104: Maybe just implement `Default` for `Config` and then use `Config::default()` within tests?

benches/nexmark/config.rs
```diff
@@ -37,16 +22,89 @@ pub struct Config {
...
+
+    #[test]
+    fn test_total_proportion_default() {
+        assert_eq!(make_default_nexmark_config().total_proportion(), 50);
+    }
```
Kixiron, line 109: We don't really need to test default literals imo

benches/nexmark/generator/auctions.rs
```diff
@@ -0,0 +1,60 @@
+//! Generates people for the Nexmark streaming data source.
+//!
+//! API based on the equivalent [Nexmark Flink PersonGenerator API](https://github.com/nexmark/nexmark/blob/v0.2.0/nexmark-flink/src/main/java/com/github/nexmark/flink/generator/model/AuctionGenerator.java).
+
+use super::NexmarkGenerator;
+use anyhow::Result;
+use rand::Rng;
+use std::cmp;
+use std::time::{Duration, SystemTime};
```
Kixiron, line 9:
```suggestion
use std::{cmp, time::{Duration, SystemTime}};
```

Kixiron (changes requested):

ryzhyk:
@absoludity , do you have the pre-push hook setup as described in the README.  I notice that the `rustfmt` test is failing, which may be because the hook did not run or because the version of `rustfmt` in CI is different (which happens with nightly toolchains).

benches/nexmark/config.rs
absoludity: Hmm, OK, so I've done this and while it is much neater (thanks!) it now means that I'm defining the defaults in two different places (perhaps not a huge issue). The only way around that that I see is to call `::parse()` instead which will return the default object as long as no env vars are set (so really isn't the default). Let me know what you think.

absoludity:
> @absoludity , do you have the pre-push hook setup as described in the README? I notice that the `rustfmt` test is failing, which may be because the hook did not run or because the version of `rustfmt` in CI is different (which happens with nightly toolchains).

Nope - I didn't, but do now. I was running `rustfmt` on save, but didn't have the extra `--config format_code_in_doc_comments=true,wrap_comments=true`.  Thanks.

ryzhyk:
@absoludity , @Kixiron , please merge this when ready.

Kixiron (approved):

PR #109, opened 2022-07-13
[nexmark] Add last_base0_auction_id and next_base0_auction_id.

absoludity:

Follows #108 , adding two more of the auction generation helper functions. Pushing separately as it took me quite a while to understand what they are actually doing, so assume it'll be similar for reviewing, though hopefully the unit test-cases that I've added make it clearer.

Next up: the actual next_auction method.

### DIFF
--- a/benches/nexmark/generator/auctions.rs
+++ b/benches/nexmark/generator/auctions.rs
@@ -11,6 +11,55 @@ use std::{
 };
 
 impl<R: Rng> NexmarkGenerator<R> {
+    /// Return the last valid auction id (ignoring FIRST_AUCTION_ID). Will be
+    /// the current auction id if due to generate an auction.
+    fn last_base0_auction_id(&self, event_id: u64) -> u64 {
+        let mut epoch = event_id / self.config.nexmark_config.total_proportion() as u64;
+        let mut offset = event_id % self.config.nexmark_config.total_proportion() as u64;
+
+        if offset < self.config.nexmark_config.person_proportion as u64 {
+            // About to generate a person.
+            // Go back to the last auction in the last epoch.
+            epoch = match epoch.checked_sub(1) {
+                Some(e) => e,
+                None => return 0,
+            };
+            offset = self.config.nexmark_config.auction_proportion as u64 - 1;
+        } else if offset
+            >= (self.config.nexmark_config.person_proportion
+                + self.config.nexmark_config.auction_proportion) as u64
+        {
+            // About to generate a bid.
+            // Go back to the last auction generated in this epoch.
+            offset = self.config.nexmark_config.auction_proportion as u64 - 1;
+        } else {
+            // About to generate an auction.
+            offset -= self.config.nexmark_config.person_proportion as u64;
+        }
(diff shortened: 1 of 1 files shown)


absoludity:
The codecov decrease of 1.5% is, I assume, because I'm using the `StepRng` which for the Rng during the unit tests. Unfortunately, this is useful for calling `gen()` (ie. generating a random number - it just steps through a predictable sequence), but for `gen_range` which I need throughout, it always just returns the minimum of the range (usually zero), so certain paths don't get executed. If it's worthwhile, happy to try creating a `StepRangeRng` that we could use to keep the code coverage.

benches/nexmark/generator/auctions.rs
```diff
@@ -11,6 +11,55 @@ use std::{
...
+                + self.config.nexmark_config.auction_proportion) as u64
+        {
+            // About to generate a bid.
+            // Go back to the last auction generated in this epoch.
+            offset = self.config.nexmark_config.auction_proportion as u64 - 1;
+        } else {
+            // About to generate an auction.
+            offset -= self.config.nexmark_config.person_proportion as u64;
+        }
+        epoch * self.config.nexmark_config.auction_proportion as u64 + offset
```
Kixiron, line 39: Maybe do these casts once and store them in variables

benches/nexmark/generator/auctions.rs
```diff
@@ -11,6 +11,55 @@ use std::{
...
+        let min_auction = cmp::max(
+            match self
+                .last_base0_auction_id(next_event_id)
+                .checked_sub(self.config.nexmark_config.num_in_flight_auctions as u64)
+            {
+                Some(e) => e,
+                None => 0,
+            },
+            0,
+        );
```
Kixiron, line 57: I'm reasonably sure this is just `.saturating_sub()`, although I'm not really sure what the `max()` is doing here
```suggestion
        let min_auction = self
                .last_base0_auction_id(next_event_id)
                .saturating_sub(self.config.nexmark_config.num_in_flight_auctions as u64);
```

Kixiron (approved):

benches/nexmark/generator/auctions.rs
absoludity: Yeah, I was wondering whether I should just use u64 for them. It seemed odd to choose u64 for something I know is small in the configuration, but given that they're consistently used in expressions for the (long) ids, perhaps it'd be better to just switch them to u64 in the config?

benches/nexmark/generator/auctions.rs
absoludity: Oh - yes, it's amazing what Rust has built in to the std types compared to Java :) I think you're right, and I assume the `max()` with zero is only there because in the Java implementation it's a long (ie. signed) type. Thanks.

benches/nexmark/generator/auctions.rs
absoludity: Actually, it makes it much more readable to create variables for each (rather than reaching through `self.config.nexmark_config.` each time) anyway, so I'll just do as you suggested :)

PR #112, opened 2022-07-14
[nexmark] Add supporting functions for the bids generator.

absoludity:

Following #110, this PR adds the supporting functions for generating a bid.

The main difference in this PR is that, following the Java implementation, I've used an in-memory cache for the function that generates a new channel URL - `get_new_channel_instance` - and, similar to the Java implementation, I'm not able to cache a function with a generic trait (`Rng`). So the function itself creates a `thread_local` `Rng` which means that the test has to ensure consistent and accurate test results while using a real Rng.

See what you think. 

Next PR will add the `next_bid` implementation to the `NexmarkGenerator` struct, just like the other models.

### DIFF
--- a/benches/nexmark/generator/bids.rs
+++ b/benches/nexmark/generator/bids.rs
@@ -2,33 +2,35 @@
 //!
 //! API based on the equivalent [Nexmark Flink PersonGenerator API](https://github.com/nexmark/nexmark/blob/v0.2.0/nexmark-flink/src/main/java/com/github/nexmark/flink/generator/model/BidGenerator.java).
 use super::strings::next_string;
-use cached::{proc_macro::cached, SizedCache};
-use rand::{thread_rng, Rng};
+use super::NexmarkGenerator;
+use cached::Cached;
+use rand::Rng;
 
-const CHANNELS_NUMBER: usize = 10_000;
+pub const CHANNELS_NUMBER: usize = 10_000;
 
 const BASE_URL_PATH_LENGTH: usize = 5;
 
-// Similar to the Java implementation, the cached version of
-// `get_new_channel_instance` needs to create its own Rng, as it's not currently
-// possible to pass a generic Rng in while using the cached macro.
-#[cached(
-    type = "SizedCache<usize, (String, String)>",
-    create = "{ SizedCache::with_size(CHANNELS_NUMBER) }"
-)]
-fn get_new_channel_instance(channel_number: usize) -> (String, String) {
-    let mut rng = thread_rng();
-    let mut url = get_base_url(&mut rng);
-    // Just following the Java implementation: 1 in 10 chance that
-    // the URL is returned as is, otherwise a channel_id query param is added to the
-    // URL. Also following the Java implementation which uses `Integer.reverse` to
-    // get a deterministic channel id.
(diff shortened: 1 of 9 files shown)


benches/nexmark/generator/bids.rs
```diff
@@ -0,0 +1,99 @@
...
+const BASE_URL_PATH_LENGTH: usize = 5;
+
+// Similar to the Java implementation, the cached version of
+// `get_new_channel_instance` needs to create its own Rng, as it's not currently
+// possible to pass a generic Rng in while using the cached macro.
+#[cached(
+    type = "SizedCache<usize, (String, String)>",
+    create = "{ SizedCache::with_size(CHANNELS_NUMBER) }"
+)]
+fn get_new_channel_instance(channel_number: usize) -> (String, String) {
```
Kixiron, line 19: Can't we just do this caching manually by storing a cache within the `NextmarkGenerator` struct, that way we cold provide an rng for it?

benches/nexmark/generator/strings.rs
```diff
@@ -7,14 +7,24 @@ use rand::{distributions::Alphanumeric, Rng};
 
 const MIN_STRING_LENGTH: usize = 3;
 
+/// Return a random string of up to `max_length`.
+pub(super) fn next_string<R: Rng>(rng: &mut R, max_length: usize) -> String {
+    let len = rng.gen_range(MIN_STRING_LENGTH..=max_length);
+    rng.sample_iter(&Alphanumeric)
+        .take(len)
+        .map(char::from)
+        .collect()
```
Kixiron, line 16: [`DistString::sample_string()`](https://docs.rs/rand/latest/rand/distributions/trait.DistString.html#method.sample_string) is a little nicer here

benches/nexmark/generator/bids.rs
absoludity: By manually, do you mean without the `cached` crate all together? In which case, yes, though I think it's nice to avoid the extra code by just using a fn here? Or maybe you mean still using the `cached` library but manually instantiating a `SizedCache`... - hmm, that could be nicer - let me try.

absoludity: I've done the latter and it works nicely - thanks. One thing I did notice while doing this is that, since the use of the `rng` requires it to be `&mut`, every method that uses the `rng` field, needs `&mut self`, which in turn means the compiler has to assume (a ref to) the whole struct is borrowed mutably, rather than just the `rng` field. It's for this reason that I've not also moved the `get_base_url` function into the structs implementation, as currently I can just pass the one field as a mutable borrow. Let me know if I'm missing something here and I'll update it in another branch.

absoludity: Huh, though with my latest change it seems the compiler is failing to compile the `time` crate with
```
thread 'rustc' panicked at 'index out of bounds: the len is 1 but the index is 2', compiler\rustc_builtin_macros\src\format.rs:1001:28
```

Kixiron (approved):

benches/nexmark/generator/bids.rs
absoludity: OK, it seems whatever issue that was causing `rustc` to panic in the nightly builds was now fixed :)

PR #116, opened 2022-07-18
Add the next_event function to the generator.

absoludity:

Follows #115, this PR adds the functions for generating the sequence nexmark data events.

As per the TODO in the code, the main pain in this PR was trying to figure out whether the Java implementation is correct or not (as my unit tests of my rust port were saying it wasn't). It's probably not a big deal in the end, but the Java implementation advertises that the generated events are consistent (in that the seller of a Bid refers to a valid person id that has been generated), but it seems the initial person 1000 to which the first bids and auctions may refer to is never generated.

I plan to triple check tomorrow and create an issue on the java repo.

Anyway, next PR will finally be looking at a DBSP source (I hope).


benches/nexmark/generator/config.rs
```diff
@@ -27,26 +34,52 @@ pub struct Config {
...
+                    / (nexmark_config.total_proportion() as u64
+                        * [
+                            nexmark_config.avg_person_byte_size as u64,
+                            nexmark_config.avg_auction_byte_size as u64,
+                            nexmark_config.avg_bid_byte_size as u64,
+                        ]
+                        .iter()
+                        .max()
+                        .unwrap())
+            }
```
Kixiron, line 62:
```suggestion
            0 => {
                let max_average = [
                    nexmark_config.avg_person_byte_size as u64,
                    nexmark_config.avg_auction_byte_size as u64,
                    nexmark_config.avg_bid_byte_size as u64,
                ]
                .iter()
                .max()
                .unwrap();
                u64::MAX / (nexmark_config.total_proportion() as u64 * max_average)
            }
```

Kixiron (approved):

benches/nexmark/generator/config.rs
absoludity: Updated, but ended up being slightly different, needing to deref `max_average` (as `.max()` returns an `Option<&usize>` here), and without the deref, the borrow-checker complains: 

```
error[E0716]: temporary value dropped while borrowed
  --> benches/nexmark/generator/config.rs:52:35
   |
52 |                   let max_average = [
   |  ___________________________________^
53 | |                     nexmark_config.avg_person_byte_size as u64,
54 | |                     nexmark_config.avg_auction_byte_size as u64,
55 | |                     nexmark_config.avg_bid_byte_size as u64,
56 | |                 ]
   | |_________________^ creates a temporary which is freed while still in use
...
59 |                   .unwrap();
   |                            - temporary value is freed at the end of this statement
60 |                   u64::MAX / (nexmark_config.total_proportion() as u64 * max_average)
   |                                                                          ----------- borrow later used here
   |
   = note: consider using a `let` binding to create a longer lived value

```

though it's not clear to me why temporary value hasn't already been used to calculate the max_average referenced value when used. Anyway, also then moved the cast to u64 so it's just done once for `max_average`.

PR #117, opened 2022-07-19
Add initial DBSP Nexmark source.

absoludity:

Just a draft, following on from #116 , this is basically me learning a bit about ZSets, how to create them for expected data in tests, and creating an example DBSP source. Also refactored the generator test a little so I can generate test data for the source to use (while staying DRY).

I don't yet have enough understanding of a lot of details (eg, whether using `NextEvent` for the source is ok - perhaps should be `Event`, or whether the source should be batchable (as I've done) or should emit one event for each time increment).

Anyway, I've left a few questions and things that aren't clear to me inline, if anyone has time.

Thanks!

### DIFF
--- a/benches/nexmark/nexmark_dbsp_source.rs
+++ b/benches/nexmark/nexmark_dbsp_source.rs
@@ -1,0 +1,120 @@
+//! DBSP Source operator that reads from a Nexmark Generator.
+
+use crate::generator::{NexmarkGenerator, NextEvent};
+use dbsp::{
+    algebra::{ZRingValue, ZSet},
+    circuit::{
+        operator_traits::{Data, Operator, SourceOperator},
+        Scope,
+    },
+};
+use rand::Rng;
+use std::{borrow::Cow, marker::PhantomData};
+
+pub struct NexmarkDBSPSource<R: Rng, W, C> {
+    /// The generator for events emitted by this source.
+    // TODO(absoludity): Longer-term, it'd be great to use a client (such as a gRPC client) here to
+    // completely separate the generator and allow the generator to be used with other languages
+    // etc.
+    generator: NexmarkGenerator<R>,
+    batch_size: usize,
+    _t: PhantomData<(C, W)>,
+}
+
+impl<R, W, C> NexmarkDBSPSource<R, W, C>
+where
+    R: Rng,
+{
+    pub fn from_generator(generator: NexmarkGenerator<R>, batch_size: usize) -> Self {
+        NexmarkDBSPSource {
+            generator,
+            batch_size,
+            _t: PhantomData,
+        }
+    }
+}
+
+impl<R, W, C> Operator for NexmarkDBSPSource<R, W, C>
+where
+    C: Data,
+    R: Rng + 'static,
+    W: 'static,
+{
+    fn name(&self) -> Cow<'static, str> {
+        Cow::from("NexmarkDBSPSource")
+    }
+
+    // Do clock_start and clock_end make sense for an infinite source? Or should
+    // the Nexmark generator use the max_events and stop generating events after
+    // that.
+    /// That would also mean then that the fixed point for this data source
+    /// would be at max_events?  Currently the generator will keep spitting out
+    /// events, but maybe `next_event` could be updated to return a
+    /// `Result<Option<NextEvent>>` and return Ok(None) after max events?
+    fn fixedpoint(&self, _scope: Scope) -> bool {
+        false
+    }
+}
+
+impl<R, W, C> SourceOperator<C> for NexmarkDBSPSource<R, W, C>
+where
+    R: Rng + 'static,
+    W: ZRingValue + 'static,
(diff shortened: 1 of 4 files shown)


ryzhyk:
Thanks! I'll read it tomorrow.

benches/nexmark/nexmark_dbsp_source.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    // Do clock_start and clock_end make sense for an infinite source? Or should
+    // the Nexmark generator use the max_events and stop generating events after
+    // that.
+    /// That would also mean then that the fixed point for this data source
+    /// would be at max_events?  Currently the generator will keep spitting out
+    /// events, but maybe `next_event` could be updated to return a
+    /// `Result<Option<NextEvent>>` and return Ok(None) after max events?
+    fn fixedpoint(&self, _scope: Scope) -> bool {
```
ryzhyk, line 54: If the natural behavior for this operator is to generate an infinite sequence of events, then returning `false` is the right thing to do.  If the generator has a predefined limit on the number of outputs, then `eval` should start producing empty sets once the limit has been reached, and `fixedpoint` should return `true` after outputting the first empty set to indicate that all future outputs will be exactly the same.

mbudiu-vmw:
At this point there is no nice consumption API for DBSP. Currently we do all the testing by essentially embedding Generator nodes in the circuit. But we should provide a higher level API, probably based on the DDlog runtime API, with beginTransaction, endTransaction, and a series of inserts/deletes for raw elements, or just inserts for zsets. There is also the question of getting back results, in DDlog this was originally a callback API, but we moved to a synchronous API where endTransaction just returns all the output changes. However, a callback API can be statically typed (since you have a callback for each output), while returning output changes will require dynamic typing. Right now there is no dynamically typed interface for DBSP circuits.

benches/nexmark/nexmark_dbsp_source.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+impl<R, W, C> SourceOperator<C> for NexmarkDBSPSource<R, W, C>
+where
+    R: Rng + 'static,
+    W: ZRingValue + 'static,
+    C: Data + ZSet<Key = NextEvent, R = W>,
+{
+    fn eval(&mut self) -> C {
+        // Q: Why is the time argument expecting ()? Same for the key tuple's second el?
```
ryzhyk, line 66: // Q: Why is the time argument expecting ()? Same for the key tuple's second el?

Notionally, all DBSP batches are collections of 4-tuple: `(key, value, time, weight)`. The `Batch::from_tuples(time, tuples)` method creates a batch where all tuples have the same `time` value, so the `tuples` vector only needs to specify `key`, `value`, and `weight`.  For Z-sets, the `value` type is `()`.

As for the time type being `()`, it's a form of optimization that may or may not survive future design iterations :) Technically, each circuit has a clock that increments sequentially from zero to infinity.  We could always model such a clock as `usize`.  A nested circuit has its own local clock, so its logical time can be modeled as a `(usize, usize)` tuple.  Every tuple in a batch is labeled by the logical time when it war produced: tuples produced during the first clock cycle have timestamp 0, tuples produced during the second cycle have timestamp 1, etc.

It turns out though that we don't need to track the time of each tuple precisely.  We need to store just enough timing information to distinguish between tuples currently in flight inside the circuit.  In our current "synchronous" model where we process each input to completion before ingesting more data, we don't need to track the timing at all in the root circuit, hence the `()` time.

There are some serious questions about the ability of this architecture to scale out, so the design may change in the future.

ryzhyk (approved):
The source should definitely be batchable. Processing one record per clock cycle is inefficient.  How does the Flink implementation batch inputs?

Kixiron (approved):

absoludity (commented):
> The source should definitely be batchable. Processing one record per clock cycle is inefficient.

OK, I'll continue using a batchable source here, but I assume I'll need to ensure this behaves correctly with Nexmark's assumption of wallclock time. More below.

> How does the Flink implementation batch inputs?

AFAICT, it doesn't batch them in the flink source and in fact, it potentially sleeps until the next (single) event should be emitted (according to its generated wallclock timestamp - I assume mimicking latency/delays). You can see the thread being slept here in the source until it should be emitted:

https://github.com/nexmark/nexmark/blob/master/nexmark-flink/src/main/java/com/github/nexmark/flink/source/NexmarkSourceFunction.java#L114-L135

Given @mbudiu-vmw 's comment about the consumption API above, it's not clear whether it's OK for us to do similarly here given that our source is embedded in the circuit. I mean, I can mimic the same sleep behavior during the source generator's `eval` before emitting an event (or a batch of events), and that will simulate the source getting events with whatever delays are configured. I'll assume that's OK and move forward with this in the next PR, unless I hear otherwise.

Given that this PR has 2 approvals as is, I'll land it so you only have to look at the changes in the next PR. But feel free to add more to any discussion here if I've made incorrect assumptions.

Thanks again for the input all.

PR #118, opened 2022-07-20
[nexmark] Nexmark source returns empty zsets when exhausted.

absoludity:

This PR follows on from #117 and addresses and tests three of the four items mentioned in the discussion:

1. The source emits `model::Event`'s rather than the generator-specific `NextEvent` which just has extra metadata used by the source (only).
2. Updates the `clock_start` to reset the generator so that it begins the sequence again when called.
3. Updates `fixedpoint` to return true when the generator is exhausted. As part of this, I also updated the generator so that it's `next_event` returns a `Result<Option<NextEvent>>` rather than just `Result<NextEvent>` so that the source can then filter out `None` results and therefore automatically returns an empty ZSet once the generator is exhausted.

The fourth item will be to update the batching as discussed with Leonid, which I'll do and discuss in the next PR.

### DIFF
--- a/benches/nexmark/nexmark_dbsp_source.rs
+++ b/benches/nexmark/nexmark_dbsp_source.rs
@@ -1,6 +1,7 @@
 //! DBSP Source operator that reads from a Nexmark Generator.
 
-use crate::generator::{NexmarkGenerator, NextEvent};
+use crate::generator::NexmarkGenerator;
+use crate::model::Event;
 use dbsp::{
     algebra::{ZRingValue, ZSet},
     circuit::{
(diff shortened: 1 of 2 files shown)


benches/nexmark/nexmark_dbsp_source.rs
```diff
@@ -44,30 +45,33 @@ where
...
 {
     fn eval(&mut self) -> C {
-        // Q: Why is the time argument expecting ()? Same for the key tuple's second el?
         C::from_tuples(
             (),
             (0..self.batch_size)
-                .map(|_| ((self.generator.next_event().unwrap(), ()), W::one()))
+                .map(|_| self.generator.next_event().unwrap())
+                .filter(|event| event.is_some())
+                .map(|event| ((event.unwrap().event, ()), W::one()))
```
Kixiron, line 74:
```suggestion
                .filter_map(|event| event.map(|event| ((event.event, ()), W::one())))
```

Kixiron (approved):

PR #119, opened 2022-07-20
[nexmark] Update source eval to batch all events ready.

absoludity:

After chatting with @ryzhyk  about the batching in the nexmark source, where he says:

> OTOH, this gives as a natural apples-to-apples comparison: DBSP will grab all the data buffered at its input by the generator, up to some fixed bound, and process it as one batch.
> This way we'll be doing natural batching simulating how the system would run in the real world, rather than introducing an artificial batch size, which may affect the performance.

So pushing this PR to discuss further. It updates the NexmarkDBSPSource's `eval` method so that it:
- Waits until it has at least one event to emit, and
- returns a batch with as many generated events as are ready (ie. in the past according to the wallclock),

~~Still need to (work out how to) test it properly (given the use of `now()`). I'll probably inject the function into the struct during a `::new()` constructor, while tests can manually set a canned version. Or even just show that when the wallclock is set automatically, only a few results are returned in each batch. But I'll wait until I chat with Leonid before going further on this PR.~~

### DIFF
--- a/benches/nexmark/generator/mod.rs
+++ b/benches/nexmark/generator/mod.rs
@@ -35,6 +35,14 @@ pub struct NexmarkGenerator<R: Rng> {
     wallclock_base_time: Option<u64>,
 }
 
+pub fn wallclock_time() -> Result<u64> {
+    let millis: u64 = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)?
+        .as_millis()
+        .try_into()?;
+    Ok(millis)
+}
+
 impl<R: Rng> NexmarkGenerator<R> {
     pub fn new(config: Config, rng: R) -> NexmarkGenerator<R> {
         NexmarkGenerator {
@@ -82,12 +90,7 @@ impl<R: Rng> NexmarkGenerator<R> {
             return Ok(None);
         }
         if self.wallclock_base_time == None {
-            self.wallclock_base_time = Some(
-                SystemTime::now()
-                    .duration_since(SystemTime::UNIX_EPOCH)?
-                    .as_millis()
-                    .try_into()?,
-            )
+            self.wallclock_base_time = Some(wallclock_time()?);
         }
 
         // When, in event time, we should generate the event. Monotonic.
--- a/benches/nexmark/generator/mod.rs
+++ b/benches/nexmark/generator/mod.rs
@@ -35,12 +35,11 @@ pub struct NexmarkGenerator<R: Rng> {
     wallclock_base_time: Option<u64>,
 }
 
-pub fn wallclock_time() -> Result<u64> {
-    let millis: u64 = SystemTime::now()
-        .duration_since(SystemTime::UNIX_EPOCH)?
-        .as_millis()
-        .try_into()?;
-    Ok(millis)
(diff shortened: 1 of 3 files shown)


benches/nexmark/generator/mod.rs
```diff
@@ -35,6 +35,14 @@ pub struct NexmarkGenerator<R: Rng> {
     wallclock_base_time: Option<u64>,
 }
 
+pub fn wallclock_time() -> Result<u64> {
+    let millis: u64 = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)?
```
Kixiron, line 40: You can unwrap this first error, `SystemTime::now()` will never be before `UNIX_EPOCH`

benches/nexmark/generator/mod.rs
```diff
@@ -35,6 +35,14 @@ pub struct NexmarkGenerator<R: Rng> {
...
+        .as_millis()
+        .try_into()?;
```
Kixiron, line 42: You can probably just use an `as u64` cast here or maybe even just call `.as_secs()` (don't know what behavior you're emulating though)

benches/nexmark/nexmark_dbsp_source.rs
```diff
@@ -66,12 +71,49 @@ where
     C: Data + ZSet<Key = Event, R = W>,
 {
     fn eval(&mut self) -> C {
+        // Grab a next event, either the last event from the previous call that
+        // was saved because it couldn't yet be emitted, or the next generated
+        // event.
+        let next_event = match self.next_event.clone() {
+            Some(e) => Some(e),
+            None => self.generator.next_event().unwrap(),
+        };
```
Kixiron, line 80:
```suggestion
        let next_event = self.next_event
            .clone()
            .unwrap_or_else(||  self.generator.next_event().unwrap());
```

Kixiron (approved):

benches/nexmark/generator/mod.rs
absoludity: Ah - ok. Updated both and the function to no longer return a `Result`. Thanks.

benches/nexmark/nexmark_dbsp_source.rs
absoludity: I'd actually need something like `is_some_or_else()` here, rather than `unwrap_or_else` to simplify as you've suggested here. It's a bit confusing as `self.next_event` as an `Option<NextEvent>`, whereas `self.generator.next_event()` returns a `Result<Option<NextEvent>>` (as it can error).

So I'm trying to end up with an `Option<NextEvent>` whereas `.unwrap_or_else` implies the type is just `NextEvent` while the or else part correctly returns an `Option<NextEvent>`.

Let me know if there's still a way to do something more concise here that I've missed.

Kixiron: That'd be `.or_else(|| self.generator.next_event().unwrap())` then

absoludity: Nice, that works - added to the next PR. Thanks.

PR #120, opened 2022-07-21
Shard batches across multiple workers.

ryzhyk:
Theory
======

We parallelize processing across `N` worker threads by creating a
replica of the same circuit per thread and sharding data across
replicas.  To ensure correctness (i.e., that the sum of outputs
produced by individual workers is equal to the output produced
by processing the entire dataset by one worker), sharding must satisfy
certain requirements determined by each operator.  In particular,
for `distinct`, and `aggregate` all tuples that share the same key
must be processed by the same worker.  For `join`, tuples from both
input streams with the same key must be processed by the same worker.

Other operators, e.g., `filter` and `flat_map`, impose no restrictions
on the sharding scheme: as long as each tuple in a batch is
processed by some worker, the correct result will be produced.  This
is true for all linear operators.

The `shard` operator
====================

This commit introduces the `shard` operator, which shards input batches
based on the hash of the key, making sure that tuples with the same key
always end up at the same worker.  More precisely, the operator
**re-shards** its input by partitioning batches in the input stream of
each worker based on the hash of the key, distributing resulting
fragments amond peers and re-assembling fragments at each peer.

I used the new operator to generalize `join`, `distinct` and `aggregate`
to work in the multithreaded environment.

Fixed point
==========

In a multithreaded environment the fixedpoint check cannot be performed
locally.  The circuit must iterate until all peers have reached a fixed
point.  I used the `Exchange` synchronization primitive to broadcast
the fixed point state from each worker to all peers.

TODOs
=====

We need to figure out how to parallelize non-equi joins.  At the moment
`range_join` will not work correctly with more than one worker thread.

Performance
===========

Very preliminary performance results:

- path benchmark: 10x speedup going from 0 to 16 threads
- galen benchmark: 3x speedup going from 0 to 8 threads, with little
  improvement beyond 4 threads.

This suggests that more complex programs with many operators that require
re-sharding input streams are harder to scale as every such operator introduces
a global barrier.

Interestingly, the DD version of galen only scales marginally better than DBSP,
with 4x speedup on 8 worker threads, but its absolute performance is still
slightly worse than DBSP.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ impl-trait-for-tuples = "0.2"
 deepsize = "0.2.0"
 deepsize_derive = "0.1.2"
 textwrap = "0.15.0"
+fxhash = "0.2"
 
 # TODO: eliminate dependency on timely-dataflow by cloning relevant
 # parts.
--- a/benches/galen.rs
+++ b/benches/galen.rs
@@ -72,7 +72,7 @@ where
 fn main() -> Result<()> {
     unpack_galen_data()?;
 
-    let hruntime = Runtime::run(1, |_runtime, _index| {
+    let hruntime = Runtime::run(8, || {
         let monitor = TraceMonitor::new_panic_on_error();
 
         let root = Root::build(|circuit| {
@@ -174,7 +174,9 @@ fn main() -> Result<()> {
                         let ir1 = child.region("IR1", || {
                             p_by_2.join::<NestedTimestamp32, _, _, _>(&p_by_1, |&_y, &x, &z| (x, z))
                         });
-                        ir1.inspect(|zs: &OrdZSet<_, _>| println!("ir1: {}", zs.len()));
+                        ir1.inspect(|zs: &OrdZSet<_, _>| {
+                            println!("{}: ir1: {}", Runtime::worker_index(), zs.len())
+                        });
 
                         // IR2: q(x,r,z) := p(x,y), q(y,r,z)
                         let ir2 = child.region("IR2", || {
@@ -184,7 +186,9 @@ fn main() -> Result<()> {
                                 })
                         });
 
-                        ir2.inspect(|zs: &OrdZSet<_, _>| println!("ir2: {}", zs.len()));
+                        ir2.inspect(|zs: &OrdZSet<_, _>| {
+                            println!("{}: ir2: {}", Runtime::worker_index(), zs.len())
+                        });
 
                         // IR3: p(x,z) := p(y,w), u(w,r,z), q(x,r,y)
                         let ir3 = child.region("IR3", || {
@@ -199,7 +203,9 @@ fn main() -> Result<()> {
                                     |&(_r, _y), &z, &x| (x, z),
                                 )
                         });
-                        ir3.inspect(|zs: &OrdZSet<_, _>| println!("ir3: {}", zs.len()));
+                        ir3.inspect(|zs: &OrdZSet<_, _>| {
+                            println!("{}: ir3: {}", Runtime::worker_index(), zs.len())
+                        });
 
                         // IR4: p(x,z) := c(y,w,z), p(x,w), p(x,y)
                         let ir4_1 = child.region("IR4-1", || {
@@ -208,7 +214,9 @@ fn main() -> Result<()> {
                                 |&_w, &(y, z), &x| ((x, y), z),
                             )
                         });
-                        ir4_1.inspect(|zs: &OrdZSet<_, _>| println!("ir4_1: {}", zs.len()));
+                        ir4_1.inspect(|zs: &OrdZSet<_, _>| {
+                            println!("{}: ir4_1: {}", Runtime::worker_index(), zs.len())
+                        });
 
                         let ir4 = child.region("IR4-2", || {
                             ir4_1.index().join::<NestedTimestamp32, _, _, _>(
@@ -216,7 +224,9 @@ fn main() -> Result<()> {
                                 |&(x, _y), &z, &()| (x, z),
                             )
                         });
-                        ir4.inspect(|zs: &OrdZSet<_, _>| println!("ir4: {}", zs.len()));
+                        ir4.inspect(|zs: &OrdZSet<_, _>| {
+                            println!("{}: ir4: {}", Runtime::worker_index(), zs.len())
+                        });
 
                         // IR5: q(x,q,z) := q(x,r,z), s(r,q)
                         let ir5 = child.region("IR5", || {
@@ -225,7 +235,9 @@ fn main() -> Result<()> {
                                     (x, q, z)
                                 })
                         });
-                        ir5.inspect(|zs: &OrdZSet<_, _>| println!("ir5: {}", zs.len()));
+                        ir5.inspect(|zs: &OrdZSet<_, _>| {
+                            println!("{}: ir5: {}", Runtime::worker_index(), zs.len())
+                        });
 
                         // IR6: q(x,e,o) := q(x,y,z), r(y,u,e), q(z,u,o)
                         let ir6_1 = child.region("IR6_1", || {
@@ -243,7 +255,9 @@ fn main() -> Result<()> {
                             )
                         });
 
-                        ir6.inspect(|zs: &OrdZSet<_, _>| println!("ir6: {}", zs.len()));
+                        ir6.inspect(|zs: &OrdZSet<_, _>| {
+                            println!("{}: ir6: {}", Runtime::worker_index(), zs.len())
+                        });
 
                         let p = p.delta0(child).sum([&ir1, &ir3, &ir4]);
                         let q = q.delta0(child).sum([&ir2, &ir5, &ir6]);
@@ -252,8 +266,10 @@ fn main() -> Result<()> {
                     },
                 )
                 .unwrap();
-            outp.inspect(|zs: &OrdZSet<_, _>| println!("outp: {}", zs.len()));
-            outq.inspect(|zs: &OrdZSet<_, _>| println!("outq: {}", zs.len()));
+            outp.gather(0)
+                .inspect(|zs: &OrdZSet<_, _>| println!("outp: {}", zs.len()));
+            outq.gather(0)
+                .inspect(|zs: &OrdZSet<_, _>| println!("outq: {}", zs.len()));
         })
         .unwrap();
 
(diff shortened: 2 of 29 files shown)


Cargo.toml
```diff
@@ -22,6 +22,7 @@ impl-trait-for-tuples = "0.2"
 deepsize = "0.2.0"
 deepsize_derive = "0.1.2"
 textwrap = "0.15.0"
+fnv = "1.0"
```
Kixiron, line 25: If we want a fast hasher then [`rustc_hash`](https://docs.rs/rustc-hash/latest/rustc_hash/) is faster than `fnv`

src/circuit/runtime.rs
```diff
@@ -24,6 +25,18 @@ thread_local! {
...
+thread_local! {
+    // Reference to the `Runtime` that manages this worker thread or `None`
+    // if the current thread is not running in a multithreaded runtime.
+    static RUNTIME: RefCell<Option<Runtime>> = RefCell::new(None);
+
+    // 0-based index of the current worker thread within its runtime.
+    // Returns `0` if the current thread in not running in a multithreaded
+    // runtime.
+    static WORKER_INDEX: RefCell<usize> = RefCell::new(0);
+}
```
Kixiron, line 38: I'd rather this be passed as context via a runtime handle

src/circuit/runtime.rs
```diff
@@ -24,6 +25,18 @@ thread_local! {
...
+// Thread-local variables used to store per-worker context.
+thread_local! {
+    // Reference to the `Runtime` that manages this worker thread or `None`
+    // if the current thread is not running in a multithreaded runtime.
+    static RUNTIME: RefCell<Option<Runtime>> = RefCell::new(None);
+
+    // 0-based index of the current worker thread within its runtime.
+    // Returns `0` if the current thread in not running in a multithreaded
+    // runtime.
+    static WORKER_INDEX: RefCell<usize> = RefCell::new(0);
```
Kixiron, line 37: This should be `Cell`

src/circuit/runtime.rs
```diff
@@ -129,6 +142,27 @@ impl Runtime {
...
+    pub fn runtime() -> Option<Runtime> {
+        RUNTIME.with(|rt| rt.borrow().clone())
+    }
+
+    /// Returns 0-based index of the current worker thread within its
+    /// runtime.  For threads that run without a runtime, this method
+    /// returns `0`.
+    pub fn worker_index() -> usize {
+        WORKER_INDEX.with(|index| *index.borrow())
+    }
```
Kixiron, line 164: Again, I'd prefer these be context passed to the worker instead of globals

src/operator/communication/exchange.rs
```diff
@@ -557,51 +566,60 @@ where
         let _res = self
             .exchange
             .try_receive_all(self.worker_index, |x| (self.combine)(&mut combined, x));
+
         debug_assert!(_res);
```
Kixiron, line 570:
```suggestion
        let res = self
            .exchange
            .try_receive_all(self.worker_index, |x| (self.combine)(&mut combined, x));

        debug_assert!(res);
```

src/operator/communication/shard.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// # Performance considerations
+    ///
+    /// In the current implementation, the `shard` operator introduces a
+    /// synchronization barrier across all workers: its output at any worker
+    /// is only produced once input batches have been collected from all
+    /// workers.  This limits the scalability since a slow worker (e.g., running
+    /// on a busy CPU core or sharing the core with other workers) or uneven
+    /// sharding can slow down the whole system and reduce gains from
+    /// parallelization.
+    pub fn shard(&self) -> Stream<Circuit<P>, IB>
```
Kixiron, line 89: Shouldn't this be `.scatter()` to parallel `.gather()`?

Kixiron (changes requested):

Kixiron:
I'm very much against introducing more global state of any kind

ryzhyk:
> I'm very much against introducing more global state of any kind

We've discussed this several times before and I am convinced that thread-local store is a perfectly clean design in this case. First, this state is semantically associated with the current worker thread, i.e., it is per-thread state, as opposed to e.g., per-circuit or per-operator, and hence can legitimately live in the  thread-local store. Threading a handle instead will require adding an extra argument in various places.  In particular, any _user_ code that creates a nested circuit internally will have to take this handle to pass it to the new circuit.

src/operator/communication/shard.rs
ryzhyk: `shard` will act as `scatter` if the input collection happens to all be on one worker, so no need for a separate operator.

ryzhyk:
PS. I actually started implementing the handle threading logic initially and realized that introduces a dependency between circuits and runtime that currently doesn't exist: code in `circuit_builder.rs` is not aware of `Runtime`.

Kixiron:
Yah, I've attempted to do the same to the same results, I think we just need to give the circuit some notion of a runtime and then make a noop runtime that's passed for cases where there isn't one

src/operator/communication/shard.rs
Kixiron: I was meaning to rename it since scatter/gather is a common pair

ryzhyk:
I know it can be done, but I'm not sure why :)

src/operator/communication/shard.rs
ryzhyk: I'd say that `shard` combines scatter and gather in one on operator (in fact this is how it's implemented internally).

src/operator/distinct.rs
```diff
@@ -48,6 +51,15 @@ where
     /// This is equivalent to `self.integrate().distinct().differentiate()`, but
     /// is more efficient.
     pub fn distinct_incremental(&self) -> Stream<Circuit<P>, Z>
+    where
```
mbudiu-vmw, line 54: should the above comment move to the _inner function?

src/operator/communication/shard.rs
```diff
@@ -0,0 +1,284 @@
...
+    ) where
+        OB: Batch<Key = IB::Key, Val = IB::Val, Time = (), R = IB::R>,
+    {
+        builders.clear();
+
+        for _ in 0..nshards {
+            // We iterate over tuples in the batch in order; hence tuples added
+            // to each shard are also ordered, so we can use the more efficient
+            // `Builder` API (instead of `Batcher`) to construct output batches.
+            builders.push(OB::Builder::with_capacity((), batch.len() / nshards));
```
mbudiu-vmw, line 208: is there a problem if you guess the capacity incorrectly?

src/operator/communication/shard.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        let mut cursor = batch.cursor();
+
+        while cursor.key_valid() {
+            let batch_index = (hash32(cursor.key()) as usize) % nshards;
```
mbudiu-vmw, line 214: any reason to have the hash function be a parameter?

src/operator/communication/shard.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// Collect all shards of a stream at the same worker.
+    ///
+    /// The output stream in `receiver_worker` will contain a union of all
+    /// input batches across all workers.  The output streams in all other
+    /// workers will contain empty batches.
+    pub fn gather(&self, receiver_worker: usize) -> Stream<Circuit<P>, IB>
+    where
+        IB: Batch + Send,
+    {
+        match Runtime::runtime() {
```
mbudiu-vmw, line 156: so circuits are deterministic in terms of the zset values produced, but not necessarily in the representation of the zsets?
In particular, gather's output may be dependent on scheduling order?
was this true before?

mbudiu-vmw (approved):

benches/galen.rs
```diff
@@ -252,8 +266,10 @@ fn main() -> Result<()> {
                     },
                 )
                 .unwrap();
-            outp.inspect(|zs: &OrdZSet<_, _>| println!("outp: {}", zs.len()));
-            outq.inspect(|zs: &OrdZSet<_, _>| println!("outq: {}", zs.len()));
+            outp.gather(0)
```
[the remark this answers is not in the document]
ryzhyk: Not necessarily.  A multithreaded consumer (e.g., another DBSP circuit) might prefer to get data from many worker threads in parallel instead of a gathered input.  In DD you typically invoke user callback from a final `inspect` operator attached to each worker.

src/circuit/runtime.rs
```diff
@@ -129,6 +142,27 @@ impl Runtime {
...
 
+    /// Returns a reference to the multithreaded runtime that
+    /// manages the current worker thread, or `None` if the thread
+    /// runs without a runtime.
+    ///
+    /// Worker threads created by the [`Runtime::run`] method can access
+    /// the services provided by this API via an instance of `struct Runtime`,
+    /// which they can obtain by calling `Runtime::runtime()`.  DBSP circuits
+    /// created without a managed runtime run in the context of the client
+    /// thread.  When invoked by such a thread, thos method returns `None`.
```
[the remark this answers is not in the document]
ryzhyk: Yes, because that thread will run in parallel with the client thread.

src/trace/ord/zset_batch.rs
```diff
@@ -109,20 +97,6 @@ where
     const CONST_NUM_ENTRIES: Option<usize> = <OrderedLeaf<K, R>>::CONST_NUM_ENTRIES;
 }
 
-impl<K, R> HasZero for OrdZSet<K, R>
```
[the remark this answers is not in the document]
ryzhyk: We now have a generic implementation in `trace.rs`

src/operator/distinct.rs
ryzhyk: `_inner` is not `pub`, so it won't be visible there.

src/operator/communication/shard.rs
ryzhyk: It won't crash, but may waste memory or take a perf hit.

src/operator/communication/shard.rs
ryzhyk: You mean to _not_ be a parameter, so it can be supplied? Definitely, but probably can wait for a use case.

src/operator/communication/shard.rs
ryzhyk: Because we call `consolidate`, the output ends up converted to the canonical representation.  But even if it didn't the trace would still yield key/value pairs in the canonical order. So it should all be deterministic.

PR #121, opened 2022-07-22
[nexmark] Fix confusion of micro and milli-seconds.

absoludity:
After people pointed out that 100ms between each event was obviously odd as a default (even if configurable), I rechecked and found I'd missed a point where the java code specifically uses *micro* seconds rather than *milli* seconds.

The actual fix was obviously trivial (and now matches the upstream java code), but testing the batching of results was trickier... I've injected an optional iterator for wallclock times that can be used in tests. Not sure if there's a better way. See what you think. FWIW, the working test then found another small issue (the change from `<` to `<=` when including events in a batch).

### DIFF
--- a/benches/nexmark/generator/mod.rs
+++ b/benches/nexmark/generator/mod.rs
@@ -8,7 +8,7 @@ use anyhow::{Context, Result};
 use bids::CHANNELS_NUMBER;
 use cached::SizedCache;
 use rand::Rng;
-use std::time::SystemTime;
+use std::{ops::Range, time::SystemTime};
 
 mod auctions;
 mod bids;
@@ -33,13 +33,10 @@ pub struct NexmarkGenerator<R: Rng> {
     /// Initialised to the current system time when the first event is
     /// emitted.
     wallclock_base_time: Option<u64>,
-}
 
-pub fn wallclock_time() -> u64 {
-    SystemTime::now()
-        .duration_since(SystemTime::UNIX_EPOCH)
-        .unwrap()
-        .as_millis() as u64
+    /// An optional iterator that provides wallclock timestamps in tests.
+    /// This is set to None by default.
+    wallclock_iterator: Option<Range<u64>>,
 }
 
 impl<R: Rng> NexmarkGenerator<R> {
@@ -50,12 +47,25 @@ impl<R: Rng> NexmarkGenerator<R> {
             bid_channel_cache: SizedCache::with_size(CHANNELS_NUMBER),
             events_count_so_far: 0,
             wallclock_base_time: None,
+            wallclock_iterator: None,
         }
     }
 
     // Reset the generator to begin at event 0 again.
     pub fn reset(&mut self) {
         self.events_count_so_far = 0;
+    }
+
+    // Returns the real current wallclock time or a specific range of numbers
+    // in tests.
+    pub fn wallclock_time(&mut self) -> u64 {
+        match &mut self.wallclock_iterator {
+            Some(i) => i.next().unwrap(),
+            None => SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .unwrap()
+                .as_millis() as u64,
+        }
     }
 
     /// Set the wallclock base time explicitly rather than leaving it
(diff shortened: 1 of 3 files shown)


benches/nexmark/nexmark_dbsp_source.rs
```diff
@@ -87,19 +87,18 @@ where
...
         // is still in the future, we sleep until we can emit it.
         let next_event = next_event.unwrap();
-        let wallclock_time_now = wallclock_time();
+        let mut wallclock_time_now = self.generator.wallclock_time();
         if next_event.wallclock_timestamp > wallclock_time_now {
-            sleep(Duration::from_millis(
-                next_event.wallclock_timestamp - wallclock_time_now,
-            ));
+            let millis_to_sleep = next_event.wallclock_timestamp - wallclock_time_now;
+            sleep(Duration::from_millis(millis_to_sleep));
```
Kixiron, line 93: @ryzhyk this is the kind of situation where I think that being able to pause and resume operators would be really valuable, we could do computation while things like network delays, disk delays, etc. are all occurring within other operators

benches/nexmark/generator/mod.rs
```diff
@@ -58,6 +56,18 @@ impl<R: Rng> NexmarkGenerator<R> {
...
+    // Returns the real current wallclock time or a specific range of numbers
+    // in tests.
+    pub fn wallclock_time(&mut self) -> u64 {
+        match &mut self.wallclock_iterator {
+            Some(i) => i.next().unwrap(),
+            None => SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .unwrap()
+                .as_millis() as u64,
+        }
```
Kixiron, line 68:
```suggestion
        self.wallclock_iterator
            .as_mut()
            .map(|wall| wall.next().unwrap())
            .unwrap_or_else(|| {
                SystemTime::now()
                    .duration_since(SystemTime::UNIX_EPOCH)
                    .unwrap()
                    .as_millis() as u64
            })
```

Kixiron (approved):
LGTM

benches/nexmark/nexmark_dbsp_source.rs
ryzhyk: I've been thinking about this lately.  Clearly, operators shouldn't block.  We discussed this with @absoludity , and the above implementation is just a temporary solution until we have proper infrastructure in place. Sleeping is kinda ok in this case because the circuit has exactly one source operator, and if the source does not have any data, the circuit can just pause (except that it could use the idle time to perform trace compaction).

We already have a way to pause the operator via the `is_async`/`ready`/`register_ready_callback` API in `trace Operator`; however pausing is not the right solution for either operators that perform I/O or operators like the one above that just need to sleep. A paused operator will prevent the entire circuit from completing the clock cycle and making forward progress.  Imagine a circuit with two source operators, one reading from file and one from a network socket.  At any given time only one of them may have data, so we shouldn't stall the circuit just because the other operator is empty.   Instead, whenever at least one of the sources has data available, the computation should go forward, with the other source returning an empty batch.  However, when both sources are empty, there is no point in keeping the circuit busy while doing no useful work, so in this case we want to pause the execution, perhaps only doing trace compaction, until at least one of the sources has new inputs available.

In summary, there are two levels of control we need exercise: within a single clock cycle the async operator mechanism can be used to tell the scheduler to postpone the evalution of an operator. As far as I can tell, this is really only useful for worker to worker communication (which is what this mechanism was created for in the first place). A different mechanism is needed to schedule the beginning of the next clock cycle, based on the availability of data across all sources or some other user-defined policy (the client application may be in the best position to decide when is the best time to start cranking the handle).

I have an idea for a solution and will write an RFC soon-ish.

benches/nexmark/generator/mod.rs
absoludity: My brain seems to parse the simple match more easily here (no mapping or extra unwrapping), so I'm interested to know if that's just my less experience or whether there's a benefit to not using a match even in this simple case?

ryzhyk: I'm with @absoludity on this one. Sometimes a match can be replaced with a single operator, eg map or unwrap_or_else. But in this case a chain of two transformations is harder to parse imo.

PR #122, opened 2022-07-22
Move Nexmark source into src/operator/nexmark.

absoludity:
Note: I don't think this is the right solution but want to discuss.

While writing my first nexmark queries to benchmark, I've found the following...

## Problem

Currently on main, I can run the unit tests for the nexmark generator with:

```
cargo test --bench nexmark
...
test result: ok. 66 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.40s
```

These are important to ensure the generator is correct.

If I update to set the nexmark benchmark to not use the default harness:

```
diff --git a/Cargo.toml b/Cargo.toml
index 4a30cfa..51294b0 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ harness = false
 
 [[bench]]
 name = "nexmark"
+harness = false
```
then not only can I can no longer run the unit-tests (since it now tries to run the currently non-existent main function), but it also views a lot of the imports in my test modules as unused (I assume since the test functions are never called by the harness):

```
warning: unused import: `super::*`
   --> benches/nexmark/generator/auctions.rs:130:9
    |
130 |     use super::*;
    |         ^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default

...
...
warning: unused import: `circuit::Root`
   --> benches/nexmark/nexmark_dbsp_source.rs:126:16
    |
126 |     use dbsp::{circuit::Root, trace::ord::OrdZSet, trace::Batch};
    |                ^^^^^^^^^^^^^

error[E0601]: `main` function not found in crate `nexmark`
  --> benches/nexmark/main.rs:17:25
   |
17 | mod nexmark_dbsp_source;
   |                         ^ consider adding a `main` function to `benches/nexmark/main.rs`

For more information about this error, try `rustc --explain E0601`.
warning: `dbsp` (bench "nexmark") generated 8 warnings
error: could not compile `dbsp` due to previous error; 8 warnings emitted
```

If I define a main function for the bench:

```
diff --git a/benches/nexmark/main.rs b/benches/nexmark/main.rs
index 52427fc..b61e8bc 100644
--- a/benches/nexmark/main.rs
+++ b/benches/nexmark/main.rs
@@ -10,8 +10,13 @@
 //! it will allow us later to extend it to support multiple data sources in
 //! parallel when DBSP can be scaled.
 #![feature(is_some_with)]
+use anyhow::Result;
 
 mod config;
 mod generator;
 mod model;
 mod nexmark_dbsp_source;
+
+fn main() -> Result<()> {
+    Ok(())
+}
```

then it no longer errors but still views all the unit tests as unused code (and I have no way to run them).

## Possible Solutions

- I think the ideal long-term would be to move the nexmark data generation to a separate crate, but that might not be worth the time at this stage,
- IMO, the ideal solution right now would be if I can update my nexmark test harness to just call the `libtest` harness when running unit tests, so nothing would need to change. But this [appears to be harder than it sounds](https://www.infinyon.com/blog/2021/04/rust-custom-test-harness/#collect-all-integration-tests) (haven't dug deeply but did take a look at the [libtest code, now in rust core](https://github.com/rust-lang/rust/tree/master/library/test) and don't yet see where test collection happens - perhaps in the `cargo test` command itself. Need to dig deeper).
- So a very much less than ideal solution in this PR is to move the Nexmark source operator (and all its dependents and tests) into the `src/operators` module for discussion.

Is it instead possible for us to create another repo (maybe under vmware-labs or somewhere else) that is quick?
Let me know if you can see a better way... If we do consider merging this, there are a few more things we may want to do first (eg. not sure if you are OK importing `anyhow` in non-dev code). I'll keep looking to see if we can use (or call) the default harness for unit tests while having our custom one for benches.

Thanks for any help.


ryzhyk:
I think it is ok to keep benchmark-specific code under `src`, as long as it is feature gated.  This is probably the best short-term solution.  I prefer `src/nexmark` to `src/operator/nexmark` though.  As we discussed the other day, there shouldn't really be a special nexmark source operator, instead the nexmark generator should use the (currently non-existing) DBSP API to feed data to the circuit.  I also agree that a separate crate would be a cleaner solution long-term, for the non-DBSP-specific part of the nexmark generator.  I suspect it's totally fine to keep that repo under your personal github space (github.com/absoludity).  The OSS approval process takes a few weeks and seems like an overkill in this case.

In summary, I am ok with the approach in this PR as long as we feature-gate the nexmark generator.   And if you want to spend a bit more time to create a separate crate for the non-DBSP-specific nexmark code, that's great too -- your call :)

absoludity:
> I think it is ok to keep benchmark-specific code under `src`, as long as it is feature gated. This is probably the best short-term solution. I prefer `src/nexmark` to `src/operator/nexmark` though. As we discussed the other day, there shouldn't really be a special nexmark source operator, instead the nexmark generator should use the (currently non-existing) DBSP API to feed data to the circuit. I also agree that a separate crate would be a cleaner solution long-term, for the non-DBSP-specific part of the nexmark generator. I suspect it's totally fine to keep that repo under your personal github space (github.com/absoludity). The OSS approval process takes a few weeks and seems like an overkill in this case.
> 
> In summary, I am ok with the approach in this PR as long as we feature-gate the nexmark generator. And if you want to spend a bit more time to create a separate crate for the non-DBSP-specific nexmark code, that's great too -- your call :)

Great, thanks @ryzhyk . In that case, I'll avoid churn and leave it where it is for now. putting up with the inconvenience (to me only at this point) of having to switch `harness = true` to be able to run tests for the generator: this so I can actually do some PRs showing the initial queries right now (keen to do these), but also give myself some friction to pull it out into a separate crate soon.

Cheers.

absoludity:
Argh, no - that doesn't quite work due to all the warnings. I'll take a closer look and see.

PR #123, opened 2022-07-23
[Don't merge] Work in progress on persistence

gz:
Opened a PR for now so I can dump information somewhere; but this isn't ready to merge. If we are fine with the current overheads I measured with the OrdZSet experiment (see below) then, my next step will be to make a persistent trace by porting this code over from the persistent OrdZSet into something that implements Trace functionality instead.

# Notes on using RocksDB

- RocksDB seek calls always start from the beginning, whereas our Cursor seek
  does not
  - The fix is just to maintain the last Key in the Cursor struct
  - Not sure if this has any downsides for the RocksDB seek performance,
    probably doesn't matter (it keeps an index to the right block)

- Comparison of complex keys
  - By default RocksDB uses bytewise comparison (of the serialized binary data).
  - If we have complex Rust keys that won't work
  - We can set a custom comparator function in RocksDB that calls into the Rust Ord implementation
  - The downside is we need to deserialize the key every time this is called
  - We might want to measure the impact of this and if it turns out to be
    significant we can avoid it in ceqrtain well-specified but common(?) cases,
    e.g., if the key is a usize then the serialized bytes (using big-endian
    encoding) preserves lexicographic ordering with bincode

- Gets
  - It seems that `get_pinned` is preferrable over get as it avoids a `Vec`
    allocation/deallocation on every lookup.

- Merging:
  - One way is to take the files of one (the bigger) batch and use
    `ingest_external_file` for creating a new "batch", then use db.merge() to
    merge in the other (smaller) half?
  - The alternative is to just do the merge in rust and write out a new DB
  - Not sure which one is faster or better...

- One database per DS/Trace or one DB with a ColumnFamily per DS/Trace?
  - Both can give a unique key-space
  - We can share a Cache in both cases (I think)
  - ColumFamily approach:
    - Odd API requires to name all ColumnFamilys on DB::open and fails if not done
    - There is a separate call that you need to use to get a Vec of all CFs
    - Then you supply that to DB::open and hope no CF got created meanwhile?
    - This seems strange: what about liveness if lots of CFs are created etc.
    - I guess the workaround is to open the DB once in the program and access
      through a singleton
    - Haven't seen anything if this feature is made for having 10s, 100s, 1000s
      of CFs?
  - Many databases approach:
    - Makes me think CFs is not what we want, maybe need one DB per DS/Trace?
    - One advantage of that is once we wrote it we can open the DB as read-only
    - Not sure if this has more overhead in the end (as opposed to ColumnFamily
      approach) for perf or mem or disk consumption
    - There are atomic operations across columnfamilies but not across different
      databases

- ThreadingMode: there are two ways, we currently use `MultiThreaded`
  - SingleThreaded apparently works multi-threaded too, so it's a bit confusing,
    but...
  - SingleThreaded takes a &mut self to create a ColumnFamily whereas
    MuliThreaded takes a &self and uses the DB's internal RWLock
  - Since we have a singleton RocksDB instance in the code, I just used
    MultiThreaded
  - According to docs: MultiThreaded should be used if you want to do
    multi-threaded access to a single ColumnFamily
  - We likely don't need this, so we can investigate if there is a perf benefit
    going from MultiThreaded to SingleThreaded
  - Note aside: we found when porting LevelDB to NrOS that we got contended on
    the LevelDB rwlock, replacing it with a better variant helped to scale
    reads...

# Notes on RocksDB Performance
  - Benchmarks: Just a simple benchmark, OrdZSet with u64 keys and two different
    size 4K and 16M keys
  - Run using `cargo criterion` (need to do `cargo install cargo-criterion`)
  - Results are put in `target/criterion`
  - Uploaded results here:
  - Some changes from default that I did:
    - Added a 2 GiB block cache (default is none or 8 MiB?)
    - Disabled compression
    - Made sure to use a RAM disk (e.g., `tmpfs`)
    - Ran into `No space left on device` during benchmarking, because tmpfs was
      small; increased with: `sudo mount -o remount,size=12G /tmp` to 12 GiB
    - Ran into some open file limits, so I set
      `global_opts.set_max_open_files(9000);` and `ulimit -n 9000` in bash
      which resolved the problem
    - More things to explore:
      - Disable the write-ahead-log
      - Can use more / higher prio background threads (for compaction)
      - There are some options to optimize for point queries
      - There are different cache types (we use LRU atm.)
      - There is a tuning guide that has some tips:
        <https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide>
  - Observations:
    - All results are here: <https://gz.github.io/database-stream-processor/criterion-2022-07-22/reports/index.html>
    - The persistent OrdZSetBuilder is 343x slower for 4k entries (2 ms vs 6us)
    - OrdZSetBuilder for 16M entries: it's 32x slower 4.2s vs. 129ms
    - Iterating through a OrdZSet with 4K entries is 125x slower; Iterating through 16M entries is 51x slower
    - Uniform random seek in 4k entries 8x slower, 16M entries
    - Uniform random seek 16M entries 29x faster in the persistent version (!)
      - Not sure why, maybe the OrdZSet seek is O(n), RocksDB will use its index?
    - zipf random seek 4k entries 14x slower
    - zipf random seek 16M entries 5x faster (!)
      - Not sure why, maybe the OrdZSet seek() is O(n), RocksDB will use its index?


gz:
ehm sorry I got the branches in the wrong order here; the github UI is confusing will close and reopen

PR #126, opened 2022-07-24
[nexmark] Move nexmark from benches to src.

absoludity:
See the closed #122 both for the problem and the agreed solution with @ryzhyk.

I had to create a new PR since GH closed the old one based on the target branch being closed (rather than re-targeting main).


absoludity:
Pushed to personal repo by mistake, sorry for the noise. Now at #127.

PR #128, opened 2022-07-25
[nexmark] Initial play with nexmark q1 query

absoludity:
Just a WIP with some notes and questions in case you have feedback.

As mentioned inline, a few TODOs:
- I still need to enable cmdline args to run just specific tests
- Learn from the java implementation how they're calculating the CPU cores and times for the report.
- Start adding the other (more interesting) queries.
- Provide a way to stop the test once a certain number of events is processed (as our implementation sends empty zsets after the max events is found, rather than returning an error... perhaps we should switch).

As it is, with this PR you can see that on my machine (Dell precision 5540), the DBSP q1 circuit exhausts the data source emitting 1M events per second, processing around 1_000 events per ms:

```
$ cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=1000000
...
129628331 events processed in 129729ms
129629331 events processed in 129730ms
129630330 events processed in 129731ms
129631329 events processed in 129731ms
129632327 events processed in 129732ms
129633326 events processed in 129734ms
129635326 events processed in 129737ms
129637324 events processed in 129739ms
129639322 events processed in 129741ms
129642321 events processed in 129745ms
129645319 events processed in 129748ms
$

```


benches/nexmark/main.rs
```diff
@@ -1 +1,67 @@
...
+use rand::prelude::ThreadRng;
 
+// TODO: Enable running specific tests
+fn main() -> Result<()> {
+    let nexmark_config = NexmarkConfig::parse();
+    let generator_config = GeneratorConfig::new(nexmark_config, 0, 0, 0, 0);
+
+    let cpu_profiler = CPUProfiler::new();
+    let root = Root::build(|circuit| {
+        // TODO(absoludity): CPUProfiler is not currently used in any of the
```
ryzhyk, line 25: The CPU profiler is useful for optimizing the performance, as it attributes CPU time to specific operators. I need to package it a bit better to make it easier to use, along with the memory profiler.  Both are fairly of rudimentary, but since we won't be building anything more serious in the immediate future, we should make it easy to consume the existing implementation.

But it is not meant for reporting the overall CPU/memory consumption of a benchmark.  We need some other tools for that.  I personally use `/usr/bin/time`, which gives you the three important numbers: memory, wall clock time, and CPU utilization. I haven't looked into what's the best way to do this in Rust.

benches/nexmark/main.rs
```diff
@@ -1 +1,67 @@
...
+};
+use rand::prelude::ThreadRng;
 
+// TODO: Enable running specific tests
+fn main() -> Result<()> {
+    let nexmark_config = NexmarkConfig::parse();
+    let generator_config = GeneratorConfig::new(nexmark_config, 0, 0, 0, 0);
+
+    let cpu_profiler = CPUProfiler::new();
+    let root = Root::build(|circuit| {
```
ryzhyk, line 24: We should wrap this in an instance of `Runtime`, to enable multithreading.

benches/nexmark/main.rs
```diff
@@ -1 +1,67 @@
...
+                        .as_millis()
+                );
+                // TODO: send signal through channel to stop iterating once
+                // sufficient data for test is processed?
+            }
+        });
+    })
+    .unwrap();
+
+    for _ in 0..100_000 {
```
ryzhyk, line 63: I'll work on a circuit lifecycle API, so we won't need to hardcode fixed numbers of iterations.

ryzhyk (approved):
This clearly suffers from the lack of high-level APIs for profiling, circuit lifecycle management, and feeding data from external sources.  I feel like this is becoming too much of a pain point, so I'll work on it next.

mbudiu-vmw:
I expect people always use separate client machines to drive benchmarks, so that's the first api to consider. Unfortunately it will probably need to be dynamically typed, like ddvalue.

benches/nexmark/main.rs
absoludity: OK, I'll have a look at `Runtime` today. Not yet clear to me how a continuous source like the Nexmark one can be run in different threads, but maybe that's related to the "feeding data from external sources" pain-point comment.

benches/nexmark/main.rs
absoludity: > We need some other tools for that. I personally use `/usr/bin/time`, which gives you the three important numbers: memory, wall clock time, and CPU utilization. I haven't looked into what's the best way to do this in Rust.

Yeah, I was using `time cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=1000000` initially to get an idea (which includes any required compile too, not great). I'll look for a way to do so from within the harness.

benches/nexmark/main.rs
absoludity: Great, thanks. Noticed that sometimes, given the way we're batching whatever is ready, this estimate works if CPU keeps up, but as soon as the CPU is busy with other things and a backlog generates, it seems to cascade (as the batches (and mem?) grow massively)

absoludity (commented):
Thanks @ryzhyk . Noticed that you +1'd this. I'm happy to land it and keep iterating, but I do first need #127 +1'd to land that first.

benches/nexmark/main.rs
ryzhyk: We'd have to setup the circuit to run the generator in worker thread 0 only. But yeah, the "feeding data from external sources" feature will solve this properly.

benches/nexmark/main.rs
absoludity: Hey @ryzhyk , do you know any way I can work around this until the circuit lifecycle API is available? Specifically, AFAICT, I either need to be able to:
1. communicate from the closure within the circuit builder (from `output.inspect()`) that the desired number of events have been processed, so that I can stop calling `root.step()`, or
2. have the circuit notice that the source has reached a fixed point and the output is no longer changing (hmm, though maybe this can't be used given delay operators etc., not sure).

As per my comment, I was initially thinking I'd use a channel as a workaround to communicate out (given any var captured by the circuit builder closure as a `&mut` won't be accessible after the circuit is built), but I hadn't realised it's not actually async (as in, I'm manually calling `root.step()` rather than a scheduler in another thread or similar.

No worries if you don't have time, just in case you've something obvious I've missed. Thanks.

ryzhyk: I think you can just create a variable of type `Rc<Cell<bool>>` and `set` it inside the source operator when all outputs have been generated. You can check the value of the variable inside the `step()` loop, aborting on `true`.

And yes, this is something that the I/O API will solve, for both single-threaded and multi-threaded cases (i.e., when using `Runtime`).

absoludity: Thanks @ryzhyk, yep, that works fine. Now I'm needing to create a macro that generates the circuits for each query it seems (given the queries have different return types). Anyway, will have that ready tomorrow I hope.

PR #129, opened 2022-07-26
Overflow error.

ryzhyk:
A typo in `layers/mod.rs` was causing overflow in a bit shift operation.


Kixiron (approved):

src/trace/layers/mod.rs
```diff
@@ -152,7 +152,7 @@ pub fn advance<T, F: Fn(&T) -> bool>(slice: &[T], function: F) -> usize {
             let mut step = 1;
             while index + step < slice.len() && function(&slice[index + step]) {
                 index += step;
-                step <<= step;
+                step <<= 1;
```
Kixiron, line 155: Can you add a test for this too?

PR #131, opened 2022-07-26
[nexmark] Add q2 and refactor for better testing.

absoludity:

Adding the Nexmark q2 query itself was trivial, but unit testing it was not, but is now. Up until now, I've been testing the nexmark source only, so it was fine to test against the generated events of a (non-random) Nexmark generator. But the q2 query (and subsequent ones that I'll be implementing) require a lot of iterations to test (and even then, not obviously), so this PR:

1. Updates the `NexmarkSource` struct to expect an object that implements the `EventGenerator` trait (rather than expecting a `NexmarkGenerator` always.
2. Adds the test-only `CannedEventGenerator` and uses it to test the `q2` query succinctly,
3. Removes the test-only code from the actual `NexmarkGenerator` for creating a generator that emits wallclock times from a range and put it into another test-only generator `RangedTimeGenerator`.

This was a pain (as in, felt unproductive to be refactoring test code so much for a single query), but should make subsequent queries simple to test as well.

### DIFF
--- a/src/nexmark/queries/q2.rs
+++ b/src/nexmark/queries/q2.rs
@@ -1,0 +1,130 @@
+use super::NexmarkStream;
+use crate::{nexmark::model::Event, operator::FilterMap, Circuit, OrdZSet, Stream};
+
+/// Selection
+///
+/// Find bids with specific auction ids and show their bid price.
+/// See https://github.com/nexmark/nexmark/blob/v0.2.0/nexmark-flink/src/main/resources/queries/q2.sql
+const AUCTION_ID_MODULO: u64 = 123;
+
+pub fn q2(input: NexmarkStream) -> Stream<Circuit<()>, OrdZSet<(u64, usize), isize>> {
+    input
+        .filter(|event| match event {
+            Event::Bid(b) => b.auction % AUCTION_ID_MODULO == 0,
+            _ => false,
+        })
+        .map(|bid_event| match bid_event {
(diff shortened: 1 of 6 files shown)


src/nexmark/queries/q2.rs
```diff
@@ -0,0 +1,130 @@
...
+
+/// Selection
+///
+/// Find bids with specific auction ids and show their bid price.
+/// See https://github.com/nexmark/nexmark/blob/v0.2.0/nexmark-flink/src/main/resources/queries/q2.sql
+const AUCTION_ID_MODULO: u64 = 123;
+
+pub fn q2(input: NexmarkStream) -> Stream<Circuit<()>, OrdZSet<(u64, usize), isize>> {
+    input
+        .filter(|event| match event {
```
ryzhyk, line 12: This can be more efficiently implemented by replacing `filter().map()` with `flat_map()`, using `Option<(auction, price)>` as output iterator type.

absoludity: Done, thanks.

Kixiron: `.filter_map()` would be a better translation

Kixiron (approved):

src/nexmark/queries/q2.rs
ryzhyk: We no longer have `filter_map`.  It's a special case of `flat_map` using `Option<>` as output type.

PR #132, opened 2022-07-27
[nexmark] Add failing q3 test until I better understand the join.

absoludity:

Help :) . I've written a test for the q3 nexmark query, which features an incremental join, but... I don't yet know enough about DBSP to see how I would achieve an incremental join with the single input stream. I've been through a few of the tests in `join.rs`, but don't see anything similar (most similar is the `join_trace_test`, which uses the input for two different operators, but it also has a recursive input (which I don't think I'd need here). I assume I need to split the input into auctions and people then join those (after indexing people so they can be looked up while going through the auctions?), but I'm still searching there.

Anyway, I've left a few questions inline. I'd be happy just to be pointed in the right direction - I may need to go back and keep working through the spec.pdf, not sure.

No rush, in the mean-time, I'll start improving the test harness itself.

### DIFF
--- a/src/nexmark/queries/q3.rs
+++ b/src/nexmark/queries/q3.rs
@@ -6,7 +6,7 @@ use crate::{nexmark::model::Event, operator::FilterMap, Circuit, OrdZSet, Stream
 /// Who is selling in OR, ID or CA in category 10, and for what auction ids?
 /// Illustrates an incremental join (using per-key state and timer) and filter.
 ///
-/// From https://github.com/nexmark/nexmark/blob/v0.2.0/nexmark-flink/src/main/resources/queries/q3.sql
+/// From [Nexmark q3.sql](https://github.com/nexmark/nexmark/blob/v0.2.0/nexmark-flink/src/main/resources/queries/q3.sql)
 ///
 /// CREATE TABLE discard_sink (
 ///   name  VARCHAR,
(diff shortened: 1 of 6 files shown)


src/nexmark/queries/q3.rs
```diff
@@ -0,0 +1,160 @@
...
+/// Illustrates an incremental join (using per-key state and timer) and filter.
+///
+/// See https://github.com/nexmark/nexmark/blob/v0.2.0/nexmark-flink/src/main/resources/queries/q3.sql
+
+const STATES_OF_INTEREST: &[&str] = &["OR", "ID", "CA"];
+const CATEGORY_OF_INTEREST: usize = 10;
+
+pub fn q3(
+    input: NexmarkStream,
+) -> Stream<Circuit<()>, OrdZSet<(String, String, String, u64), isize>> {
```
ryzhyk, line 16: I guess the output could also use static strings in this case.  In general we probably want to use interned strings to represent SQL strings. We rely on interning a lot in DDlog via the [`internment`](https://crates.io/crates/internment) crate. It saves both CPU and memory.

ryzhyk (approved):
The `q3` test is passing now.  Happy to go over it in detail.

ryzhyk:
PS. Would it make sense to keep the original SQL code with each query as a comment?

absoludity (commented):
> The `q3` test is passing now. Happy to go over it in detail.

Great, thanks for the help there. It makes much more sense now:
1. Construct two streams from the input, both *indexed by the same value* (in this case, the person id).
2. Join the two streams (that are indexed on the same value), choosing the output.

I'd not properly understood how either of those worked in DBSP before.

> PS. Would it make sense to keep the original SQL code with each query as a comment?

Yep, I'll do that - I've included the URL to it, but it'll  save having to open the URL and compare.

Thanks again for the help there!

PR #136, opened 2022-07-29
[nexmark] Add q4 with multiple aggregations.

absoludity:
I'm guessing there's a more succinct way to do the multiple aggregations - let me know.

Also interested to know if I should be using incremental versions, and if so, perhaps I'll wait until have a basic bench harness to verify the improvement (similar for the interning of strings).

### DIFF
--- a/src/nexmark/queries/q4.rs
+++ b/src/nexmark/queries/q4.rs
@@ -78,6 +78,8 @@ pub fn q4(input: NexmarkStream) -> Q4Stream {
 
     // winning_bids_by_category: once we have the winning bids, we don't
     // need the auction ids anymore.
+    // TODO: We can optimize this given that there are no deletions, as DBSP
+    // doesn't need to keep records of the bids for future max calculations.
     let winning_bids_by_category: Stream<Circuit<()>, OrdZSet<(usize, usize), isize>> =
         bids_for_auctions_indexed.aggregate_incremental(|&key, vals| -> (usize, usize) {
             // `vals` is sorted in ascending order for each key, so we can
--- a/src/nexmark/queries/q4.rs
+++ b/src/nexmark/queries/q4.rs
@@ -49,11 +49,11 @@ pub fn q4(input: NexmarkStream) -> Q4Stream {
         _ => None,
     });
 
+    type BidsAuctionsJoin =
+        Stream<Circuit<()>, OrdZSet<((u64, usize, u64, u64), (usize, u64)), isize>>;
+
     // Join to get bids for each auction.
-    let bids_for_auctions: Stream<
-        Circuit<()>,
-        OrdZSet<((u64, usize, u64, u64), (usize, u64)), isize>,
-    > = auctions_by_id.join::<(), _, _, _>(
+    let bids_for_auctions: BidsAuctionsJoin = auctions_by_id.join::<(), _, _, _>(
         &bids_by_auction,
         |&auction_id, &(category, a_date_time, a_expires), &(bid_price, bid_date_time)| {
             (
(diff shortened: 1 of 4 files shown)


src/nexmark/queries/q4.rs
```diff
@@ -0,0 +1,208 @@
...
+            } else {
+                None
+            }
+        },
+    );
+
+    // winning_bids_by_category: once we have the winning bids, we don't
+    // need the auction ids anymore.
+    let winning_bids_by_category: Stream<Circuit<()>, OrdZSet<(usize, usize), isize>> =
+        bids_for_auctions_indexed.aggregate(|&key, vals| -> (usize, usize) {
```
ryzhyk, line 80: We must use incremental version of `aggregate` here for correctness: since all previous operators are incremental, inputs to `aggregate` are changes to the `bids_for_auctions_indexed` relation.  The non-incremental `aggregate` will aggregate values in the last delta only.

src/nexmark/queries/q4.rs
```diff
@@ -0,0 +1,208 @@
...
+            let max = vals.drain(..).map(|(bid, _)| bid).max().unwrap();
```
ryzhyk, line 81: As an optimization, you can take advantage of the fact that `vals` is sorted in ascending order, so you can just pick the last element without iterating over the entire slice.

src/nexmark/queries/q4.rs
```diff
@@ -0,0 +1,208 @@
...
+                (auction_id, category, a_date_time, a_expires),
+                (bid_price, bid_date_time),
+            )
+        },
+    );
+
+    // Filter out the invalid bids while indexing.
+    let bids_for_auctions_indexed = bids_for_auctions.flat_map_index(
+        |&((auction_id, category, a_date_time, a_expires), (bid_price, bid_date_time))| {
+            if bid_date_time >= a_date_time && bid_date_time <= a_expires {
```
ryzhyk, line 69: I should implement an incremental version of `join_range`, so we can do this filtering as part of the previous join. I created an issue for this (#137 )

src/nexmark/queries/q4.rs
```diff
@@ -0,0 +1,208 @@
...
+    // need the auction ids anymore.
+    let winning_bids_by_category: Stream<Circuit<()>, OrdZSet<(usize, usize), isize>> =
+        bids_for_auctions_indexed.aggregate(|&key, vals| -> (usize, usize) {
+            let max = vals.drain(..).map(|(bid, _)| bid).max().unwrap();
+            (key.1, *max)
+        });
+    let winning_bids_by_category_indexed = winning_bids_by_category.index();
+
+    // Finally, calculate the average winning bid per category.
+    winning_bids_by_category_indexed.aggregate(|&key, vals| -> (usize, usize) {
```
ryzhyk, line 87: This should also use `aggregate_incremental`.

Also, let's add a TODO comment here to use linear aggregation when it's ready (#138).

ryzhyk (changes requested):

src/nexmark/queries/q4.rs
absoludity: OK, I'm going to try to update my test to fail to better understand this. Thanks.

absoludity (commented):
Thanks for the feedback!

src/nexmark/queries/q4.rs
absoludity: PTAL. I've updated to use `aggregate_incremental` in both instances. I didn't spend too long on it, but wasn't able to make my test fail only for the non-incremental version: Is the last delta calculated for each batch? I tried adding an extra `root.step.unwrap`, but then it fails both for the incremental and non-incremental - which surprised me, I assumed that the next step would just not have any deltas and so the expected value would not have changed.

ryzhyk: You do need more than one iteration to observe the difference between incremental and non-incremental implementation. Given a query `q` and an input `i` at time 0, the non-incremental implementation computes `q(i)`, while the incremental one computes `q(i) - q(0)`, and since for this query`q(0) = 0`, the two outputs are the same.

Furthermore, you need the second iteration of your terst to take a non-empty input.  If your first input is `i` and your second input is `0`, then the non-incremental query computes `q(i)` at the first iteration and `q(0) = 0` at the second iteration.  The incremental query computes `q(i) - q(0) = q(i)` at the first iteration and `q(i+0) - q(i) = q(i) - q(i) = 0` at the second iteration.  In other words, if your input doesn't change then neither does the output.

> I assumed that the next step would just not have any deltas and so the expected value would not have changed.

Right, and for the incremental implementation this means outputting an empty batch at the second step.

ryzhyk (approved):

PR #141, opened 2022-08-01
[nexmark] Add initial macros for running nexmark queries.

absoludity:

Given each Nexmark query potentially has a different output type (and so each query has a different type), I couldn't see any way to iterate the various queries without using macros (and it was a good chance to learn more of Rust's macro support). Let me know if there's a better way.

This PR allows running the Nexmark tests (all by default, but you can select which ones) with whatever configuration you like, and prints out a table similar to the original [Nexmark benchmark results](https://github.com/nexmark/nexmark#benchmark-results).

I've left the defaults as they are for Nexmark itself, but you can run it with the actual config used for the above benchmark results, ie. 100,000,000 events at a rate of 1M per second, with:

```
$ cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=1000000 --max-events=100000000
...
Starting q0 bench of 100000000 events...
Starting q1 bench of 100000000 events...
Starting q2 bench of 100000000 events...
Starting q3 bench of 100000000 events...
Starting q4 bench of 100000000 events...
┌───────────────┬─────────────┬───────┬─────────┬─────────────────┬──────────────────┐
│ Nexmark Query │ Events Num  │ Cores │ Time(s) │ Cores * Time(s) │ Throughput/Cores │
├───────────────┼─────────────┼───────┼─────────┼─────────────────┼──────────────────┤
│ q0            │ 100,000,000 │ 1     │ 99.999  │ 99.999          │ 1000.010 K/s     │
│ q1            │ 100,000,000 │ 1     │ 100.005 │ 100.005         │ 999.949 K/s      │
│ q2            │ 100,000,000 │ 1     │ 100.012 │ 100.012         │ 999.881 K/s      │
│ q3            │ 100,000,000 │ 1     │ 100.000 │ 100.000         │ 1000.004 K/s     │
│ q4            │ 100,000,000 │ 1     │ 116.244 │ 116.244         │ 860.260 K/s      │
└───────────────┴─────────────┴───────┴─────────┴─────────────────┴──────────────────┘
```
Note that it is really only Nexmark q4 that starts to get behind the generated events (by 16%). Early days yet (to put too much value on the results here until I've gone through them some more), but it'll be fun to play with as we optimise queries or internals. You can run particular tests as follows (in this case, q0 and q3, but with only 1,000,000 events:

```
$ cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=1000000 --max-events=1000000 --query q0 --query q3
...
Starting q0 bench of 1000000 events...
Starting q3 bench of 1000000 events...
┌───────────────┬────────────┬───────┬─────────┬─────────────────┬──────────────────┐
│ Nexmark Query │ Events Num │ Cores │ Time(s) │ Cores * Time(s) │ Throughput/Cores │
├───────────────┼────────────┼───────┼─────────┼─────────────────┼──────────────────┤
│ q0            │ 1,000,000  │ 1     │ 1.006   │ 1.006           │ 993.680 K/s      │
│ q3            │ 1,000,000  │ 1     │ 1.000   │ 1.000           │ 1000.478 K/s     │
└───────────────┴────────────┴───────┴─────────┴─────────────────┴──────────────────┘
```

~~There are a couple of remaining issues, in particular, after discovering that I couldn't count the `zset.len()`s to accurately count the number of events processed (since duplicate bids will be merged with an accumulated weight - see inline comment), I've then incorrectly assumed that I can instead just stop iterating when the stream contains empty zsets. This worked for q0 and q1, but doesn't for the other queries (and queries generally) which emit empty zsets initial until (as can be sen above, and is obvious in retrospect :) ).~~ I ended up using channels rather than shared memory, which is (in my mind) much easier to reason about - even for a single-threaded binary.

### DIFF
--- a/benches/nexmark/main.rs
+++ b/benches/nexmark/main.rs
@@ -2,9 +2,13 @@
 //!
 //! CLI for running Nexmark benchmarks with DBSP.
 #![feature(is_some_with)]
-use std::{cell::Cell, rc::Rc, time::Instant};
+use std::{
+    sync::{mpsc, mpsc::TryRecvError},
+    time::{Duration, Instant},
+};
 
 use anyhow::Result;
+use ascii_table::AsciiTable;
 use clap::Parser;
 use dbsp::{
     circuit::{Circuit, Root},
@@ -17,6 +21,7 @@ use dbsp::{
     },
     trace::{ord::OrdZSet, BatchReader},
 };
+use num_format::{Locale, ToFormattedString};
 use rand::prelude::ThreadRng;
 
 // TODO: Ideally these macros would be in a separate `lib.rs` in this benchmark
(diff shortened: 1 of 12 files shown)


benches/nexmark/main.rs
```diff
@@ -2,64 +2,122 @@
...
+
+    run_queries!(
+        generator_config,
+        max_events,
+        queries_to_run,
+        (String::from("q0"), q0),
+        (String::from("q1"), q1),
+        (String::from("q2"), q2),
+        (String::from("q3"), q3),
+        (String::from("q4"), q4)
```
Kixiron, line 119: You should just be able to use literals here, no need for the from calls

benches/nexmark/main.rs
```diff
@@ -2,64 +2,122 @@
...
+
+        let circuit = nexmark_circuit!($q, $generator_config, max_events_reached);
+
+        let root = Root::build(circuit).unwrap();
+
+        let start = Instant::now();
+        loop {
+            if max_events_reached_cloned.get() {
+                break;
+            }
```
Kixiron, line 71: Can't this be a while loop?

absoludity: Yep. I'd initially thought that putting the condition outside the loop (ie. the while condition) would mean that it wouldn't go out of scope between iterations, but seems fine, thanks.

benches/nexmark/main.rs
absoludity: I wasn't able to... if I change these to literals, it works through to requiring the `query` field of the config to be `Vec<&str>` which then needs lifetimes... Or did you mean that the literals should be coerced into strings? Let me know if I've missed something.

absoludity (commented):
I'll try to sort out a consistent way to know that the full number of events has indeed been processed. Thanks for the input.

absoludity:
OK, PTAL - I've updated the description showing the output for the same config options as the original Nexmark test (as well as noting that I'm using channels rather than shared memory, even though it's a single thread).

benches/nexmark/main.rs
Kixiron: Ah, in that case just move the string creation into the macro by putting a `.to_string()` call wherever the strings are used

absoludity: Right, done.

PR #144, opened 2022-08-04
Additional implementations of MulByRef

mbudiu-vmw:


Kixiron:
The new trait isn't really needed since we already have `MulByRef`/`Mul`/`MulAssign`

mbudiu-vmw:
The types are different

Kixiron:
`MulByRef` has the exact same signature, except more general
```rust
pub trait MulByRef<Rhs = Self> {
    type Output;

    fn mul_by_ref(&self, other: &Rhs) -> Self;
}
```

mbudiu-vmw:
I guess now it does. I will try to do without it. Let's not merge this yet.

mbudiu-vmw:
Can I please get a review for this?
These new implementations are needed by the SQL compiler.

src/algebra/mod.rs
```diff
@@ -270,6 +270,43 @@ where
     }
 }
 
+/// Trait that allows values to be multiplied by weights.  Usually W
+/// will be a ZRingValue.  This is used for example by SQL's SUM
+/// aggregation, whose general implementation on Z-sets needs to
+/// multiply each value with its weight before summing.
+pub trait MulByWeight<W> {
```
ryzhyk: I thought we decided not to introduce this trait.

PR #148, opened 2022-08-09
[nexmark] Add Query 6 for nexmark

absoludity:

This query is windowed not by time, but rather by latest data, so should be achievable (unlike the event-time-windowed q5), and shares a lot with q4.

~~While testing my initial q6 (sans the limit of 10 latest) with multiple batches, I realised that for previous queries (q4) I've been `join`ing on the incremental batch data, which works fine when all the data is in a single batch, does not handle things like a bid in a second batch which is for an auction in the first batch. The test `test_q6_average_bids_per_seller_single_seller_single_auction` verifies that this works here for q6 after adding the `.integrate()` on the join operands. Let me know if this is incorrect (seems odd to be moving away from incremental calculations, but don't see how the join for a bid in a different batch to the auction can work otherwise). If it is correct, I'll go back to test and fix q4 for multiple batches.~~ (It was a bug which was already fixed in main).

The given q6 tests now pass as expected.

### DIFF
--- a/src/nexmark/generator/mod.rs
+++ b/src/nexmark/generator/mod.rs
@@ -88,7 +88,7 @@ impl<R: Rng> EventGenerator<R> for NexmarkGenerator<R> {
         if !self.has_next() {
             return Ok(None);
         }
-        if self.wallclock_base_time == None {
+        if self.wallclock_base_time.is_none() {
             self.wallclock_base_time = Some(self.wallclock_time());
         }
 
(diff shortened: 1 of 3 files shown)


ryzhyk:
Haven't looked at q6 yet, but I think the q4 implementation is correct. Join internally integrates both its input streams and should correctly compute changes to the result, matching bids to auctions even across different batches.

ryzhyk (approved):

absoludity:
Took a bit to realise I needed to install clippy from nightly to be able to reproduce the clippy error that was showing here in CI... but fixed now.

PR #149, opened 2022-08-09
Allow joins to use disjoint weights and optimize PR & BFS

Kixiron:
- Adds the `F64`, `F32` and `Present` difference types
- Allows joins to take in two collections with different difference types and produce a collection with yet another difference type
- Changes pagerank & bfs benchmarks to use `Present` weights for their inputs (big memory reduction and speedup)


ryzhyk (approved):
Looks good.  You have to seriously know what you're doing when choosing `Present` for weight.  A comment to that effect, along with a list of sufficient conditions for using `Present` in the `present` module would be nice.

PR #154, opened 2022-08-18
Float

mbudiu-vmw:


mbudiu-vmw:
Clippy was unhappy about the const annotations:
```
+ cargo clippy -- -D warnings
    Checking dbsp v0.1.0 (/home/mbudiu/git/database-stream-processor)
error: a `const` item should never be interior mutable
  --> src/circuit/runtime.rs:37:1
   |
37 | / thread_local! {
38 | |     // Reference to the `Runtime` that manages this worker thread or `None`
39 | |     // if the current thread is not running in a multithreaded runtime.
40 | |     static RUNTIME: RefCell<Option<Runtime>> = const { RefCell::new(None) };
...  |
45 | |     pub(crate) static WORKER_INDEX: Cell<usize> = const { Cell::new(0) };
46 | | }
   | |_^
   |
   = note: `-D clippy::declare-interior-mutable-const` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const
   = note: this error originates in the macro `$crate::__thread_local_inner` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Kixiron:
> Clippy was unhappy about the const annotations:

Yah, that's a false positive since all this does is create an initializing const

ryzhyk:
> Yah, that's a false positive since all this does is create an initializing const, just add #[allow(clippy::declare_interior_mutable_const)] to them

I guess this is ok because the const will be evaluated at most once, and I can't think of a sensible implementation that would do anything else, but I don't see any documentation that guarantees this behavior.

Kixiron:
Well in the general case this lint is fine, but in the case of `thread_local!` it just creates the const value before [setting a thread local static's value to said constant](https://doc.rust-lang.org/src/std/thread/local.rs.html#209) so the const isn't really "used" any and there's no way for bad things to happen

ryzhyk:
> Well in the general case this lint is fine, but in the case of thread_local! it just creates the const value before [setting a thread local static's value to said constant](https://doc.rust-lang.org/src/std/thread/local.rs.html#209) so the const isn't really "used" any and there's no way for bad things to happen

Yes, that's what I meant. But this behavior is not guaranteed by the `thread_local` API. And if it changes for whatever reasons (unlikely, but unlikely things have happened before), it will break our code.

Frankly this is not a big deal.  I'm ok merging this with or without @Kixiron 's suggested fix.

mbudiu-vmw:
I don't really understand what is suggested, but I need this to make progress, so I will merge it.

PR #157, opened 2022-08-22
Aggregation revamp.

ryzhyk:
See commit messages for details.

@mbudiu-vmw , this is still groupby-aggregate only.  I will add an aggregation operator that works on the entire Z-set shortly.

Some other TODOs:

- Proptests
- Optimized Max and Top-k aggregates.
- Argmin, Argmax and other useful library aggregates (@mbudiu-vmw , what do we need for SQL?)


src/operator/aggregate/fold.rs
```diff
@@ -0,0 +1,86 @@
...
+/// applying a user-provided `output` function to the final value of
+/// the accumulator.
+///
+/// # Type arguments
+///
+/// * `A` - accumulator
+/// * `SF` - step function
+/// * `OF` - output function
+#[derive(Clone)]
+pub struct Fold<A, SF, OF> {
```
mbudiu-vmw, line 19: Call this FoldOps?

src/operator/aggregate/fold.rs
```diff
@@ -0,0 +1,86 @@
...
+                (self.step)(&mut acc, cursor.key(), weight);
+            }
+
+            cursor.step_key();
+        }
+
+        if non_empty {
+            Some((self.output)(acc))
+        } else {
+            None
```
mbudiu-vmw, line 83: This needs to be documented.

src/operator/aggregate/fold.rs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        let mut acc = self.init.clone();
+        let mut non_empty = false;
+
+        while cursor.key_valid() {
+            let mut weight = R::zero();
+
+            cursor.map_times(|_t, w| {
+                weight.add_assign_by_ref(w);
+            });
```
mbudiu-vmw, line 71: This is asymmetric, it doesn't accumulate elements with zero weight at the start but it does afterwards?

src/operator/aggregate/fold.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub fn new(init: A, step: SF) -> Self {
+        Self {
+            init,
+            step,
+            output: identity,
+        }
+    }
+}
+
+impl<A, SF, OF> Fold<A, SF, OF> {
```
mbudiu-vmw, line 41: How about a parallel composition of fold which makes a fold on tuples out of n folds point wise? May need to be a macro.

src/operator/aggregate/mod.rs
```diff
@@ -0,0 +1,1058 @@
...
+/// aggregators.  Users will typicaly work with ready-made implementations
+/// like [`Min`] and [`Fold`].
+pub trait Aggregator<K, T, R> {
+    /// Aggregate type output by this aggregator.
+    type Output;
+
+    /// Compute an aggregate over a Z-set.
+    ///
+    /// Takes a cursor pointing to the first key of a Z-set and outputs
+    /// an aggregate of the Z-set.  This is a low-level API that
```
mbudiu-vmw, line 58: Duplicated below

src/operator/aggregate/mod.rs
```diff
@@ -0,0 +1,1058 @@
...
+struct WeightAggregator;
+
+impl<T, R> Aggregator<(), T, R> for WeightAggregator
+where
+    T: Timestamp,
+    R: MonoidValue,
+{
+    type Output = R;
+
+    fn aggregate<'s, C>(&mut self, cursor: &mut C) -> Option<Self::Output>
```
mbudiu-vmw, line 83: This is more like weighted count.

src/operator/aggregate/mod.rs
```diff
@@ -0,0 +1,1058 @@
...
+    ///
+    /// ```text
+    ///    __
+    ///    ╲
+    ///    ╱ f(k,v) * w
+    ///    ‾‾
+    /// (k,v,w) ∈ Z
+    /// ```
+    ///
+    /// This is a linear operator.
```
mbudiu-vmw, line 283: Should this be a property?

src/operator/aggregate/mod.rs
```diff
@@ -0,0 +1,1058 @@
...
+            time: IT::Time::clock_start(),
+            empty_input: false,
+            empty_output: false,
+            keys_of_interest: BTreeMap::new(),
+            output_delta: Vec::new(),
+            _type: PhantomData,
+        }
+    }
+
+    /// Compute output of the operator for `key`.
```
mbudiu-vmw, line 456: Can't say I got all this. Looking forward to the paper.

mbudiu-vmw (approved):
SQL logic test doesn't use other aggregates. Some dialects of SQL may have more. Aggregate distinct will în fact have to call distinct and then a normal aggregate.

src/operator/aggregate/fold.rs
ryzhyk: It should always skip zero-weight elements in the cursor. It handles all keys in the cursor in the same way.

src/operator/aggregate/mod.rs
ryzhyk: What do you mean by this?

src/operator/aggregate/fold.rs
mbudiu-vmw: Doesn't seem to do that. E.g. If the weight becomes zero in the middle.

src/operator/aggregate/fold.rs
ryzhyk: Yep, that make sense. I'll create an issue (need to work on something else now).

src/operator/aggregate/fold.rs
ryzhyk: It will continue iterating over other keys in the same way.

src/operator/aggregate/fold.rs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        let mut acc = self.init.clone();
+        let mut non_empty = false;
+
+        while cursor.key_valid() {
+            let mut weight = R::zero();
+
+            cursor.map_times(|_t, w| {
+                weight.add_assign_by_ref(w);
+            });
```
Kixiron, line 71:
```suggestion
            cursor.map_times(|_t, w| weight.add_assign_by_ref(w));
```

src/operator/aggregate/fold.rs
```diff
@@ -0,0 +1,88 @@
...
+            cursor.step_key();
+        }
+
+        // Aggregator must return None iff the input cursor is empty (all keys have
+        // weight 0).
+        if non_empty {
+            Some((self.output)(acc))
+        } else {
+            None
+        }
```
Kixiron, line 86:
```suggestion
        non_empty.then(|| (self.output)(acc))
```

src/operator/aggregate/max.rs
```diff
@@ -0,0 +1,38 @@
...
+        C: Cursor<'s, V, (), T, R>,
+    {
+        let mut result = None;
+
+        while cursor.key_valid() {
+            let mut weight = R::zero();
+
+            cursor.map_times(|_t, w| {
+                weight.add_assign_by_ref(w);
+            });
```
Kixiron, line 27:
```suggestion
            cursor.map_times(|_t, w| weight.add_assign_by_ref(w));
```

src/operator/aggregate/min.rs
```diff
@@ -0,0 +1,37 @@
...
+    fn aggregate<'s, C>(&mut self, cursor: &mut C) -> Option<Self::Output>
+    where
+        C: Cursor<'s, V, (), T, R>,
+    {
+        while cursor.key_valid() {
+            let mut weight = R::zero();
+
+            cursor.map_times(|_t, w| {
+                weight.add_assign_by_ref(w);
+            });
```
Kixiron, line 27:
```suggestion
            cursor.map_times(|_t, w| weight.add_assign_by_ref(w));
```

src/operator/aggregate/mod.rs
```diff
@@ -0,0 +1,1057 @@
...
+            DelayedTraceId::new(output_trace.origin_node_id().clone()),
+            output_trace_delayed,
+        );
+        circuit.cache_insert(
+            ExportId::new(output_trace.origin_node_id().clone()),
+            output_trace_export,
+        );
+        circuit.cache_insert(TraceId::new(output.origin_node_id().clone()), output_trace);
+
+        output
```
Kixiron, line 231: Shouldn't this all be within a cache check?

src/operator/aggregate/mod.rs
```diff
@@ -0,0 +1,1057 @@
...
+    A: Aggregator<Z::Val, (), Z::R> + 'static,
+    O: Clone + IndexedZSet<Key = Z::Key, Val = A::Output> + 'static,
+    O::R: ZRingValue,
+{
+    fn eval(&mut self, i: &Z) -> O {
+        let mut builder = O::Builder::with_capacity((), i.len());
+
+        let mut cursor = i.cursor();
+
+        while cursor.key_valid() {
```
Kixiron, line 363:
```suggestion
        let mut cursor = i.cursor();
        while cursor.key_valid() {
```

src/operator/aggregate/mod.rs
```diff
@@ -0,0 +1,1057 @@
...
+            if let Some(agg) = self
+                .aggregator
+                .aggregate(&mut CursorGroup::new(&mut cursor, ()))
+            {
+                builder.push((O::item_from(cursor.key().clone(), agg), O::R::one()));
```
Kixiron, line 368: Kinda don't like that we're forced to clone the key value for this

src/operator/aggregate/mod.rs
```diff
@@ -0,0 +1,1057 @@
...
+                    //     output_cursor.val(),
+                    //     weight
+                    // );
+                    self.output_delta
+                        .push((output_cursor.val().clone(), weight.neg()));
+                }
+                output_cursor.step_val();
+            }
+        }
+        consolidate(&mut self.output_delta);
```
Kixiron, line 619:
```suggestion
        }

        consolidate(&mut self.output_delta);
```

src/trace/cursor/cursor_group.rs
```diff
@@ -0,0 +1,104 @@
+use crate::{trace::cursor::Cursor, Timestamp};
+use std::marker::PhantomData;
+
+/// A `CursorGroup` iterates over values associated with a single key of a base
+/// cursor of type `C: Cursor<K, V, R, T>`.
+pub struct CursorGroup<'c, K, T, C> {
```
Kixiron, line 6: Lots of missing tests here

Kixiron (changes requested):

src/operator/aggregate/mod.rs
ryzhyk: Hmm, I really don't see how to avoid this.

Kixiron: Yah, it was more of a general gripe

lalithsuresh:
Not sure which aggregates @mbudiu-vmw is implementing for SQL, but the ones that came up with ddlog-dcm were:
* Top-K (ORDER BY + LIMIT)
* min/max
* sum/count
* all/any
* array_agg, set_agg (we used array_agg because we didn't have set_agg, but set_agg would have an efficient incremental implementation unlike array_agg)
* window operators (SQL partition - over)

src/trace/cursor/cursor_group.rs
ryzhyk: @gz kindly promised to share his proptest-based test harness for cursors, which should improve the situation.

PR #160, opened 2022-08-24
Stopped shifting elements during join

Kixiron:
Previously we used `.drain()` on a vec to grab elements while joining, but this change switches us to use a `Vec<(T, MaybeUninit<U>)>` where we successively "claim" the `MaybeUninit` elements from the vector, meaning that we can skip out on a lot of unnecessary memcpy calls. Seems to shave off a not-insignificant amount of time from some benchmarks, around 2 seconds from bfs over `datagen-8_4-fb`

### DIFF
--- a/src/operator/join.rs
+++ b/src/operator/join.rs
@@ -21,6 +21,7 @@ use std::{
     fmt::Write,
     hash::Hash,
     marker::PhantomData,
+    mem::{needs_drop, MaybeUninit},
 };
 use timely::PartialOrder;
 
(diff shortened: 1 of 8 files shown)


src/operator/join.rs
```diff
@@ -570,12 +576,56 @@ where
         // timestamp to the appropriate batcher.
         output_tuples.sort_by(|(t1, _), (t2, _)| t1.cmp(t2));
 
-        let mut batch = Vec::new();
-        while !output_tuples.is_empty() {
-            let batch_time = output_tuples[0].0.clone();
+        /// Ensures that we don't leak any values if we panic while creating
```
ryzhyk, line 579: wouldn't it be enough to tweak the original logic to `drain` the vector from the tail? I think this way no memcpy would be happening and you wouldn't need all this complexity.

Kixiron: I don't think partitioning would work the same way, the binary search it provides is a big deal perf-wise

ryzhyk: Right, it's silly that there is no version of `partition_point` and `binary_search_by` that work in reverse.

PR #161, opened 2022-08-24
Pre-shard ldbc inputs and propagate stream sharding where possible

Kixiron:
This PR shards graphalytics inputs at load time and provides the facilities for operators to propagate sharding information.  
For example, the `.semijoin()` operator shards both of its inputs before performing a join on their keys, but then it outputs the result stream without mutating its keys any, meaning that the output streams are already sharded. With the `.mark_sharded()` and `.is_sharded()` methods we can check if an input stream has been sharded and then conditionally mark the produced output as also being sharded, allowing us to potentially remove a lot of sync points from within our dataflow graph

### DIFF
--- a/benches/ldbc-graphalytics/main.rs
+++ b/benches/ldbc-graphalytics/main.rs
@@ -8,13 +8,12 @@ use crate::data::{
 };
 use clap::Parser;
 use dbsp::{
-    algebra::{NegByRef, Present},
     circuit::{trace::SchedulerEvent, Runtime},
     monitor::TraceMonitor,
     operator::Generator,
     profile::CPUProfiler,
     trace::{BatchReader, Cursor},
-    zset, Circuit,
+    Circuit,
 };
 use deepsize::DeepSizeOf;
 use hashbrown::HashMap;
@@ -29,10 +28,11 @@ use std::{
     cell::RefCell,
     fmt::Write as _,
     io::{self, Write},
-    mem::size_of,
+    mem::{replace, size_of, take},
     num::{NonZeroU8, NonZeroUsize},
-    ops::Add,
+    ops::{Add, Neg},
     rc::Rc,
+    sync::{Arc, Mutex},
     thread,
     time::{Duration, Instant},
 };
@@ -83,15 +83,22 @@ fn main() {
     io::stdout().flush().unwrap();
     let start = Instant::now();
 
-    let (properties, edges, vertices, _) = dataset.load::<NoopResults>().unwrap();
+    let (properties, edge_data, vertex_data, _) =
+        dataset.load::<NoopResults>(threads.get()).unwrap();
     let elapsed = start.elapsed();
 
     // Calculate the amount of data that we've loaded in, including weights
-    let actual_size = edges.deep_size_of() + vertices.deep_size_of();
+    let actual_size = edge_data.deep_size_of() + vertex_data.deep_size_of();
     let theoretical_size = {
         let node = size_of::<Node>() as u64;
-        let edges = edges.len() as u64;
-        let vertices = vertices.len() as u64;
+        let edges = edge_data
+            .iter()
+            .map(|edges| edges.len() as u64)
+            .sum::<u64>();
+        let vertices = vertex_data
+            .iter()
+            .map(|vertices| vertices.len() as u64)
+            .sum::<u64>();
 
         let vertex_bytes = vertices * node;
         let edge_bytes = edges * node * 2;
@@ -114,12 +121,14 @@ fn main() {
         properties.edges,
     );
 
-    let mut root_data = Some(zset! { properties.source_vertex => Present });
-    let mut vertex_data = Some(vertices);
-    let mut edge_data = Some(edges);
+    let edge_data = Arc::new(edge_data.into_iter().map(Mutex::new).collect::<Vec<_>>());
+    let vertex_data = Arc::new(vertex_data.into_iter().map(Mutex::new).collect::<Vec<_>>());
 
     Runtime::run(threads.get(), move || {
-        let init_stats = if Runtime::worker_index() == 0 {
+        let worker_idx = Runtime::worker_index();
+        let is_leader = worker_idx == 0;
+
+        let init_stats = if is_leader {
             print!(
                 "building dataflow{}... ",
                 if config.profile { " with profiling" } else { "" },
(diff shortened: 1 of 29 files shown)


benches/ldbc-graphalytics/main.rs
```diff
@@ -136,38 +147,34 @@ fn main() {
...
         let output_inner = output.clone();
         let root = Circuit::build(move |circuit| {
-            if config.profile && Runtime::worker_index() == 0 {
+            if config.profile && is_leader {
                 attach_profiling(dataset, circuit);
             }
 
             let roots = circuit.region("roots", || {
-                circuit.add_source(Generator::new(move || if Runtime::worker_index() == 0 {
+                circuit.add_source(Generator::new(move || if is_leader {
```
ryzhyk, line 155: This can be much cleaner with the new input handle API.  You can use `add_input` to create a "raw" `InputHandle`, which can be used to push pre-batched data to workers.

src/operator/communication/shard.rs
```diff
@@ -230,6 +230,47 @@ where
...
+
+    /// Returns `true` if the current stream is properly sharded across workers
+    pub(crate) fn is_sharded(&self) -> bool {
+        Runtime::runtime()
+            .map(|runtime| {
+                if runtime.num_workers() == 1 {
+                    true
+                } else {
+                    self.circuit()
+                        .cache_contains(&ShardId::<Circuit<P>, T>::new((
```
ryzhyk, line 264: This can return `true` is `self` is not sharded, but there is another stream created by `self.shard()` that carries the sharded version of `self`.

ryzhyk (changes requested):

ryzhyk:
This also applies to `filter`. And to the new aggregation operators (about to get merged).

src/operator/communication/shard.rs
Kixiron: I think I fixed this with `DirectShardId` but I'm not 100% sure

src/operator/filter_map.rs
```diff
@@ -236,8 +243,15 @@ where
     where
         F: Fn(Self::ItemRef<'_>) -> bool + 'static,
     {
-        self.circuit()
-            .add_unary_operator(FilterVals::new(filter_func), self)
+        let filtered = self
+            .circuit()
+            .add_unary_operator(FilterVals::new(filter_func), self);
+
+        if filtered.is_sharded() {
```
ryzhyk, line 250: `self.is_sharded`?

src/operator/filter_map.rs
```diff
@@ -168,8 +168,15 @@ where
     where
         F: Fn(Self::ItemRef<'_>) -> bool + 'static,
     {
-        self.circuit()
-            .add_unary_operator(FilterKeys::new(filter_func), self)
+        let filtered = self
+            .circuit()
+            .add_unary_operator(FilterKeys::new(filter_func), self);
+
+        if filtered.is_sharded() {
```
ryzhyk, line 175: `self.is_sharded()`?

src/operator/communication/shard.rs
ryzhyk: This seems too heavy-handed.  We shouldn't need another cache id for this. How about this solution:
- `shard_generic` records `Cache[ShardId(sharded)] := sharded` in addition to `Cached[ShardId(self)] := sharded`.
- `mark_sharded` records `Cache[ShardId(self)] := self`.
- All operators that preserve "shardedness" of inputs:
  - if `let Some(sharded) = Cache[ShardId(stream)]` then use `sharded` instead of `self` as input and mark output as sharded.
- remove `is_sharded` method, as it answers the wrong question: we don't normally care if the stream is sharded, but rather whether there exists a sharded version of the stream.

Kixiron: I considered that but I'm not sure I like it to be honest, it seems like just a little bit too much magic to shard operators that weren't previously sharded and don't technically need sharding

ryzhyk: We certainly don't want to do that, but that's not what I'm suggesting.  The operator should simply check if there already _exists_ a cached sharded version of the stream and use that if possible.  Otherwise it should just work with non-sharded inputs and produce non-sharded outputs.

src/operator/communication/shard.rs
```diff
@@ -230,6 +242,46 @@ where
...
+{
+    /// Marks the data within the current stream as sharded, meaning that all
+    /// further calls to `.shard()` will have no effect.
+    ///
+    /// This must only be used on streams of values that are properly sharded
+    /// across workers, otherwise this will cause the dataflow to yield
+    /// incorrect results
+    pub fn mark_sharded(&self) -> Self {
+        self.circuit().cache_insert(
+            DirectShardId::new((
```
ryzhyk, line 258: Why can't this be just `ShardId`?

src/operator/communication/shard.rs
```diff
@@ -230,6 +242,46 @@ where
...
+                sharding_policy(self.circuit()),
+            )),
+            self.clone(),
+        );
+        self.clone()
+    }
+
+    pub fn has_sharded_version(&self) -> bool {
+        self.circuit()
+            .cache_contains(&DirectShardId::<Circuit<P>, T>::new((
```
ryzhyk, line 269: `ShardId`

src/operator/communication/shard.rs
```diff
@@ -230,6 +242,46 @@ where
...
+                self.origin_node_id().clone(),
+                sharding_policy(self.circuit()),
+            )))
+    }
+
+    pub fn try_sharded_version(&self) -> Self {
+        self.circuit()
+            .cache_get(&DirectShardId::new((
```
ryzhyk, line 277: `ShardId`

src/operator/communication/shard.rs
```diff
@@ -13,6 +13,7 @@ use crate::{
 use std::hash::Hash;
 
 circuit_cache_key!(ShardId<C, D>((GlobalNodeId, ShardingPolicy) => Stream<C, D>));
+circuit_cache_key!(DirectShardId<C, D>((GlobalNodeId, ShardingPolicy) => Stream<C, D>));
```
ryzhyk, line 16: I don't think this is needed anymore.

Kixiron: Sorry, forgot to remove it

ryzhyk (approved):

PR #164, opened 2022-08-26
[nexmark] Support `--num-event-generators` for multithreaded source.

absoludity:
Note that adding support for multi-threaded source means that the output stats can only contain total cpu time now :/ (until the source generation is moved to a separate process).

Running with 2 source generators is an improvement, but not a huge one. As an example, here's running q4 with 4 CPUs in DBSP but only one source generator:

```
$ cargo flamegraph -c "record --call-graph lbr -g" --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=10000000 --cpu-cores 2 --query q4 --num-event-generators 1
...
│ q4    │ 10,000,001 │ 2     │ 6.608s  │ 13.216s         │ 756.642 K/s      │ 20.087s       │ 3.784s        │ 770,648     │
```
versus with 2 source generators:

```
$ cargo flamegraph -c "record --call-graph lbr -g" --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=10000000 --cpu-cores 2 --query q4 --num-event-generators 2
...
│ q4    │ 10,000,001 │ 2     │ 5.221s  │ 10.442s         │ 957.694 K/s      │ 19.914s       │ 1.884s        │ 756,112     │
```

So an improvement of over 1 second for 10M events.

I'll mark this PR as ready for review/landing, and will try some further improvements in a follow-up PRs. Specifically,
- Now that the source generator(s) have backpressure (size of the channel), I don't think there's a need to batch in the source generator itself, but rather just control how much is inputted into dbsp via the input handle as a batch for each step.
- I think it is possible to load an initial input into DBSP of a configured size, then call step and concurrently start inputting the next input (as is happening now), but signalling when step completes so the input thread can itself control whether to increase or decrease the input size.
- start looking more closely at the cpu usage during next_event.


benches/nexmark/main.rs
```diff
@@ -93,14 +87,15 @@ fn spawn_dbsp_consumer(
...
-    generator_config: GeneratorConfig,
+    nexmark_config: NexmarkConfig,
     mut input_handle: CollectionHandle<Event, isize>,
     step_do_rx: mpsc::Receiver<usize>,
     step_done_tx: mpsc::SyncSender<StepCompleted>,
     source_exhausted_tx: mpsc::SyncSender<InputStats>,
 ) {
     thread::spawn(move || {
-        let source = NexmarkSource::<isize, OrdZSet<Event, isize>>::new(generator_config);
+        println!("max events: {}", nexmark_config.max_events);
```
Kixiron, line 97: Is this a leftover debug print?

Kixiron (approved):

benches/nexmark/main.rs
absoludity: Yes, removed - thanks.

PR #166, opened 2022-08-29
Code cleanup

Kixiron:
* Added source locations to exchange operators
* Renamed renamed `Merger::new()` to `Merger::new_merger()`, `Batcher::new()` to `Batcher::new_batcher()` and `Builder::new()` to `Builder::new_builder()` to decrease method ambiguity
* Made a few `Batch::empty()` implementations simpler, they previously used `Self::Builder::new_builder(time).done()` which could conceivably allocate buffers but they now do exactly nothing (exactly what we want 'em to do)
* Removed a bunch of `<O as TryFrom<usize>>::Error: Debug` bounds that were infecting everything, usize conversion is now part of the `OrdOffset` trait


ryzhyk (approved):
Looks good!

PR #168, opened 2022-08-31
[nexmark] Reduce Send/Recv with batched receivers

absoludity:
This PR abstracts the concept of a `BatchedReceiver` - a receiver of `VecDecue<T>`, and uses it for the generator to main collector communication, further reducing the CPU required for the generator. I've also then pulled out the batched sending functionality into a `BatchedSender` and added a `batched_channel()` helper to create the sender/receiver. The only difference in usage is that you need to call `.flush()` on the sender before finishing (in case any batched results remain unsent).

For example, q4 with 10M events, 4 CPUs and 2 generators (see the create_generators_for_config closure).
Before:
![before](https://user-images.githubusercontent.com/497518/187578184-eac3ce30-08b9-47a5-9b6c-5422d3b32405.png)

After:
![after](https://user-images.githubusercontent.com/497518/187578208-4c5759bc-a76e-440f-8535-af9e17d50e8e.png)

Timing-wise, easiest to see with the q0 results with different generators:
```
cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=10000000 --cpu-cores 4 --query q0 --num-event-generators 1 --source-buffer-size 10000 --input-batch-size 10000
     Running benches/nexmark/main.rs (target/release/deps/nexmark-c782d17920c2b5ef)
Starting q0 bench of 10000000 events...
Done                                                                                                                                                                                                                           ┌───────┬────────────┬───────┬─────────┬─────────────────┬──────────────────┬───────────────┬───────────────┬─────────────┐
│ q0    │ 10,000,001 │ 4     │ 4.199s  │ 16.795s         │ 595.426 K/s      │ 12.113s       │ 0.146s        │ 0           │

cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=10000000 --cpu-cores 4 --query q0 --num-event-generators 2 --source-buffer-size 10000 --input-batch-size 10000
     Running benches/nexmark/main.rs (target/release/deps/nexmark-c782d17920c2b5ef)
Starting q0 bench of 10000000 events...
9990000 / 10000000 [=================================================================================================================================================================================] 99.90 % 4186248.08/s 0s thread 'nexmark collector' panicked at 'called `Option::unwrap()` on a `None` value', /home/michael/dev/vmware/database-stream-processor/src/nexmark/mod.rs:58:20
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Done                                                                                                                                                                                                                           ┌───────┬────────────┬───────┬─────────┬─────────────────┬──────────────────┬───────────────┬───────────────┬─────────────┐
│ q0    │ 10,000,000 │ 4     │ 2.389s  │ 9.558s          │ 1046.281 K/s     │ 11.715s       │ 0.170s        │ 20,856      │

cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=10000000 --cpu-cores 4 --query q0 --num-event-generators 4 --source-buffer-size 10000 --input-batch-size 10000
    Finished bench [optimized + debuginfo] target(s) in 0.12s
     Running benches/nexmark/main.rs (target/release/deps/nexmark-c782d17920c2b5ef)
Starting q0 bench of 10000000 events...
9980000 / 10000000 [=================================================================================================================================================================================] 99.80 % 5325678.40/s 0s thread 'nexmark collector' panicked at 'called `Option::unwrap()` on a `None` value', /home/michael/dev/vmware/database-stream-processor/src/nexmark/mod.rs:58:20
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Done                                                                                                                                                                                                                           ┌───────┬────────────┬───────┬─────────┬─────────────────┬──────────────────┬───────────────┬───────────────┬─────────────┐
│ q0    │ 10,000,000 │ 4     │ 1.882s  │ 7.526s          │ 1328.713 K/s     │ 12.525s       │ 0.204s        │ 53,692      │
```

~~Note: I need to fix one of the sync options (see above panic when it completes the input), but otherwise, think this is ready for review (but no rush - can wait until tomorrow).~~

### DIFF
--- a/src/nexmark/mod.rs
+++ b/src/nexmark/mod.rs
@@ -36,6 +36,7 @@ pub mod generator;
 pub mod model;
 pub mod queries;
 
+/// BatchedReceiver abstracts the Receiver interface for channels of VecDeque's.
 pub struct BatchedReceiver<T> {
     next_queue: mpsc::Receiver<VecDeque<T>>,
     current_queue: VecDeque<T>,
@@ -92,7 +93,7 @@ fn create_generators_for_config<R: Rng + Default>(
         .unwrap()
         .as_millis() as u64;
     let buffer_size = nexmark_config.source_buffer_size;
-    let next_event_rxs: Vec<mpsc::Receiver<Option<NextEvent>>> = (0..nexmark_config
+    let mut next_event_rxs: Vec<BatchedReceiver<NextEvent>> = (0..nexmark_config
         .num_event_generators)
         .map(|generator_num| {
             GeneratorConfig::new(
@@ -103,34 +104,40 @@ fn create_generators_for_config<R: Rng + Default>(
             )
         })
         .map(|generator_config| {
-            let (tx, rx) = mpsc::sync_channel(generator_config.nexmark_config.source_buffer_size);
+            let (tx, rx) = mpsc::sync_channel(3);
             thread::Builder::new()
                 .name(format!("generator-{}", generator_config.first_event_number))
                 .spawn(move || {
+                    let capacity = generator_config.nexmark_config.source_buffer_size;
                     let mut generator =
                         NexmarkGenerator::new(generator_config, R::default(), wallclock_base_time);
+                    let mut v = VecDeque::with_capacity(capacity);
                     while let Ok(Some(event)) = generator.next_event() {
-                        tx.send(Some(event)).unwrap();
+                        v.push_back(event);
+                        if v.len() == capacity {
+                            tx.send(v).unwrap();
+                            v = VecDeque::with_capacity(capacity);
+                        }
                     }
+                    tx.send(v).unwrap();
                 })
                 .unwrap();
-            rx
+            BatchedReceiver::new(rx)
         })
         .collect();
 
     // Finally, read from the generators round-robin, sending the ordered
     // events down a single channel buffered.
-    let (next_events_tx, next_events_rx) = mpsc::sync_channel(buffer_size);
-    let next_events_rx = BatchedReceiver::new(next_events_rx);
+    let (next_events_tx, next_events_rx) = mpsc::sync_channel(3);
     thread::Builder::new()
         .name("nexmark collector".into())
         .spawn(move || {
             let mut num_completed_receivers = 0;
             let mut events = VecDeque::<NextEvent>::with_capacity(buffer_size);
             while num_completed_receivers < next_event_rxs.len() {
-                for rx in &next_event_rxs {
+                for rx in &mut next_event_rxs {
                     match rx.recv() {
-                        Ok(Some(e)) => {
+                        Ok(e) => {
                             events.push_back(e);
                             if events.len() == buffer_size {
                                 next_events_tx.send(events).unwrap();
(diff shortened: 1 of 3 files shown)


src/nexmark/mod.rs
```diff
@@ -36,6 +36,72 @@ pub mod generator;
...
+        BatchedReceiver::new(rx),
+    )
+}
+
+impl<T> BatchedSender<T> {
+    pub fn send(&mut self, t: T) -> Result<(), mpsc::SendError<VecDeque<T>>> {
+        // The order here ensures that we never end up sending an empty
+        // vector.
+        if self.current_queue.len() == self.batch_size {
+            self.tx.send(std::mem::replace(
```
absoludity, line 88: I'd not seen the use of `std::mem::replace` like this to be able to get ownership of a field in an `&mut self` to use, while setting the field to a new value... neat. Lucky I found [a relevant post](https://smallcultfollowing.com/babysteps/blog/2018/11/10/after-nll-moving-from-borrowed-data-and-the-sentinel-pattern/) or I would have struggled for a while.

src/nexmark/mod.rs
```diff
@@ -36,6 +36,72 @@ pub mod generator;
 pub mod model;
 pub mod queries;
 
+/// BatchedReceiver abstracts the Receiver interface for channels of VecDeque's.
+pub struct BatchedReceiver<T> {
```
Kixiron, line 40: Something I'm not entirely clear on: Does the order of events matter? If generator 1 creates an event, does it need to be in a specific sequence relative to the events produced  by other generators?

absoludity: > Does the order of events matter? 

Yes,

> If generator 1 creates an event, does it need to be in a specific sequence relative to the events produced by other generators?

Not necessarily - the order of the events produced by the `NexmarkSource` is deterministic  ([example tests](https://github.com/vmware/database-stream-processor/blob/38378b25b92dd7b94daa8898f38b290c224d0194/src/nexmark/mod.rs#L298-L352)), even when explicitly configuring to ensure events are "out of order". How the generation of those events are parallelised so that the `NexmarkSource` can collect and emit them in the correct order is up to us, and is the only factor in the ordering of events from a generator thread relative to the other generator threads.

So without batching, the simplest way was to ensure that the generator threads produce interleaved events, ie. the first of three generators produces events [0, 2, 5, ...], while the second [1, 3, 6, ...] etc. as [demo'd in this test](https://github.com/vmware/database-stream-processor/blob/38378b25b92dd7b94daa8898f38b290c224d0194/src/nexmark/generator/mod.rs#L223-L250). Then the `NexmarkSource` collection thread can simply iterate through the receiving queues to get the events in the correct order (see the test link in the previous paragraph for demo).

With this PR the events are batched when sent from the generator threads to the collecting `NexmarkSource` thread, so they wouldn't need to be interleaved necessarily, but it's still a neat, trivial way to ensure we can pull them off in the correct order, and it appears to be fast (since they're now `pop_front`'d from a vec, rather than received from the channel). So I haven't (yet) seen a need to change the ordering of the events in the generator threads themselves.

Does that make sense? Let me know if I've missed something.

PR #169, opened 2022-08-31
Implemented an owned Consumer api for OrdZSet

Kixiron:
Found a fairly basic consumer api that seems to work, only implemented it for `OrdZSet` for now


src/trace/cursor/mod.rs
```diff
@@ -113,6 +113,28 @@ pub trait Cursor<'s, K, V, T, R> {
...
+
+    fn next_key(&mut self) -> (K, Self::ValueConsumer<'_>);
+
+    fn seek_key(&mut self, key: &K)
+    where
+        K: Ord;
+}
+
+pub trait ValueConsumer<'a, V, R> {
+    fn value_valid(&self) -> bool;
```
ryzhyk, line 131: Let's call this `val_valid` for consistency with `Cursor`.

src/trace/cursor/mod.rs
```diff
@@ -113,6 +113,28 @@ pub trait Cursor<'s, K, V, T, R> {
     fn rewind_vals(&mut self);
 }
 
+pub trait Consumer<K, V, R> {
```
ryzhyk, line 116: Can you please document the new traits?

ryzhyk (approved):

src/trace/cursor/mod.rs
Kixiron: I personally dislike the shortened names to be honest, that's why I changed it

Kixiron:
Extended this to work with `OrdIndexedZSet` and `OrderedLayer<K, OrderedColumnLeaf<V, R>, O>`

Kixiron:
Implemented consumers for `OrdKeyBatch`

src/operator/aggregate/average.rs
```diff
@@ -153,11 +154,58 @@ where
...
             self.aggregate_linear::<TS, _, _>(move |key, val| Avg::new(f(key, val), 1isize));
 
-        // TODO: We can probably use some sort of `.map_index_owned()` here since in all
-        // likelihood we'll be the only consumer of the aggregated value
-        // (meaning we wouldn't need to clone the average's sum and would be
-        // able to elide one clone of the key value)
-        let average = aggregate.map_index(|(k, avg)| (k.clone(), (avg.sum.clone()) / avg.count));
+        // We're the only possible consumer of the aggregated stream so we can use an
+        // owned consumer to skip any cloning
+        let average = aggregate.apply_owned_named("MapAverage", |aggregate| {
```
ryzhyk, line 159: Any chance you could add a test (proptest?) for the `average` operator?

ryzhyk: Also, how does this affect the performance of q4, which is the only benchmark that uses `average` at the moment?

Kixiron: On my machine with 1 thread & 1 generator I get 89.412s for the previous impl and 74.469s with my change

PR #173, opened 2022-09-02
[nexmark] Add query 14 showing custom types, functions and user-defined ranges.

absoludity:
This query uses a Decimal (3 decimal places) for price with a calculation, a custom type for whether the bid happens during the day, night (or other?!), a custom function for counting the 'c's in the extra data and a normal filter for bids in a certain range, as per the original java implementation.

Note, for now (until we can discuss them) I've skipped the following queries:
- [q10](https://github.com/nexmark/nexmark/blob/master/nexmark-flink/src/main/resources/queries/q10.sql) - tests logging to rolling files using flink-specific syntax (I think?). I can implement this using standard tools if we want to? (ie, csv output using serde).
- [q11](https://github.com/nexmark/nexmark/blob/master/nexmark-flink/src/main/resources/queries/q11.sql) - uses session windows... does dbsp have such a concept?
- [q12](https://github.com/nexmark/nexmark/blob/master/nexmark-flink/src/main/resources/queries/q12.sql) - windows in processing time... does dbsp have such a concept yet?
- [q13](https://github.com/nexmark/nexmark/blob/master/nexmark-flink/src/main/resources/queries/q13.sql) - uses side-loaded input from the filesystem to combine with the stream. This is another one that I could do manually, but I don't think we'd have something specific to DBSP?

Let me know what you think about the above 4 queries and I'll start those Monday (or continue past q14).

Thanks!

### DIFF
--- a/src/nexmark/queries/q14.rs
+++ b/src/nexmark/queries/q14.rs
@@ -1,0 +1,131 @@
+use super::NexmarkStream;
+use crate::{nexmark::model::Event, operator::FilterMap, Circuit, OrdZSet, Stream};
+use chrono::{TimeZone, Timelike, Utc};
+use rust_decimal::Decimal;
+
+/// Query 14: Calculation (Not in original suite)
+///
+/// Convert bid timestamp into types and find bids with specific price.
+/// Illustrates duplicate expressions and usage of user-defined-functions.
+///
+/// ```sql
+/// CREATE FUNCTION count_char AS 'com.github.nexmark.flink.udf.CountChar';
+///
+/// CREATE TABLE discard_sink (
+///     auction BIGINT,
+///     bidder BIGINT,
+///     price  DECIMAL(23, 3),
+///     bidTimeType VARCHAR,
+///     dateTime TIMESTAMP(3),
+///     extra VARCHAR,
+///     c_counts BIGINT
+/// ) WITH (
+///   'connector' = 'blackhole'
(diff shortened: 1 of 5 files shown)


src/nexmark/queries/q14.rs
```diff
@@ -0,0 +1,139 @@
...
+    let millis_for_day = millis % (24 * 60 * 60 * 1000);
+    (millis_for_day / (60 * 60 * 1000)) as usize
+}
+
+pub fn q14(input: NexmarkStream) -> Q14Stream {
+    input.flat_map(|event| match event {
+        Event::Bid(b) => {
+            let new_price = Decimal::new((b.price * 100) as i64, 2) * Decimal::new(908, 3);
+            match usize::try_from(new_price).unwrap() {
+                1_000_000..=50_000_000 => Some(Q14Output(
```
ryzhyk, line 67: The SQL query has strict inequalities (`<`, `>`)

ryzhyk (approved):

src/nexmark/queries/q14.rs
absoludity: Yeah, it does, but our `price` field on the model itself is an integer (as in the original), so I can't represent or test that boundary condition (1_000_000.0 / 0.908 = 1_101_321.6). But right, that shouldn't stop me from representing it correctly.

PR #176, opened 2022-09-06
[nexmark] Add q13 with bounded side-input.

absoludity:

```
cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=10000000 --cpu-cores 4 --query q13 --num-event-generators 4 --source-buffer-size 10000 --input-batch-size 10000
     Running benches/nexmark/main.rs (target/release/deps/nexmark-c95c15947b20642f)
Starting q13 bench of 10000000 events...
Done                                                                                                                                                                                                                
│ q13   │ 10,000,000 │ 4     │ 4.136s  │ 16.545s         │ 604.420 K/s      │ 26.442s       │ 0.687s        │ 91,628      │
```


src/nexmark/queries/q13.rs
```diff
@@ -0,0 +1,162 @@
...
+/// is not that used above, since the new connector (in 1.17-SNAPSHOT) does not
+/// allow specifying a file name, but only a directory to monitor. Rather, the
+/// Nexmark test appears to have used the previous legacy filesystem connector, [stable filesystem connector](https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/connectors/datastream/filesystem/)
+/// which does allow specifying a file path.
+///
+/// Also see [Flink's Join with a Temporal Table](https://nightlies.apache.org/flink/flink-docs-release-1.11/dev/table/streaming/joins.html#join-with-a-temporal-table).
+///
+/// So, although Flink supports monitoring the side-loaded file for updates, a
+/// simple static file is used for this bounded side-input for the Nexmark tests
+/// and that is also what is tested here.
```
absoludity, line 63: Not sure if this makes this query comparable... for example, we have no concept of joining using process time as per the above SQL query. We could implement this manually (updating the hash-map as new data is detected), but it's not needed if the benchmark doesn't use it.

src/nexmark/queries/q13.rs
```diff
@@ -0,0 +1,162 @@
...
+    let mut hm: HashMap<usize, String> = HashMap::new();
+    for result in csv_reader.deserialize() {
+        let (key, val): (usize, String) = result?;
+        hm.insert(key, val);
+    }
+    Ok(hm)
+}
+
+pub fn q13(input: NexmarkStream) -> Q13Stream {
+    let side_input = read_side_input(File::open(SIDE_INPUT_CSV).unwrap()).unwrap();
```
ryzhyk, line 83: We should feed `side_input` via the input handle API as a normal input stream and use the `join` operator to join with `input`. As you describe in the comment above, in general we should implement a file listener that ingests new records as they get appended to the file.  But in this case, since we know that the file is static, we can feed the entire contents as a separate transaction executed before starting the generator. 

Re processing time modeling: Processing time doesn't need to be a first-class primitive -- it's just a computed column added to a stream when it gets ingested by DBSP. One option would be to add it as a `p_time` field to `Bids`. The generator sets this field to zero.  Queries that use processing time set the value of the field, e.g., using `SystemTime::now()`.  The `p_time` field can then be used as any other field to filter/join on.

ryzhyk (changes requested):

src/nexmark/queries/q13.rs
```diff
@@ -0,0 +1,200 @@
...
+        .unwrap()
+        .as_millis() as u64
+}
+
+pub fn q13(input: NexmarkStream, side_input: SideInputStream) -> Q13Stream {
+    // Index bids by the modulo value.
+    let bids_by_auction_mod = input.flat_map_index(move |event| match event {
+        Event::Bid(b) => Some((
+            (b.auction % 10_000) as usize,
+            (b.auction, b.bidder, b.price, b.date_time, process_time()),
```
ryzhyk, line 75: If calling `process_time` for each record is too expensive, we can easily optimize it down to one call per batch.

ryzhyk (approved):

ryzhyk:
Looks good, except for the test failure.

src/nexmark/queries/q13.rs
absoludity: Yep, makes sense. Done.

absoludity:
> Looks good, except for the test failure.

Yes, it took me a while to work out how I can treat q13 differently in the macro (what took a while was discovering that a missing `$` was causing my pattern not to match :/ )

Thanks.

PR #177, opened 2022-09-08
[nexmark] Add q15 - distinct aggregations with filters.

absoludity:
So this was a pain to write for a number of reasons, and even then, I'm not sure that the implementation I've added is a fair comparison: As we were already aggregating to count the number of bids per day, I've used a `Fold` within the same aggregation to also calculate the distinct filtered aggregations (total bidders, total auctions, and their respective ranked versions). The SQL query runs each as a separate filter, but I assume that's because there's no other way to express it. See what you think.

And it's slooow (it's slow in the original too, but faster than their q4, for example), so any performance tips here will be gratefully accepted.

```
$ cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=1000000 --cpu-cores 4 --query q15 --num-event-generators 4 --source-buffer-size 10000 --input-batch-size 10000
    Finished bench [optimized + debuginfo] target(s) in 52.12s
     Running benches/nexmark/main.rs (target/release/deps/nexmark-db7e78eeb5fe8ae0)
Starting q15 bench of 1000000 events...
1,000,000 / 1,000,000 [=====================================================================================================================================================================] 100 % 176102.3803/s 0s
┌───────┬───────────┬───────┬─────────┬─────────────────┬──────────────────┬───────────────┬───────────────┬─────────────┐
│ Query │ #Events   │ Cores │ Elapsed │ Cores * Elapsed │ Throughput/Cores │ Total Usr CPU │ Total Sys CPU │ Max RSS(Kb) │
├───────┼───────────┼───────┼─────────┼─────────────────┼──────────────────┼───────────────┼───────────────┼─────────────┤
│ q15   │ 1,000,000 │ 4     │ 5.679s  │ 22.716s         │ 44.022 K/s       │ 7.434s        │ 0.151s        │ 138,344     │
└───────┴───────────┴───────┴─────────┴─────────────────┴──────────────────┴───────────────┴───────────────┴─────────────┘
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -590,6 +590,7 @@ dependencies = [
  "serde",
  "tar",
  "textwrap 0.15.0",
+ "time",
  "typedmap",
  "zip",
  "zstd",
@@ -2041,12 +2042,11 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.12"
+version = "0.3.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "74b7cc93fc23ba97fde84f7eea56c55d1ba183f495c6715defdfc7b9cb8c870f"
+checksum = "3c3f9a28b618c3a6b9251b6908e9c99e04b9e5c02e6581ccbb67d59c34ef7f9b"
 dependencies = [
  "itoa 1.0.3",
- "js-sys",
  "libc",
  "num_threads",
  "time-macros",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 default = ["with-serde"]
 with-serde = ["serde"]
 with-csv = ["csv"]
-with-nexmark = ["cached", "clap", "csv", "rust_decimal", "rand"]
+with-nexmark = ["cached", "clap", "csv", "rust_decimal", "rand", "time"]
 
 [dependencies]
 anyhow = "1.0.57"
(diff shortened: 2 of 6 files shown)


ryzhyk (commented):
This is a great start.  Computing multiple aggregates at the same time is totally ok, and is in fact how @mbudiu-vmw compiles SQL. What makes this expensive (I suspect) is that all bids in this benchmark have the same date (is that true?), so there is only one group here, and we end up scanning all of it at every `step`.  In addition, since there is only one group, a single worker thread ends up doing all the work, so parallelization doesn't help.   I will play with this query to see if I can improve the performance.

absoludity:
> This is a great start. Computing multiple aggregates at the same time is totally ok, and is in fact how @mbudiu-vmw compiles SQL. What makes this expensive (I suspect) is that all bids in this benchmark have the same date (is that true?), so there is only one group here, and we end up scanning all of it at every `step`.

Yes, but so would the original benchmark right? I mean, We could add the seconds to the `day` field so that we get groups per second, but that'd be a different benchmark. Funny that most of the other queries adjustments like that would be made in the original to make it more useful as a bench...

> In addition, since there is only one group, a single worker thread ends up doing all the work, so parallelization doesn't help. I will play with this query to see if I can improve the performance.

Ah - nice, thanks!

ryzhyk:
> Yes, but so would the original benchmark right? I mean, We could add the seconds to the `day` field so that we get groups per second, but that'd be a different benchmark. Funny that most of the other queries adjustments like that would be made in the original to make it more useful as a bench...

Not necessarily.  There is a way to compute partial aggregates in parallel and combine them in the end, which is what they probably do, and we should support that too.

absoludity:
Let me know if you want this one to merge on its on with your improvements separate, or just wait for your improvements - fine either way.

PR #180, opened 2022-09-11
[nexmark] Optimized q15

ryzhyk:

PR #183, opened 2022-09-12
Size of

Kixiron:
Switches to the [`size-of`](https://crates.io/crates/size-of) crate for memory monitoring and more deeply integrates it within dbsp


src/operator/join.rs
```diff
@@ -69,6 +69,29 @@ where
         )
     }
 
+    #[track_caller]
```
ryzhyk, line 72: In #180 I added `join_index`, which does something similar for incremental joins. We shouldn't need a new operator for this.  Instead, we should generalize `Join` to return `IndexedZSet` (somewhat counterintuitively, `ZSet` is a specialization of `IndexedZSet`) and unify `stream_join_XXX` methods with `join_XXX`.  Feel free to leave this as an issue for me.

ryzhyk (changes requested):

ryzhyk (approved):

PR #181, opened 2022-09-12
[nexmark] Add q7 - Highest bid per tumbling window.

absoludity:
Signed-off-by: Michael Nelson <minelson@vmware.com>

Very similar to Q5, just a max on bid rather than count (and different output, of course).

```
cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=10000000 --cpu-cores 4 --query q7 --num-event-generators 4 --source-buffer-size 10000 --input-batch-size 10000
   Compiling dbsp v0.1.0 (/home/michael/dev/vmware/database-stream-processor)
    Finished bench [optimized + debuginfo] target(s) in 54.82s
     Running benches/nexmark/main.rs (target/release/deps/nexmark-db7e78eeb5fe8ae0)
Starting q7 bench of 10000000 events...
10,000,000 / 10,000,000 [======================================================================================================================================================================================] 100 % 1200036.3645/s 0s
┌───────┬────────────┬───────┬─────────┬─────────────────┬──────────────────┬───────────────┬───────────────┬─────────────┐
│ Query │ #Events    │ Cores │ Elapsed │ Cores * Elapsed │ Throughput/Cores │ Total Usr CPU │ Total Sys CPU │ Max RSS(Kb) │
├───────┼────────────┼───────┼─────────┼─────────────────┼──────────────────┼───────────────┼───────────────┼─────────────┤
│ q7    │ 10,000,000 │ 4     │ 8.334s  │ 33.335s         │ 299.987 K/s      │ 31.673s       │ 1.054s        │ 1,465,516   │
└───────┴────────────┴───────┴─────────┴─────────────────┴──────────────────┴───────────────┴───────────────┴─────────────┘
```

PR #185, opened 2022-09-14
[nexmark] Use a WATERMARK_INTERVAL_SECONDS of 4, matching Java impl.

absoludity:

Based on conversation with Leonid. Although this [separate comment](https://github.com/vmware/database-stream-processor/pull/181#discussion_r969260608):

> We should use `WATERMARK` constant here, which should probably match the out-of-ordedness bounds used by the generator.

confuses me, in that the out-of-orderness bounds are configurable (`--out-of-order-group-size`) only in terms of the size of the group. Anyway, the small change here just updates to use the watermark constant set based on the Java implementation. Let me know if that's not what you meant.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,3 @@
-#![feature(generic_associated_types)]
 #![cfg_attr(feature = "with-nexmark", feature(is_some_with))]
 
 mod error;
--- a/src/nexmark/queries/mod.rs
+++ b/src/nexmark/queries/mod.rs
@@ -18,6 +18,9 @@ pub use q15::q15;
 
 type NexmarkStream = Stream<Circuit<()>, OrdZSet<Event, isize>>;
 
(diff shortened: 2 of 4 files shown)


ryzhyk (approved):

ryzhyk:
This is great, thanks!  I was under the impression that out-of-ordedness in Java is correlated with the watermark declaration in SQL.  Sounds like it's not the case.

absoludity:
> This is great, thanks! I was under the impression that out-of-ordedness in Java is correlated with the watermark declaration in SQL. Sounds like it's not the case.

Ah - ok. Right, the out-of-orderness is just based on the `out-of-order-groupsize` - so a value of 5 means that within 5 events, the events may be out of order, but the 6th event will correctly come after those 5 (but may be out of order within the next group of 5) etc..

PR #186, opened 2022-09-14
[nexmark] Add Q12 demoing a process-time window.

absoludity:
This query demo's grouping and aggregation using process-time windows.

There are two TODOs that I'd be keen for any insight on, one is DBSP related (calling process_time() once per batch rather than on every event?), the other is related to the way I'm testing this query by injecting a vector of process times to emit.

Otherwise, 40s elapsed for 100M events (4 cores), compared to the Java implementation of 96s with 13 cores. DBSP is faring pretty well!

```
cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=100000000 --cpu-cores 4 --query q12 --num-event-generators 4 --source-buffer-size 10000 --input-batch-size 10000
    Finished bench [optimized + debuginfo] target(s) in 0.12s
     Running benches/nexmark/main.rs (target/release/deps/nexmark-db7e78eeb5fe8ae0)
Starting q12 bench of 100000000 events...
100,000,000 / 100,000,000 [=====================================================================================================================================================] 100 % 2562005.5364/s 0s
┌───────┬─────────────┬───────┬─────────┬─────────────────┬──────────────────┬───────────────┬───────────────┬─────────────┐
│ Query │ #Events     │ Cores │ Elapsed │ Cores * Elapsed │ Throughput/Cores │ Total Usr CPU │ Total Sys CPU │ Max RSS(Kb) │
├───────┼─────────────┼───────┼─────────┼─────────────────┼──────────────────┼───────────────┼───────────────┼─────────────┤
│ q12   │ 100,000,000 │ 4     │ 39.033s │ 156.131s        │ 640.487 K/s      │ 258.609s      │ 6.593s        │ 369,768     │
└───────┴─────────────┴───────┴─────────┴─────────────────┴──────────────────┴───────────────┴───────────────┴─────────────┘
```

### DIFF
--- a/src/nexmark/queries/q12.rs
+++ b/src/nexmark/queries/q12.rs
@@ -41,23 +41,13 @@ fn window_for_process_time(ptime: u64) -> (u64, u64) {
 // This function enables us to test the q12 functionality without using the
 // actual process time, while the actual q12 function below uses the real
 // process time.
-// TODO: I originally planned to pass a FnMut closure for process_time that
-// just emits a new u64 each time it is called, but can't do this as the
-// closure of `flat_map_index` is Fn not FnMut, and would need to capture the
-// process_time closure. So right now, it's quite ugly: to avoid an `FnMut`,
-// I'm instead passing an optional vector of times with the assumption that
-// those times are indexed by the bid.auction.
-// There must be a better way without resorting to interior mutability? Anyway,
-// it works for the tests and is only used in the tests.
-fn q12_for_process_time(input: NexmarkStream, process_times: Option<Vec<u64>>) -> Q12Stream {
+fn q12_for_process_time<F>(input: NexmarkStream, process_time: F) -> Q12Stream
+where
+    F: Fn() -> u64 + 'static,
+{
     let bids_by_bidder_window = input.flat_map_index(move |event| match event {
-        // TODO: Can I call process_time() just once per batch, rather than for every Bid? How?
         Event::Bid(b) => {
-            let t = match &process_times {
-                Some(v) => v[b.auction as usize],
-                None => process_time(),
-            };
-            let (starttime, endtime) = window_for_process_time(t);
+            let (starttime, endtime) = window_for_process_time(process_time());
             Some(((b.bidder, starttime, endtime), ()))
         }
         _ => None,
@@ -69,7 +59,7 @@ fn q12_for_process_time(input: NexmarkStream, process_times: Option<Vec<u64>>) -
 }
 
 pub fn q12(input: NexmarkStream) -> Q12Stream {
-    q12_for_process_time(input, None)
+    q12_for_process_time(input, process_time)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


src/nexmark/queries/q12.rs
```diff
@@ -0,0 +1,150 @@
...
+// actual process time, while the actual q12 function below uses the real
+// process time.
+// TODO: I originally planned to pass a FnMut closure for process_time that
+// just emits a new u64 each time it is called, but can't do this as the
+// closure of `flat_map_index` is Fn not FnMut, and would need to capture the
+// process_time closure. So right now, it's quite ugly: to avoid an `FnMut`,
+// I'm instead passing an optional vector of times with the assumption that
+// those times are indexed by the bid.auction.
+// There must be a better way without resorting to interior mutability? Anyway,
+// it works for the tests and is only used in the tests.
```
ryzhyk, line 51: If it's only used in tests where performance doesn't matter, you should be able to get away with `Fn` by relying on internal mutability, by wrapping whatever mutable state the function needs in `Arc<Mutex<>>`.

src/nexmark/queries/q12.rs
```diff
@@ -0,0 +1,150 @@
...
+fn q12_for_process_time(input: NexmarkStream, process_times: Option<Vec<u64>>) -> Q12Stream {
+    let bids_by_bidder_window = input.flat_map_index(move |event| match event {
+        // TODO: Can I call process_time() just once per batch, rather than for every Bid? How?
```
ryzhyk, line 54: Since the performance is looking good, I wouldn't bother.

ryzhyk: I see even better performance when increasing the number of generators to, so this is one of those queries where we struggle to produce inputs fast enough.  Nice!

Unrelated: most queries run faster when I set `input-batch-size` to 40K.  Shall we make that that a default or is this cheating somehow?  With 40K batch size, 6 workers, and 8 generators I get:

```
cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=100000000 --cpu-cores 6 --query q12 --num-event-generators 8 --source-buffer-size 10000 --input-batch-size 40000
    Finished bench [optimized + debuginfo] target(s) in 0.06s
     Running benches/nexmark/main.rs (target/release/deps/nexmark-db7e78eeb5fe8ae0)
Starting q12 bench of 100000000 events...
100,000,000 / 100,000,000 [==========================================================================================================================================] 100 % 6672068.9755/s 0s
┌───────┬─────────────┬───────┬─────────┬─────────────────┬──────────────────┬───────────────┬───────────────┬─────────────┐
│ Query │ #Events     │ Cores │ Elapsed │ Cores * Elapsed │ Throughput/Cores │ Total Usr CPU │ Total Sys CPU │ Max RSS(Kb) │
├───────┼─────────────┼───────┼─────────┼─────────────────┼──────────────────┼───────────────┼───────────────┼─────────────┤
│ q12   │ 100,000,000 │ 6     │ 14.989s │ 89.932s         │ 1111.955 K/s     │ 170.236s      │ 4.562s        │ 414,800     │
└───────┴─────────────┴───────┴─────────┴─────────────────┴──────────────────┴───────────────┴───────────────┴─────────────┘

```

ryzhyk (approved):

src/nexmark/queries/q12.rs
absoludity: In fact, with q0 (which doesn't even do string copying - literally a noop) with the same options I get 29s (or 32s with `--input-batch-size 1000`). I assume you get way under 15s with q0? The fact that you're getting only 15s with q12 also means that there's only one or two process windows during the benchmark (perhaps not an issue, and we can easily change the size of the window for q12 if we want).

src/nexmark/queries/q12.rs
absoludity: Since each test runs in its own thread, it seems that a `RefCell<>` is enough? Pushed, I'll land if everything passes.

PR #188, opened 2022-09-15
Generalize `Upsert`, use it in aggregation.

ryzhyk:
The `Upsert` operator was introduced as part of the implementation of `UpsertHandle` (i.e., an input data source that generates inputs with upsert semantics).  In a nutshell, it takes a vector of `(key, value)` pairs and generates a delta that removes the old value for each `key` and inserts the new value with weight `1`.

Nearly identical logic is used in the implementation of `aggregate`: it computes the new value for each affected key from scratch and then looks up the old aggregate value in its output trace and emits an insertion and a retraction update.  I am currently working on a new operator that implements the SQL `OVER` clause, which also uses upsert for a similar reason.

To avoid duplicate logic, we generalize the `Upsert` operator to work in arbitrary scopes, and break up the implementation of `aggregate` into the actual aggregation logic that computes the new value of the aggregate, and an upsert, which takes care of emitting any necessary insertions and retractions.

### DIFF
--- a/src/operator/upsert.rs
+++ b/src/operator/upsert.rs
@@ -1,140 +1,206 @@
 use crate::{
-    algebra::{HasOne, HasZero},
+    algebra::{AddAssignByRef, HasOne, HasZero, PartialOrder, ZRingValue},
     circuit::{
         operator_traits::{BinaryOperator, Operator},
-        OwnershipPreference, Scope,
+        ExportId, ExportStream, OwnershipPreference, Scope,
     },
-    trace::{cursor::Cursor, Batch, Trace},
+    operator::trace::{DelayedTraceId, TraceAppend, TraceId, Z1Trace},
+    trace::{
+        consolidation::consolidate, cursor::Cursor, spine_fueled::Spine, Batch, Builder, Trace,
+        TraceReader,
+    },
+    utils::VecExt,
+    Circuit, Stream, Timestamp,
 };
-use std::{borrow::Cow, marker::PhantomData, mem::swap, ops::Neg};
+use size_of::SizeOf;
+use std::{borrow::Cow, marker::PhantomData, ops::Neg};
 
-pub struct Upsert<K, VI, F, V, T, B> {
-    val_func: F,
-    phantom: PhantomData<(K, VI, V, T, B)>,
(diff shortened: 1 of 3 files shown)


src/operator/upsert.rs
```diff
@@ -1,140 +1,206 @@
...
+    where
+        K: Ord + Clone + Eq + SizeOf + 'static,
+        V: Ord + Clone + Ord + SizeOf + 'static,
+        B::R: ZRingValue + SizeOf,
+        TS: Timestamp + SizeOf,
+        B: Batch<Key = K, Val = V, Time = ()> + 'static,
+    {
+        let circuit = self.circuit();
+
+        // We build the following circuit to implement the upsert semantics.
```
mbudiu-vmw, line 49: can distinct be implemented using upsert too?

mbudiu-vmw (approved):
I wonder whether this can be generalized to add an arbitrary weight, not just 1.
Then this could be used to implement windows - where upserting with a weight of 0 is removal from the window

src/operator/upsert.rs
ryzhyk: There are two ways to implement `distinct`: (1) assemble output into a trace and use upsert to determine values to retract, (2) compute retractions dynamically at the same time we scan the _input_ trace.  The second option can be more memory-efficient as it doesn't require output trace, and that's what we implement today, hence no need for upsert.

mbudiu-vmw: but distinct needs to keep the output trace somewhere too, right?

ryzhyk: It doesn't if you compute the value to retract from the input trace.

mbudiu-vmw: Is the input trace the integral of the input?
distinct needs to keep the integral around.

ryzhyk: A trace is the history of the input.  It can be used to compute the integral for any timestamp, current or past.

mbudiu-vmw: We need a glossary. I think of a trace as Frank's datatype.

ryzhyk: Yes, it is.

PR #190, opened 2022-09-16
Fix typo in `OrdKeyMerger`.

ryzhyk:
A typo caused time bounds of a merged batch to be computed incorrectly. Found by @gz .


gz (commented):
LGTM, at least the persistence tests pass now with this change

gz (approved):

PR #191, opened 2022-09-19
[nexmark] Add q18 - illustrating deduplicate query.

absoludity:

This one is simpler in DBSP than it is in Flink SQL, I think (ie. just create the index and join). Though it has an even higher memory usage than q17, as it never discards old data from the index?

```
cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=10000000 --cpu-cores 8 --query q18 --num-event-generators 6 --source-buffer-size 10000 --input-batch-size 40000 
    Finished bench [optimized + debuginfo] target(s) in 0.99s
     Running benches/nexmark/main.rs (target/release/deps/nexmark-733304a45ec676db)
Starting q18 bench of 10000000 events...
10,000,000 / 10,000,000 [==================================================================================================================================================================] 100 % 1767134.2974/s 0s
│ q18   │ 10,000,000 │ 8     │ 5.677s  │ 45.417s         │ 220.182 K/s      │ 34.054s       │ 1.477s        │ 2,161,116   │
```

(again, unable to run complete 100M query without EOM).

That said, although it's not in the original query explicitly (but is implicitly), this query will probably benefit from only indexing a window of the data?

### DIFF
--- a/src/nexmark/queries/q18.rs
+++ b/src/nexmark/queries/q18.rs
@@ -1,7 +1,7 @@
 use super::NexmarkStream;
 use crate::{
     nexmark::model::{Bid, Event},
-    operator::{FilterMap, Max},
+    operator::{FilterMap, Fold},
     Circuit, OrdZSet, Stream,
 };
 
@@ -35,21 +35,17 @@ type Q18Stream = Stream<Circuit<()>, OrdZSet<Bid, isize>>;
 
 pub fn q18(input: NexmarkStream) -> Q18Stream {
     let bids_by_auction_bidder = input.flat_map_index(|event| match event {
-        Event::Bid(b) => Some(((b.auction, b.bidder), b.date_time)),
+        Event::Bid(b) => Some(((b.auction, b.bidder), b.clone())),
         _ => None,
     });
 
-    let latest_bid_per_auction_bidder = bids_by_auction_bidder
-        .aggregate::<(), _>(Max)
-        .map(|(&(auction, bidder), &date_time)| ((auction, bidder, date_time), ()))
-        .index();
-
-    input
-        .flat_map_index(|event| match event {
-            Event::Bid(b) => Some(((b.auction, b.bidder, b.date_time), b.clone())),
-            _ => None,
-        })
-        .join::<(), _, _, _>(&latest_bid_per_auction_bidder, |_, bid, _| bid.clone())
+    bids_by_auction_bidder
+        .aggregate::<(), _>(Fold::new(Bid::default(), |top: &mut Bid, val: &Bid, _w| {
+            if val.date_time > top.date_time {
+                *top = val.clone();
+            }
+        }))
+        .map(|((_, _), bid)| bid.clone())
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


absoludity:
> This one is simpler in DBSP than it is in Flink SQL, I think (ie. just create the index and join).

Actually, I just realised while working on the next query that this implementation is slightly different in that the flink `ROW_NUMBER()` is assigning a unique row number to each, so if there are two bids from the same bidder with the same timestamp (maybe we don't need to care as this isn't a case to worry about) then it would still return only one of those as the latest, whereas the DBSP query could return both (I need to come back and update the test to verify).

src/nexmark/queries/q18.rs
```diff
@@ -0,0 +1,401 @@
...
+type Q18Stream = Stream<Circuit<()>, OrdZSet<Bid, isize>>;
+
+pub fn q18(input: NexmarkStream) -> Q18Stream {
+    let bids_by_auction_bidder = input.flat_map_index(|event| match event {
+        Event::Bid(b) => Some(((b.auction, b.bidder), b.date_time)),
+        _ => None,
+    });
+
+    let latest_bid_per_auction_bidder = bids_by_auction_bidder
+        .aggregate::<(), _>(Max)
```
ryzhyk, line 43: I think the `join` below can be avoided by aggregating over bids and returning the bid with the largest `date_time` value.  Ideally this requires an `ArgMax` aggregator (returns the record that maximizes the value of a user-defined closure) but you can also do it with `Fold`.

absoludity: True, done. Performance is slightly better for 10M:

```
cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=10000000 --cpu-cores 8 --query q18 --num-event-generators 6 --source-buffer-size 10000 --input-batch-size 40000
    Finished bench [optimized + debuginfo] target(s) in 3m 35s
     Running benches/nexmark/main.rs (target/release/deps/nexmark-733304a45ec676db)
Starting q18 bench of 10000000 events...
10,000,000 / 10,000,000 [======================================================================================================================================================================================] 100 % 1910681.4912/s 0s
│ q18   │ 10,000,000 │ 8     │ 5.235s  │ 41.876s         │ 238.798 K/s      │ 32.143s       │ 1.394s        │ 1,999,048   │
```

But still unable to run with 100M (with the same args)
```
$ cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=10000000 --max-events=100000000 --cpu-cores 8 --query q18 --num-event-generators 6 --source-buffer-size 10000 --input-batch-size 40000 
     Running benches/nexmark/main.rs (target/release/deps/nexmark-733304a45ec676db)
Starting q18 bench of 100000000 events...
83,200,000 / 100,000,000 [========================================================================================================================================================>------------------------------] 83 % 166423.4625/s 2merror: bench failed, to rerun pass `--bench nexmark`

Caused by:
  process didn't exit successfully: `/home/michael/dev/vmware/database-stream-processor/target/release/deps/nexmark-733304a45ec676db --first-event-rate=10000000 --max-events=100000000 --cpu-cores 8 --query q18 --num-event-generators 6 --source-buffer-size 10000 --input-batch-size 40000 --bench` (signal: 9, SIGKILL: kill)
```

ryzhyk (approved):

PR #200, opened 2022-10-03
Structured operator metadata

Kixiron:
Adds structured operator metadata, allowing for more detailed and machine-readable operator statistic reporting


src/operator/join.rs
```diff
@@ -651,10 +704,8 @@ where
     It: IntoIterator<Item = (Z::Key, Z::Val)> + 'static,
 {
     fn eval(&mut self, index: &I, trace: &T) -> Z {
-        /*println!("JoinTrace::eval@{}:\n  index:\n{}\n  trace:\n{}",
-        self.time,
-        textwrap::indent(&index.to_string(), "    "),
-        textwrap::indent(&trace.to_string(), "    "));*/
+        self.stats.lhs_tuples += index.len();
+        self.stats.rhs_tuples += trace.len();
```
ryzhyk, line 708: This is probably not a good metric.  `index` is a batch of changes and it makes sense to sum them up, but `trace` stores accumulated changes of the rhs stream, so it doesn't make sense to sum them up again.

src/operator/join.rs
```diff
@@ -590,35 +611,67 @@ where
...
+            let mut context = Context::new();
+            for batcher in self.output_batchers.values() {
+                batcher.size_of_with_context(&mut context);
+            }
+
+            context.size_of()
+        };
+
+        // Find the percentage of consolidated outputs
+        let mut output_redundancy = ((self.stats.output_tuples as f64
```
ryzhyk, line 653: Not sure this metric makes sense.  `produced_tuples` are tuples sent to the output stream of the operator. `output_tuples` are all computed outputs, including those stored inside `output_batchers`, which will be pushed to the output stream later.  It's not clear to me what this ratio represents.

src/operator/join.rs
```diff
@@ -590,35 +611,67 @@ where
...
-        .unwrap();
+        meta.extend(metadata! {
+            "total size" => total_size,
+            "batch sizes" => batch_sizes,
+            "used bytes" => MetaItem::bytes(bytes.used_bytes()),
+            "allocations" => bytes.distinct_allocations(),
+            "shared bytes" => MetaItem::bytes(bytes.shared_bytes()),
+            "left inputs" => self.stats.lhs_tuples,
+            "right inputs" => self.stats.rhs_tuples,
+            "raw outputs" => self.stats.output_tuples,
```
ryzhyk, line 671: "computed_outputs" might be a better name.

src/operator/trace.rs
```diff
@@ -374,31 +375,26 @@ where
...
-            HumanBytes::from(bytes.total_bytes()),
-            HumanBytes::from(bytes.used_bytes()),
-            bytes.distinct_allocations(),
-            HumanBytes::from(bytes.shared_bytes()),
-        )
-        .unwrap();
+
+        meta.extend(metadata! {
+            "total size" => total_size,
+            "allocated bytes" => MetaItem::bytes(bytes.total_bytes()),
```
ryzhyk, line 393: It might be worth having a method to convert `TotalSize` to metadata.  If we also support nested metadata objects, all `TotalSize` elements could be grouped under one attribute.

src/operator/z1.rs
```diff
@@ -188,19 +189,30 @@ where
...
-            "allocated: {}, used: {}, allocations: {}, shared: {}",
-            HumanBytes::from(bytes.total_bytes()),
-            HumanBytes::from(bytes.used_bytes()),
-            bytes.distinct_allocations(),
-            HumanBytes::from(bytes.shared_bytes()),
-        )
-        .unwrap();
+        meta.extend([
+            (
+                Cow::Borrowed("total size"),
```
ryzhyk, line 196: use macro here?

src/operator/join.rs
Kixiron: It's basically the ratio of produced outputs to consolidated outputs, if `(x, y)` is produced twice then `produced_tuples` will have a value of 2 while `output_tuples` will have a value of 1. It's basically a metric of "how many duplicate outputs are made from this join?"

ryzhyk: Yes, I figured this is what you were going for, but I don't think this formula reflects that due to the way join buffers outputs for future time stamps.

ryzhyk: Unrelated: can't this cause a divide-by-zero exception?

Kixiron: No because of the magic of floats, it turns into nan or inf (hence the nan/inf check afterwards)

ryzhyk: Is that true in both debug and release builds?

Kixiron: Yes

benches/ldbc-graphalytics/main.rs
```diff
@@ -333,31 +339,45 @@ fn attach_profiling(dataset: DataSet, circuit: &mut Circuit<()>) {
...
+                            MetaItem::Duration(profile.total_time()),
+                        ),
+                    ];
+
+                    for item in default_meta {
+                        meta.insert(0, item);
+                    }
+                }
+
+                for (label, item) in &meta {
```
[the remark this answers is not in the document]
Kixiron: I don't think so since it's fairly dot-specific

src/operator/join.rs
Kixiron: What would you do instead?

ryzhyk: Perhaps just: `self.stats.rhs_tuples = trace.len();` (and we don't need to invoke it here for each input batch, just once in `fn metadata()`).

Kixiron: We don't have access to the trace in `.metadata()` so I may just assign it every time, shouldn't really matter that much

ryzhyk: Right, I wasn't thinking.

ryzhyk:
@Kixiron , is this ready for review?

ryzhyk (approved):

PR #202, opened 2022-10-07
Bug in `map_times_through`.

ryzhyk:


src/trace/ord/key_batch.rs
```diff
@@ -329,11 +329,13 @@ where
...
-                self.cursor.child.current_key(),
-                self.cursor.child.current_diff(),
-            );
+        while self.cursor.child.valid() {
+            if self.cursor.child.key().0.less_equal(upper) {
+                logic(
+                    self.cursor.child.current_key(),
+                    self.cursor.child.current_diff(),
+                );
+            }
```
Kixiron, line 338: We probably need some tests covering this

ryzhyk: We do now, in aggregation operator test harness.  We do need dedicated unit tests for cursors though.

Kixiron: Do we? I mentioned it because if the coverage warning saying the function was uncovered

ryzhyk: Ah, I guess that one tests `val_batch`

PR #203, opened 2022-10-10
Remove `TraceReader` trait.

ryzhyk:
`TraceReader` only defined a single method, `map_batches`.  Work on persistent traces showed that this is not a good design, as it exposes the internal structure of the trace.  We only used this API in one place in `distinct.rs`, and that code would have to be re-written anyway as we generalize `distinct` to work in arbitrary nested scopes.

We eliminate the `map_batches` method from the public API and downgrade it to a private method of `Spine`.  We also remove the no longer useful trait `TraceReader`.

Addresses a TODO in PR #124.

### DIFF
--- a/src/operator/distinct.rs
+++ b/src/operator/distinct.rs
@@ -9,17 +9,16 @@ use crate::{
     },
     circuit_cache_key,
     time::NestedTimestamp32,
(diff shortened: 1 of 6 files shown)


gz (approved):
lgtm

mbudiu-vmw (approved):

PR #207, opened 2022-10-11
Simplify trait bounds.

ryzhyk:
Introduce trait bounds that capture common features of relational data:
 
`DBData` - trait bound on `BatchReader::Key`, `BatchReader::Val`
`DBWeight: DBData + MonoidValue` - trait bound on `BatchReader::R`
`DBTime: DBData + Timestamp` - trait bound on `BatchReader::Time`

This simplifies trait bounds throughput the code base thanks to: (1) implied bounds, (2) shorted bounds when type inference doesn't apply.

This also simplifies future refactoring as we adjust the bounds, e.g., to support persistence.

@gz: It should now be sufficient to add `encode/decode` bounds to `DBData`.

### DIFF
--- a/src/trace/mod.rs
+++ b/src/trace/mod.rs
@@ -27,12 +27,29 @@ use crate::{
 use size_of::SizeOf;
 use std::{fmt::Debug, hash::Hash};
 
+/// Trait for data stored in batches.
+///
+/// This trait is used as a bound on `BatchReader::Key` and `BatchReader::Val`
+/// associated types (see [`trait BatchReader`]).  Hence when writing code that
+/// must be generic over any relational data, it is sufficient to impose
+/// `DBData` as a trait bound on types.  Conversely, a trait bound of the form
+/// `B: BatchReader` implies `B::Key: DBData` and `B::Val: DBData`.
 pub trait DBData: Clone + Eq + Ord + Hash + SizeOf + Send + Debug + 'static {}
 impl<T> DBData for T where T: Clone + Eq + Ord + Hash + SizeOf + Send + Debug + 'static {}
 
+/// Trait for data types used as weights.
+///
+/// A type used for weights in a batch (i.e., as `BatchReader::R`) must behave
+/// as a monoid, i.e., a set with an associative `+` operation and a neutral
+/// element (zero).
+///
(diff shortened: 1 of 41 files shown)


src/trace/mod.rs
```diff
@@ -19,19 +19,47 @@ pub mod spine_fueled;
...
+use std::{fmt::Debug, hash::Hash};
+
+/// Trait for data stored in batches.
+///
+/// This trait is used as a bound on `BatchReader::Key` and `BatchReader::Val`
+/// associated types (see [`trait BatchReader`]).  Hence when writing code that
+/// must be generic over any relational data, it is sufficient to impose
+/// `DBData` as a trait bound on types.  Conversely, a trait bound of the form
+/// `B: BatchReader` implies `B::Key: DBData` and `B::Val: DBData`.
+pub trait DBData: Clone + Eq + Ord + Hash + SizeOf + Send + Debug + 'static {}
```
Kixiron, line 37: What's the `Debug` requirement for?

src/trace/mod.rs
```diff
@@ -104,18 +132,18 @@ pub trait Trace: BatchReader {
...
 /// construction of batches from the type of data in the view (for example,
 /// filtered views, or views with extended time coordinates).
-pub trait BatchReader
+pub trait BatchReader: NumEntries + SizeOf + 'static
 where
     Self: Sized,
 {
     /// Key by which updates are indexed.
-    type Key;
+    type Key: DBData;
```
mbudiu-vmw, line 140: I guess this is where most of the simplifications come from

mbudiu-vmw (approved):

src/trace/mod.rs
ryzhyk: Debugging :)

I changed `impl Display` for all batch types to require `Key: Debug`, `Val: Debug`, `Time: Debug`, `R: Debug` (otherwise it's nearly useless).  This way I will be easily dump the contents of the batch in any operator implementation without adding `Debug` bounds everywhere.

src/trace/mod.rs
ryzhyk: yes

gz (approved):

Kixiron (approved):

PR #210, opened 2022-10-12
CI uses self-hosted runners.

gz:


gz:
Looks like it also shaves few minutes off the sanitizer and cargo tests which is always good (though I fear I'm going to utterly destroy any savings with merging the persistent branch again)

.github/workflows/main.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           - {
               rust: nightly,
               target: x86_64-unknown-linux-gnu,
-              os: ubuntu-latest,
+              os: [self-hosted, Linux, skylake-2x],
```
ryzhyk, line 34: What does this do, i.e., why three different OSs?

mbudiu-vmw: Why not windows?

gz: This is just one OS (it's ubuntu 22.04 that runs on the self-hosted runners). The list here is an ANDed list of tags, if you go to Settings->Actions->Runner you'll see the tags that the currently available runners have and you can use them in the OS config yaml to select the right self-hosted runners.

The self-hosted and Linux tag is automatically given by the runner software, the skylake-2x tag is one I gave when setting up the runner to identify the server (skylake CPU, 2-socket numa node).

Technically, just skylake-2x or self-hosted etc. would be enough since we have just one self-hosted runner type now (two exactly identical machines).

@mbudiu-vmw It was running on `ubuntu-latest` before so the PR should preserve that the test runs on the same type of OS. The windows tests still exist (though they do not run self-hosted as we don't have any windows servers).

ryzhyk: I thought if it's used in `matrix` it will create three separate jobs. Apparently I was wrong.

ryzhyk (approved):

PR #211, opened 2022-10-14
require serde for floats

mbudiu-vmw:

This change allows dbsp to run all tests fine.
However, clippy is unhappy:
```
Checking Rust rules prior to push.  To run this check by hand invoke 'tools/pre-push'
+ cargo fmt -- --check
+ cargo clippy -- -D warnings
    Checking dbsp v0.1.0 (/home/mbudiu/git/database-stream-processor)
error: cannot find derive macro `Serialize` in this scope
   --> src/algebra/floats.rs:15:90
    |
15  |               #[derive(Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, SizeOf, Serialize, Deserialize)]
    |                                                                                            ^^^^^^^^^
...
383 | / float! {
384 | |     F32(f32),
385 | |     F64(f64),
386 | | }
    | |_- in this macro invocation
    |
note: `Serialize` is imported here, but it is only a trait, without a derive macro
   --> src/algebra/floats.rs:3:26
    |
3   | use serde::{Deserialize, Serialize};
    |                          ^^^^^^^^^
    = note: this error originates in the macro `float` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I would appreciate help fixing this problem.
The SQL layer will need support for float serde.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1386,6 +1386,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "129d36517b53c461acc6e1580aeb919c8ae6708a4b1eae61c4463a615d4f0411"
 dependencies = [
  "num-traits",
+ "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Kixiron:
You need to put the derives under `with-serde` feature gates

PR #212, opened 2022-10-14
ML feature extraction benchmark.

ryzhyk:
Initial implementation of the credit card fraud benchmark and related infrastructure improvements.  See individual commits for details.


benches/fraud.rs
```diff
@@ -0,0 +1,399 @@
...
+    gender: u32,
+    street: u32,
+    city: u32,
+    state: u32,
+    zip: u32,
+    lat: F64,
+    long: F64,
+    city_pop: u32,
+    job: u32,
+    #[size_of(skip)]
```
mbudiu-vmw, line 131: what does this do and why is it possible?

src/operator/aggregate/average.rs
```diff
@@ -10,35 +13,74 @@ use std::{
...
+
+    /// Returns the `count` component of the `(sum, count)` tuple.
+    pub fn count(&self) -> R
+    where
+        R: Clone,
+    {
+        self.count.clone()
+    }
+
+    /// Returns `sum / count` or `None` if `count` is zero.
```
mbudiu-vmw, line 52: If count is zero the sum better be zero too, and then 0 is also an acceptable result.
In SQL it is NULL anyway

src/operator/time_series/rolling_aggregate.rs
```diff
@@ -15,17 +17,82 @@ use crate::{
...
     OrdPartitionedIndexedZSet<PK, TS, (V, Option<A>), R>;
 pub type OrdPartitionedOverStream<PK, TS, V, A, R> =
     Stream<Circuit<()>, OrdPartitionedOverBatch<PK, TS, V, A, R>>;
 
+/// `Aggregator` object that computes a linear aggregation function.
+// TODO: we need this because we currently compute linear aggregates
+// using the same algorithm as general aggregates.  Additional performance
+// gains can be obtained with an optimized implementation of radix trees
+// for linear aggregates (specifically, updating a node when only
+// some of its children have chanded can be done without computing
```
mbudiu-vmw, line 38: changed

mbudiu-vmw: how come rust does not spellcheck comments?

src/operator/time_series/rolling_aggregate.rs
```diff
@@ -133,6 +200,49 @@ impl<B> Stream<Circuit<()>, B> {
             output
         })
     }
+
+    /// A version of [`Self::partitioned_rolling_aggregate`] optimized for
+    /// linear aggregation functions.
+    ///
+    /// This method only works for linear aggregation functions `f`, i.e.,
```
mbudiu-vmw, line 207: but linear in the group of result values, not in the ZSet group

src/hash.rs
```diff
@@ -0,0 +1,13 @@
+//! Hashing utilities.
+
+use std::hash::{Hash, Hasher};
+use twox_hash::Xxh3Hash64;
+
+const SEED: u64 = 0x7f95_ef85_be33_c337u64;
+
+/// Default hashing function used to shard records across workers.
+pub fn default_hash<T: Hash>(x: &T) -> u64 {
```
mbudiu-vmw, line 9: What are the rules about the hash function? Do users need to know about this? Can they get a program wrong by somehow mixing multiple different hash functions? If so this has to be documented somewhere.

benches/fraud.rs
```diff
@@ -0,0 +1,399 @@
...
+    stdin: bool,
+
+    #[doc(hidden)]
+    #[clap(long = "bench", hide = true)]
+    __bench: bool,
+}
+
+type Weight = i32;
+
+struct Fraud {
```
mbudiu-vmw, line 188: I don't like the implication that DBSP is part of a Fraud.
Why don't you call this FraudBenchmarkHandle?

benches/fraud.rs
```diff
@@ -0,0 +1,399 @@
...
+                (t.cc_num, (timestamp, t.amt))
+            });
+
+            let transactions_by_ccnum = transactions.map_index(|t| (t.cc_num, t.clone()));
+            let demographics_by_ccnum = demographics.map_index(|d| (d.cc_num, d.clone()));
+
+            let enriched_transactions: Stream<
+                _,
+                OrdIndexedZSet<(F64, i64), (Transaction, Demographics), Weight>,
+            > = transactions_by_ccnum.join_index::<(), _, _, _, _, _>(
```
mbudiu-vmw, line 211: why does join_index need a bound with almost everything _?

benches/fraud.rs
```diff
@@ -0,0 +1,399 @@
...
+                |cc_num, tran, dem| {
+                    let timestamp = tran.trans_date_trans_time.assume_utc().unix_timestamp();
+                    Some(((*cc_num, timestamp), (tran.clone(), dem.clone())))
+                },
+            );
+
+            // AVG(amt) OVER(
+            //     PARTITION BY CAST(cc_num AS NUMERIC)
+            //     ORDER BY unix_time
+            //     -- 1 week is 604800  seconds
```
mbudiu-vmw, line 222: best to have an explicit definition WEEK_IN_SECONDS = 7 * 24 * 3600;

benches/fraud.rs
```diff
@@ -0,0 +1,399 @@
...
+            println!("{total_size} parsing: {} ", chunk_start.elapsed());
+            total_size += batch_size;
+
+            tx.send(true).unwrap();
+        }
+
+        tx.send(false).unwrap();
+        thread_handle.join().unwrap();
+
+        println!("total time: {}", start.elapsed());
```
mbudiu-vmw, line 371: how do you know the results are correct?

mbudiu-vmw (approved):

benches/fraud.rs
ryzhyk: It tells the `size_of` crate to skip this field when calculating the size of the struct.  This it no longer necessary as @Kixiron added support for the `time` crate to `size_of`.

src/operator/aggregate/average.rs
ryzhyk: > If count is zero the sum better be zero too, and then 0 is also an acceptable result.

This isn't obvious to me. 0 seems like an arbitrary choice in this case.

src/operator/time_series/rolling_aggregate.rs
ryzhyk: I think it's linear in both.

src/hash.rs
ryzhyk: I don't think there are any special requirements.

benches/fraud.rs
ryzhyk: Most of its type arguments are automatically inferred, except the logical timestamp. There are todo's in the code to refactor things so that this type is derived from the circuit.  Once that is done, `join` and variants won't need explicit type arguments.

benches/fraud.rs
ryzhyk: I don't.  We need a test harness.

benches/fraud.rs
```diff
@@ -0,0 +1,401 @@
...
+    #[clap(long = "bench", hide = true)]
+    __bench: bool,
+}
+
+type Weight = i32;
+
+struct FraudBenchmark {
+    dbsp: DBSPHandle,
+    hdemographics: CollectionHandle<Demographics, Weight>,
+    htransactions: CollectionHandle<Transaction, Weight>,
```
Kixiron, line 193:
```suggestion
    demographics: CollectionHandle<Demographics, Weight>,
    transactions: CollectionHandle<Transaction, Weight>,
```

Cargo.toml
```diff
@@ -29,31 +29,31 @@ with-nexmark = [
...
-fxhash = "0.2"
-ordered-float = "3.0.0"
+ordered-float = { version = "3.0.0", features = ["serde"]}
 arcstr = { version = "1.1.4", optional = true }
 rust_decimal = { version = "1.26.1", optional = true }
 regex = { version = "1.6.0", optional = true }
 time = { version = "0.3.14", features = ["formatting"], optional = true }
 paste = { version = "1.0.9", optional = true }
 bitvec = "1.0.1"
+twox-hash = "1.6.3"
```
Kixiron, line 50: I tend to prefer [`xxhash-rust`](https://docs.rs/xxhash-rust) since it utilizes simd and is a really solid implementation

Kixiron (approved):

ryzhyk:
> Can we merge this? I need the serde for floats part...

Let me apply @Kixiron 's suggestions.

PR #213, opened 2022-10-14
Simplified trait bounds.

ryzhyk:
Simplified trait bounds in `distinct.rs` and `window.rs` (missed them during recent refactoring).

@gz, with this you shouldn't need encode/decode bounds in these files in #124.


gz (approved):
lgtm will update my branch after this is merged

PR #218, opened 2022-10-25
GDELT demo with limited hashjoin

Kixiron:
* Implemented basic [GDELT](https://www.gdeltproject.org/) demo
* Implemented limited hashjoin operator (doesn't handle timestamps other than `()` right now)


ryzhyk:
Closing. Use #221  instead.

PR #220, opened 2022-10-27
Support for serialized inputs.

ryzhyk:
This PR starts building an infrastructure to consume data from/emit data to various sources/sinks.  The full pipeline will look smth like:

```
transport -> format -> deserialize -> [circuit] -> serialize -> format -> transport
```

where
 
1. **Transport** is any technology used to send/receive bytes,  e.g., Kafka/Kinesis/Google pub-sub/file/postgres
1. **Format** describes the way to package data records into byte streams, e.g., JSON, bincode, CSV, ...
1. **(de)serialize** - type-specific (de)serialization logic.

the idea is that all three can be implemented as independent libraries that can be combined in arbitrary ways at runtime. This PR focuses on the third component, specifically the deserialization side.  It wraps the input stream handles exposed by a circuit into deserializing handles that consume serialized data and deserialize it on the fly. The new API is object-safe and  can be connected to the appropriate format library at runtime without recompilation.  In practice this means that the pipeline shown above can be assembled at runtime by combining arbitrary transport and format libraries.


Cargo.toml
```diff
@@ -36,8 +36,9 @@ crossbeam-utils = "0.8.6"
 once_cell = "1.9.0"
 priority-queue = "1.2.1"
 hashbrown = "0.12.0"
-csv = { version = "1.1", optional = true }
+csv = { git = "https://github.com/ryzhyk/rust-csv.git", optional = true }
```
mbudiu-vmw, line 39: Are we reviewing this one too?

src/operator/input.rs
```diff
@@ -424,23 +424,26 @@ where
     type Value = InputHandle<T>;
 }
 
-/// Mailbox used to buffer data sent via an `InputHandle` to a worker thread.
+/// Mailbox that buffers data between the circuit and the outside world.
+/// It is use inside an `InputHandle` to store data sent to a worker
```
mbudiu-vmw, line 428: used

src/operator/input.rs
```diff
@@ -424,23 +424,26 @@ where
...
+/// thread and inside an `OutputHandle` to store data sent by a worke
```
mbudiu-vmw, line 429: worker

src/operator/output.rs
```diff
@@ -0,0 +1,311 @@
...
+use typedmap::TypedMapKey;
+
+impl<T> Stream<Circuit<()>, T>
+where
+    T: Clone + Send + 'static,
+{
+    /// Create an output handle that makes the contents of `self` available
+    /// outside the circuit.
+    ///
+    /// This API make result of the computation performed by the circuit
```
mbudiu-vmw, line 25: the result

src/operator/output.rs
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    fn mailbox(&self, worker: usize) -> &Mailbox<Option<T>> {
+        self.0.mailbox(worker)
+    }
+
+    /// Read the value produced by `worker` worker thread during the last
+    /// clock cycle.
+    ///
+    /// This method is invoked between two subsequent
```
mbudiu-vmw, line 161: subsequent -> consecutive

src/operator/output.rs
```diff
@@ -0,0 +1,311 @@
...
+    /// [`DBSPHandle::step`](`crate::DBSPHandle::step`)
+    /// calls to retrieve the value written to the stream during the last
+    /// clock cycle.  The first call is guaranteed to return a value
+    /// (since a synchronous circuit writes exactly one value to each
+    /// stream at every clock cycle).  It removes the value from the
+    /// mailbox, so subsequent calls will return `None`.
+    ///
+    /// Invoking this method in the middle of a clock cycle, i.e., during
```
mbudiu-vmw, line 169: Shouldn't this be rather be undefined?

src/operator/output.rs
```diff
@@ -0,0 +1,311 @@
...
+        for batch in batches {
+            spine.insert(batch);
+        }
+
+        spine.consolidate().unwrap_or_else(|| T::empty(()))
+    }
+}
+
+/// Sink operator that stores the contents of its input stream in
+/// the `OutputHandle`.
```
mbudiu-vmw, line 232: there is no outputhandle here. Which "the"?

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
+use crate::{algebra::ZRingValue, CollectionHandle, DBData, DBWeight, InputHandle, UpsertHandle};
+use erased_serde::{deserialize, Deserializer as ErasedDeserializer, Error as EError};
+use serde::Deserialize;
+
+const MAX_REUSABLE_CAPACITY: usize = 100_000;
+
+/// An input handle that deserializes values before pushing them to
```
mbudiu-vmw, line 7: maybe it should be called deser_input instead of deinput.

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+/// a stream.
+///
+/// A trait for a type that wraps around an [`InputHandle`] and
+/// pushes serialized data to an input stream.  The client passes a
+/// deserializer object that provides access to a serialized data
```
mbudiu-vmw, line 12: who is the client?

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+///
+/// This trait is object-safe, i.e., it can be converted to `dyn DeInputHandle`
+/// and invoked using dynamic dispatch.  This allows choosing the input
+/// format (encapsulated by the deserializer object) at runtime or even
+/// switching between different formats dynamically.  Object safety is
+/// achived by using the `erased_serde` crate that exposes an object-safe
+/// deserializer API ([`erased_serde::Deserializer`]), so that `DeInputHandle`
+/// methods don't need to be parameterized by the deserializer type.
+///
+/// Note: this trait is useful for feeding scalar values to the circuit.
```
mbudiu-vmw, line 26: the naming should reflect this then
Call the other one DeScalarHandleImpl

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+    /// methods support deletion by value, hence the serialized record must
+    /// match the value type of the stream.  Streams created with
+    /// [`Circuit::add_input_map`](`crate::Circuit::add_input_map`)
+    /// support deletion by key, so the serialized record must match the key
+    /// type of the stream.
+    ///
+    /// The record gets deserialized and pushed to the underlying input stream
+    /// handle as a delete update.
+    ///
+    /// Returns an error if deserialization fails, i.e., the serialized
```
mbudiu-vmw, line 152: what happens if the deletion itself cannot be executed, i.e., the key does not exist?

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+    /// to [`DBSPHandle::step`](`crate::DBSPHandle::step`).  `flush` can
+    /// be called multiple times between two subsequent `step`s.  Every
+    /// `flush` call adds new updates to the previously flushed updates.
+    ///
+    /// Updates queued after the last `flush` remain buffered in the handle
+    /// until the next `flush` or `clear_buffer` call or until the handle
+    /// is destroyed.
+    fn flush(&mut self);
+
+    /// Clear all buffered updates.
```
mbudiu-vmw, line 179: undo instead of clear?

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+/// returned by [`Circuit::add_input_zset`](`crate::Circuit::add_input_zset`).
+///
+/// The [`insert`](`Self::insert`) method of this handle deserializes value
+/// `v` type `V` and buffers a `(v, +1)` update for the underlying
+/// `CollectionHandle`.
+///
+/// The [`delete`](`Self::delete`) method of this handle deserializes value
+/// `v` type `V` and buffers a `(v, -1)` update for the underlying
+/// `CollectionHandle`.
+pub struct DeZSetHandle<K, R> {
```
mbudiu-vmw, line 203: no type constraints?

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+        let key = deserialize::<K>(deserializer)?;
+
+        self.updates.push((key, R::one()));
+        Ok(())
+    }
+
+    fn delete(&mut self, deserializer: &mut dyn ErasedDeserializer) -> Result<(), EError> {
+        let key = deserialize::<K>(deserializer)?;
+
+        self.updates.push((key, R::one().neg()));
```
mbudiu-vmw, line 238: can you just do this to a collection? I understand for a zset

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+        Ok(())
+    }
+
+    fn reserve(&mut self, reservation: usize) {
+        self.updates.reserve(reservation);
+    }
+
+    fn flush(&mut self) {
+        self.handle.append(&mut self.updates);
+        self.updates.shrink_to(MAX_REUSABLE_CAPACITY);
```
mbudiu-vmw, line 309: What is this MAX thing?

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+        // println!("csv:\n{}", csv_string);
+
+        // CSV reader iterates over CSV records.
+        let mut csv_reader = CsvReader::from_reader(csv_string.as_bytes());
+
+        for input in csv_reader.records() {
+            let record = input.unwrap();
+            // Wrap CSV record in erased deserializer.
+            let mut deserializer = string_record_deserializer(&record, None);
+            let mut deserializer = <dyn ErasedDeserializer>::erase(&mut deserializer);
```
mbudiu-vmw, line 584: this is odd, to create a new deserializer for each input record.

mbudiu-vmw (approved):

src/operator/deinput.rs
Kixiron: Clear is better, we're not "undoing" anything here

src/operator/deinput.rs
Kixiron: Trait bounds on raw structs should be avoided when possible

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+    pub fn new(handle: CollectionHandle<K, R>) -> Self {
+        Self {
+            updates: vec![],
+            handle,
+        }
+    }
+
+    fn clear(&mut self) {
+        if self.updates.capacity() > MAX_REUSABLE_CAPACITY {
+            self.updates = vec![];
```
Kixiron, line 218:
```suggestion
            self.updates = Vec::new();
```

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+
+impl<K, R> DeZSetHandle<K, R> {
+    pub fn new(handle: CollectionHandle<K, R>) -> Self {
+        Self {
+            updates: vec![],
+            handle,
+        }
+    }
+
+    fn clear(&mut self) {
```
Kixiron, line 216: `clear()` isn't the best name here since no clearing is done, maybe `.retain_buffers()` or something would be better

src/operator/deinput.rs
```diff
@@ -0,0 +1,703 @@
...
+/// `UpsertHandle`.
+pub struct DeSetHandle<K> {
+    updates: Vec<(K, bool)>,
+    handle: UpsertHandle<K, bool>,
+}
+
+impl<K> DeSetHandle<K> {
+    pub fn new(handle: UpsertHandle<K, bool>) -> Self {
+        Self {
+            updates: vec![],
```
Kixiron, line 279:
```suggestion
            updates: Vec::new(),
```

src/operator/deinput.rs
Kixiron: Additionally, the `.shrink_to(MAX_REUSABLE_CAPACITY)` strategy is probably the better one here

src/operator/input.rs
```diff
@@ -424,23 +424,26 @@ where
...
-    fn new() -> Self {
+    pub(super) fn new() -> Self {
         Self {
             value: Arc::new(Mutex::new(Default::default())),
         }
     }
 
-    fn take(&self) -> T {
+    pub(super) fn take(&self) -> T {
         take(self.value.lock().unwrap().deref_mut())
```
Kixiron, line 447:
```suggestion
        take(&mut *self.value.lock().unwrap())
```

src/operator/input.rs
```diff
@@ -451,7 +454,7 @@ where
         f(self.value.lock().unwrap().deref_mut());
```
Kixiron, line 454:
```suggestion
        f(&mut *self.value.lock().unwrap());
```

src/operator/output.rs
Kixiron, line 25:
```suggestion
    /// This API makes the result of the computation performed by the circuit
```

src/operator/output.rs
```diff
@@ -0,0 +1,311 @@
...
+    type Value = OutputHandle<T>;
+}
+
+struct OutputHandleInternal<T> {
+    mailbox: Vec<Mailbox<Option<T>>>,
+}
+
+impl<T> OutputHandleInternal<T> {
+    fn new(num_workers: usize) -> Self {
+        assert_ne!(num_workers, 0);
```
Kixiron, line 86: We could also take a `NonZeroUsize` as our input type

src/operator/output.rs
Kixiron: Unspecified at the very most, it's safe code so it shouldn't be able to cause undefined behavior

src/operator/output.rs
```diff
@@ -0,0 +1,311 @@
...
+    pub fn take_from_all(&self) -> Vec<T> {
+        let num_workers = self.0.mailbox.len();
+        let mut res = Vec::with_capacity(num_workers);
+
+        for worker in 0..num_workers {
+            if let Some(v) = self.take_from_worker(worker) {
+                res.push(v);
+            }
+        }
+        res
```
Kixiron, line 193:
```suggestion

        let mut result = Vec::with_capacity(num_workers);
        for worker in 0..num_workers {
            if let Some(v) = self.take_from_worker(worker) {
                result.push(v);
            }
        }

        result
```

Kixiron: Although honestly I kinda dislike the index loop we've got going on here, we could probably just get a slice/iterator of mailboxes to take from

src/operator/output.rs
Kixiron: The one created in `Output::new()`, I assume. Docs could be a little more clear here, I agree with that

src/operator/output.rs
```diff
@@ -0,0 +1,311 @@
...
+where
+    T: Clone + 'static,
+{
+    fn eval(&mut self, val: &T) {
+        self.mailbox.set(Some(val.clone()));
+    }
+
+    fn eval_owned(&mut self, val: T) {
+        self.mailbox.set(Some(val));
+    }
```
Kixiron, line 274: Doesn't this mean that any consumer can miss values? If I forget to receive at any given clock cycle and the circuit progresses to another clock cycle then I miss the previous value

src/operator/output.rs
ryzhyk: it's not undefined or unsafe, but the outcome depends on the relative ordering of operations.

src/operator/deinput.rs
ryzhyk: Makes sense.  TBH, I am not happy with the term "scalar".  This handle can be used with values of any type, it's just that it accepts the entire value at once vs the the handle that works record-by-record.

src/operator/deinput.rs
ryzhyk: Deletion never fails.  It either creates a value with a negative weight (if it's a regular input stream) or gets ignored (if it's an upsert handle).

src/operator/deinput.rs
ryzhyk: Updates haven't been applied yet, only buffered.  This is literally clearing the buffer.

src/operator/deinput.rs
ryzhyk: This is a Z-set handle, so this is well-defined.

src/operator/deinput.rs
ryzhyk: I'm ok with this change,  but just curious, what's the reasoning here?

src/operator/deinput.rs
ryzhyk: That's what I did initially, but then I realized that if the buffer is `>MAX_REUSABLE_CAPACITY`, it will get deallocated and a new one will be allocated instead, whereas all I want to do is reuse existing buffer if it's not too huge and free it otherwise.

src/operator/deinput.rs
ryzhyk: A deserializer usually just store a reference to the underlying data, so it's not an expensive abstraction if that's your concern.

Having said that, this may not be the best API design.  It's tricky to come up with an interface that allows switching between different serialized representations of the same type at runtime.  `erased_serde` makes this possible by encapsulating the knowledge of the format in the `ErasedDeserializer` trait.  I can't think of another way.  Let's see if this even works (i.e., can we have an end-to-end data pipeline going).

Cargo.toml
ryzhyk: Sorry, forgot to mention this in the PR.  All I did is exposed the deserializer type from that crate in the public API.  Unfortunately, the crate maintainer isn't keen to accept this change, so we'll have to live with a fork.

src/operator/deinput.rs
Kixiron: It shouldn't deallocate and reallocate, all it does is make a `realloc()` call if the capacity is larger

src/operator/deinput.rs
Kixiron: It's clearer and consistent with the rest of the codebase, `vec![]` generally means "we could put stuff here if we wanted to" but in this case we're just making an empty one so there's not really that connotation

src/operator/output.rs
```diff
@@ -0,0 +1,311 @@
...
+        output_handle
+    }
+}
+
+/// `TypedMapKey` entry used to share `OutputHandle` objects across workers in a
+/// runtime. The first worker to create the handle will store it in the map,
+/// subsequent workers will get a clone of the same handle.
+struct OutputId<T> {
+    id: usize,
+    _marker: PhantomData<T>,
```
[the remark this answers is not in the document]
ryzhyk: This makes it complain:
```
error[E0277]: `*const _` cannot be sent between threads safely
```

src/operator/output.rs
ryzhyk: Correct, it's a synchronous circuit, so there's a well-defined time when the output is available.  If you choose not to read it, it gets overwritten.

src/operator/deinput.rs
ryzhyk: `realloc` either deallocates and reallocates or leaves the allocation unmodified.  Most memory allocators don't have a way to give part of an allocated memory region back to the allocator.

PR #222, opened 2022-10-28
Run and save nexmark benchmark results as part of CI.

gz:


gz:
Windows runner is running out of disk-space:

```
   Compiling dbsp v0.1.0 (D:\a\database-stream-processor\database-stream-processor)
LLVM ERROR: IO failure on output stream: no space on device
error: could not compile `dbsp`
warning: build failed, waiting for other jobs to finish...
LLVM ERROR: IO failure on output stream: no space on device
error: could not compile `dbsp`
Error: The process 'C:\Users\runneradmin\.cargo\bin\cargo.exe' failed with exit code 101
```

scripts/ci.bash
```diff
@@ -0,0 +1,53 @@
...
+#
+set -ex
+
+# Clean-up old results
+rm -rf gh-pages
+NEXMARK_CSV_FILE='nexmark_results.csv'
+rm -f ${NEXMARK_CSV_FILE}
+
+# Run benchmarks
+cargo bench --bench nexmark --features with-nexmark -- --first-event-rate=1000000 --max-events=100000 --cpu-cores 8  --num-event-generators 6 --source-buffer-size 10000 --input-batch-size 40000 --csv
```
ryzhyk, line 13: This is without persistence, right?  In that case, we want `--first-event-rate=10000000` (10M).

.github/workflows/main.yml
```diff
@@ -38,11 +38,11 @@ jobs:
               os: [self-hosted, Linux, skylake-2x],
             }
           - { rust: nightly, target: x86_64-apple-darwin, os: macos-latest }
-          - {
-              rust: nightly,
-              target: x86_64-pc-windows-msvc,
-              os: windows-latest,
-            }
+          #- {
```
ryzhyk, line 41: Giving up on windows completely is a bit too radical.  Perhaps we could just build/test with minimal features?

scripts/ci.bash
gz: Yes, I'll increase it, left it low for now so it completes fast

.github/workflows/main.yml
gz: It runs out of disk-space during build -- I'll try to build with less features/dependencies.

ryzhyk: Even running with no features would be fairly useful, compared to dropping windows altogether.

scripts/ci.bash
gz: Fixed

.github/workflows/main.yml
gz: Disabling the CI cache for windows makes it work again with no need to limit features.

PR #228, opened 2022-11-08
Update contribution guidelines.

gz:
This clarifies some workflow changes to ensure that the benchmark framework is effective.
- fast-forward merging so we don't lose the mapping of commit ids to results
- use feature branches on vmware so we get comments for benchmark regressions


Kixiron:
Maybe we should skip benchmarking if the lockfile and code doesn't change

Issue #239, opened 2022-11-25
Documentation of the fraud query is incorrect


mbudiu-vmw:
The Rust code does not implement the SQL query exactly, in particular the LAG part of the query is not implemented.
Either the Rust or the SQL should be updated.

PR #244, opened 2022-11-29
Simplified rolling aggregate

ryzhyk:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,7 +59,7 @@ jobs:
           - { rust: stable, target: i686-pc-windows-msvc, os: windows-latest }
           - { rust: beta, target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
           - {
-              rust: nightly,
+              rust: nightly-2022-11-10,
               target: x86_64-unknown-linux-gnu,
               os: ubuntu-latest,
             }
@@ -189,7 +189,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: nightly-2022-11-10
           target: ${{ matrix.target }}
           components: rust-src
           default: true
@@ -237,7 +237,7 @@ jobs:
         with:
           profile: minimal
           # Run nightly clippy, which often finds additional issues.
-          toolchain: nightly
+          toolchain: nightly-2022-11-10
           components: clippy
           default: true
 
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           - { rust: stable, target: i686-pc-windows-msvc, os: windows-latest }
           - { rust: beta, target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
           - {
-              rust: nightly,
+              rust: nightly-2022-11-10,
```
mbudiu-vmw, line 62: why test on nightly?

benches/fraud.rs
```diff
@@ -195,11 +195,9 @@ struct Args {
 type Weight = i32;
 
 type EnrichedTransactions = OrdIndexedZSet<(F64, i64), (Transaction, Demographics), Weight>;
-type AverageSpendingPerWeek =
-    OrdPartitionedIndexedZSet<F64, i64, (F64, Option<Avg<F64, i32>>), Weight>;
-type AverageSpendingPerMonth =
-    OrdPartitionedIndexedZSet<F64, i64, (F64, Option<Avg<F64, i32>>), Weight>;
-type TransactionFrequency = OrdPartitionedIndexedZSet<F64, i64, (F64, Option<i32>), Weight>;
+type AverageSpendingPerWeek = OrdPartitionedIndexedZSet<F64, i64, Option<Avg<F64, i32>>, Weight>;
```
mbudiu-vmw, line 198: you didn't fix the SQL... #239

src/operator/time_series/rolling_aggregate.rs
```diff
@@ -391,23 +390,21 @@ where
...
                                     delta_cursor.key().clone(),
-                                    (
-                                        *input_range_cursor.key(),
-                                        (input_range_cursor.val().clone(), agg.clone()),
-                                    ),
+                                    (*input_range_cursor.key(), agg),
                                 ),
                                 HasOne::one(),
                             ));
+                            break;
```
mbudiu-vmw, line 407: I find this surprising

mbudiu-vmw (approved):

.github/workflows/main.yml
ryzhyk: to catch issues that we'll run into when nightly becomes stable.

ryzhyk: also, things like sanitizers only work on nightly

src/operator/time_series/rolling_aggregate.rs
ryzhyk: As soon as we encounter an input record with non-zero weight, we know that we need to compute the sliding window aggregate for the current timestamp.  We don't care about the actual value or the number of different values, so we can break out of the loop.

PR #245, opened 2022-12-02
Kafka output transport.

ryzhyk:
Output endpoint backed by Kafka producer.

### DIFF
--- a/adapters/src/controller/mod.rs
+++ b/adapters/src/controller/mod.rs
@@ -590,7 +590,16 @@ impl ControllerInner {
                 ControllerError::unknown_output_transport(&endpoint_config.transport.name)
             })?;
 
-        let endpoint = transport.new_endpoint(&endpoint_config.transport.config)?;
+        let endpoint_name_str = endpoint_name.to_string();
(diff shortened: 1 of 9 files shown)


adapters/src/controller/mod.rs
```diff
@@ -590,7 +590,16 @@ impl ControllerInner {
                 ControllerError::unknown_output_transport(&endpoint_config.transport.name)
             })?;
 
-        let endpoint = transport.new_endpoint(&endpoint_config.transport.config)?;
+        let endpoint_name_str = endpoint_name.to_string();
+        let self_weak = Arc::downgrade(self);
```
Kixiron, line 594: Why does this use a weak ref instead of a strong one?

Kixiron (approved):

adapters/src/controller/mod.rs
ryzhyk: To avoid the controller->endpoint->controller circular reference.  It might happen that the controller is gone while the endpoint is still running for a short time, in which case it's ok to drop the error notification.

mbudiu-vmw (approved):

PR #247, opened 2022-12-10
Minimal DBSP server

ryzhyk:
A server that allows managing a DBSP circuit via HTTP.  Only the
following commands are currently implemented:

- `/start` - start pipeline
- `/pause` - pause pipeline
- `/shutdown` - destroy the pipeline
- `index.html` - serve static `index.html` file.
- `static/filename` - serve files from the `static` directory (which
  will need to be distributed with the server).

TODOs:
- API to retrieve circuit configuration (input/output endpoints)
- API to retrieve endpoint status and statistics
- Support for grafana or similar dashboard.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ cached = { version = "0.38.0", optional = true }
 zip = "0.6.2"
 tar = "0.4.38"
 rand = "0.8.5"
-zstd = "0.10.0"
+zstd = "0.12.0"
 paste = "1.0.9"
 rstest = "0.15"
 cached = "0.38.0"
--- a/adapters/Cargo.toml
+++ b/adapters/Cargo.toml
@@ -5,8 +5,10 @@ edition = "2021"
 license = "MIT OR Apache-2.0"
 
 [features]
-default = ["with-kafka"]
+default = ["with-kafka", "server"]
 with-kafka = ["rdkafka"]
+server = ["actix-files", "actix-web", "with-kafka"]
+test-utils = ["size-of", "futures", "proptest", "proptest-derive"]
 
 [dependencies]
 num-traits = "0.2.15"
@@ -22,6 +24,13 @@ csv = { git = "https://github.com/ryzhyk/rust-csv.git" }
 bincode = { version = "2.0.0-rc.2", features = ["serde"] }
 # cmake-build is required on Windows.
 rdkafka = { version = "0.29.0", features = ["cmake-build"], optional = true }
+actix-files = { version = "0.6.2", optional = true }
+actix-web = { version = "4.2", optional = true }
+log = "0.4.17"
+size-of = { version = "0.1.2", features = ["time-std"], optional = true}
+futures = { version = "0.3.25", optional = true}
+proptest = { version = "1.0.0", optional = true} 
+proptest-derive = { version = "0.3.0", optional = true }
 
 [dev-dependencies]
 serde_json = "1.0.87"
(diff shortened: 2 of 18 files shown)


adapters/examples/server.rs
```diff
@@ -0,0 +1,91 @@
...
+            // appear in the output topic before sending more.
+            let data = generate_test_batches(5, 100)
+                .new_tree(&mut runner)
+                .unwrap()
+                .current();
+
+            producer.send_to_topic(&data, "server_example_input_topic");
+            buffer_consumer.wait_for_output_unordered(&data);
+            buffer_consumer.clear();
+            print!(".");
```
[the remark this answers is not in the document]
ryzhyk: Not over HTTP. I thought we agreed to use separate tools for log analysis.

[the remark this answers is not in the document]
mbudiu-vmw: nothing to do with http; you could just log the fact that a message was sent.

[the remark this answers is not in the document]
ryzhyk: Yeah, it does get logged, you should see it in the console when you run the example.

Cargo.toml
```diff
@@ -76,7 +76,7 @@ cached = { version = "0.38.0", optional = true }
 zip = "0.6.2"
 tar = "0.4.38"
 rand = "0.8.5"
-zstd = "0.10.0"
+zstd = "0.12"
```
Kixiron, line 79:
```suggestion
zstd = "0.12.0"
```

adapters/Cargo.toml
```diff
@@ -22,6 +23,13 @@ csv = { git = "https://github.com/ryzhyk/rust-csv.git" }
 bincode = { version = "2.0.0-rc.2", features = ["serde"] }
 # cmake-build is required on Windows.
 rdkafka = { version = "0.29.0", features = ["cmake-build"], optional = true }
+actix-files = "0.6.2"
+actix-web = "4.2"
```
Kixiron, line 27: These should be optional dependencies

adapters/src/server/mod.rs
```diff
@@ -0,0 +1,263 @@
+use crate::{Catalog, Controller, ControllerConfig, ControllerError};
```
Kixiron, line 1: This server should really be a separate crate from `adapters`

adapters/src/test/kafka.rs
```diff
@@ -0,0 +1,261 @@
...
+
+pub struct KafkaResources {
+    admin_client: AdminClient<DefaultClientContext>,
+    topics: Vec<String>,
+}
+
+/// An object that creates a couple of topics on startup and deletes them
+/// on drop.  Helps make sure that test runs don't leave garbage behind.
+impl KafkaResources {
+    pub fn create_topics(topics: &[(&str, i32)]) -> Self {
```
Kixiron, line 31: Most of this code is async, this should probably be an async function instead of internally blocking

adapters/Cargo.toml
ryzhyk: good catch

adapters/src/test/kafka.rs
ryzhyk: I use the sync version of the Kafka API.  There are a couple of APIs that don't have a sync version (presumably an omission), so I wrap them in `block_on`, but I don't want to propagate asynchronicity to the rest of the framework.

adapters/src/server/mod.rs
ryzhyk: That was my first instinct, but at the moment the server is a really thin layer on top of the controller API, really just an HTTP binding to the controller API.  I'll just feature gate it.

PR #248, opened 2022-12-12
Implemented specalized gather operator

Kixiron:
Implements a specialized `.gather()` operator with minimal synchronization overhead, utilizing a spsc oneshot channel for sending batches to the gathering thread instead of the more heavyweight `Exchange` operator

### DIFF
--- a/src/operator/communication/gather.rs
+++ b/src/operator/communication/gather.rs
@@ -1,0 +1,423 @@
+use crate::{
+    circuit::{
+        metadata::OperatorLocation,
+        operator_traits::{Operator, SinkOperator, SourceOperator},
+        GlobalNodeId, LocalStoreMarker, OwnershipPreference, Scope,
+    },
+    circuit_cache_key,
+    trace::{spine_fueled::Spine, Batch, Trace},
+    Circuit, Runtime, Stream,
+};
+use crossbeam::{
+    atomic::AtomicConsume,
+    channel::{Receiver, Sender},
+};
+use once_cell::sync::OnceCell;
+use std::{
+    borrow::Cow,
+    hash::{Hash, Hasher},
+    marker::PhantomData,
+    panic::Location,
+    ptr,
+    sync::{
+        atomic::{AtomicPtr, Ordering},
+        Arc,
+    },
+};
+use typedmap::TypedMapKey;
+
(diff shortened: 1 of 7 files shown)


src/operator/communication/gather.rs
```diff
@@ -0,0 +1,423 @@
...
+}
+
+/// A very specialized oneshot single producer, single consumer channel
+///
+/// # Safety
+///
+/// Requires external synchronization via circuit epochs, the [`Producer`]
+/// associated with a channel should only be pushed to once every clock cycle.
+///
+struct OneshotSpsc<T> {
```
ryzhyk, line 341: Can we replace mailboxes of type `Mutex<Option<T>>` in `exchange.rs` with this?

src/operator/communication/gather.rs
```diff
@@ -0,0 +1,423 @@
...
+    }
+}
+
+impl<T> SinkOperator<T> for GatherProducer<T>
+where
+    T: Clone + Send + 'static,
+{
+    fn eval(&mut self, input: &T) {
+        self.queue.push(input.clone());
+        self.notify();
```
ryzhyk, line 207: This will spuriously invoke the callback once per sender, which doesn't affect correctness, but will cause the scheduler to do redundant work.

ryzhyk (commented):
Seems good.  I guess I expected this to have an analogous internal design to `Exchange`, which uses state shared between producers and consumers.  One advantage of that is that the shared state can contain an atomic counter of the number of produced messages that can be used to determine the right time to wake up the scheduler.

src/operator/communication/gather.rs
Kixiron: Correct, I figure it's relatively cheaper than keeping track of how many values have been pushed, in most situations this will ony cause a handful of extra activations on the gathering thread

src/operator/communication/gather.rs
Kixiron: Yah, I think so

Kixiron:
@gz Do you know why the benchmark workflow didn't comment?

ryzhyk:
> @gz Do you know why the benchmark workflow didn't comment?

It only works for branches in this repo.  I no longer user my fork of the repo for PRs because of that.

ryzhyk (approved):

PR #253, opened 2022-12-15
Implemented Exchange over spsc oneshot channels

Kixiron:


Kixiron:
@ryzhyk What's your opinion on the notification callbacks, currently we store two for each worker but in every case both of them are identical, so it's really a waste for everything we do. We could just store one single callback for each worker and that'd be smaller, less wasteful & easier to reason about

PR #261, opened 2023-01-10
Deploy and run pipeline server demo as a pod in a kind cluster

lalithsuresh:


demo/kind-deploy/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+RUN git clone --branch pipeline_server http://www.github.com/lalithsuresh/database-stream-processor \
+    && git clone http://www.github.com/vmware/sql-to-dbsp-compiler
+
+RUN cd database-stream-processor && ~/.cargo/bin/cargo build
+
+RUN cd sql-to-dbsp-compiler/SQL-compiler && mvn -DskipTests package
+
+# Install rpk
+RUN curl -1sLf 'https://packages.vectorized.io/nzc4ZYQK3WRGd9sy/redpanda/cfg/setup/bash.deb.sh' | bash && \
```
ryzhyk, line 20: I would move this and the next command up, right before cloning the DBSP repos, to speed up container rebuilds.

ryzhyk: Also, we probably want to add container creation to the DBSP build process, which means that we want to use local copies of DBSP and SQL compiler rather than pulling from github (I am reorganizing the repos now so that the SQL compiler will probably become a git submodule of dbsp -- still thinking of the best way to do this, but point is, you should be able to count on matching versions of both repos being available locally).

demo/demo.sh
```diff
@@ -4,8 +4,10 @@ THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 ROOT="${THIS_DIR}/../"
 SERVER_DIR="${ROOT}/pipeline_server/"
 
+service postgresql restart
```
ryzhyk, line 7: Why do we need to restart Postgres?

ryzhyk (commented):
Most of this probably should not (at least eventually) be under the `demo` directory. This is more of a dev deployment that a user can run on their laptop in order to play with DBSP before deploying it in their data pipeline.  Eventually we'll have at least one or two other deployments (cloud-based and on-prem?).

demo/demo.sh
lalithsuresh: Postgres isn't started by default when running the container for the first time. We should probably should just do "service postgresql start" in the login script instead.

demo/demo.sh
```diff
@@ -4,8 +4,6 @@ THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 ROOT="${THIS_DIR}/../"
 SERVER_DIR="${ROOT}/pipeline_server/"
 
-service postgresql restart
-
 # Kill previous server instance if any.
 pkill -9 dbsp_pipeline_server
```
ryzhyk, line 8: Do you remember why this was needed.  If we indeed must restart postgres, we should `pkill` the server first, since it might have a live connection to postgres (no need to make any changes, I am in the process of cherry-picking this PR in a separate branch, so just trying to understand)

Issue #265, opened 2023-01-12
The signature of the Fold class seems to be incorrect


mbudiu-vmw:
In particular:
```
impl<V, T, R, A, S, O, SF, OF> Aggregator<V, T, R> for Fold<A, S, SF, OF>
where
    T: Timestamp,
    R: MonoidValue,
    A: Clone + 'static,
    SF: Fn(&mut A, &V, R) + Clone + 'static,
    OF: Fn(A) -> O + Clone + 'static,
    S: Semigroup<O> + Clone + 'static,
    O: DBData,
```

S is `Semigroup<O>`, whereas it should be `Semigroup<A>`.

ryzhyk:
Fixed via #268

PR #268, opened 2023-01-13
Add `finalize` step to the `Aggregator` trait.

ryzhyk:
A recent extension of the `Aggregator` trait requires aggregated values to form a semigroup, which is not the case for instance for the AVG aggregate.  The standard solution is to use an intermediate representation of the aggregate that forms a semigroup and only compute the actual value of the aggregate after all partitions have been combined.

To support such aggregates we extend the `Aggregator` trait with the `Accumulator` associated type returned by `aggregate`, which must implement `trait Semigroup`.  The
`Aggregator::Output` associated type represents the final output of the aggregator computed by applying the `finalize` method to the final value of the accumulator.

### DIFF
--- a/benches/fraud.rs
+++ b/benches/fraud.rs
@@ -195,8 +195,8 @@ struct Args {
 type Weight = i32;
 
 type EnrichedTransactions = OrdIndexedZSet<(F64, i64), (Transaction, Demographics), Weight>;
-type AverageSpendingPerWeek = OrdPartitionedIndexedZSet<F64, i64, Option<Avg<F64, i32>>, Weight>;
-type AverageSpendingPerMonth = OrdPartitionedIndexedZSet<F64, i64, Option<Avg<F64, i32>>, Weight>;
+type AverageSpendingPerWeek = OrdPartitionedIndexedZSet<F64, i64, Option<F64>, Weight>;
+type AverageSpendingPerMonth = OrdPartitionedIndexedZSet<F64, i64, Option<F64>, Weight>;
 type TransactionFrequency = OrdPartitionedIndexedZSet<F64, i64, Option<i32>, Weight>;
 
 struct FraudBenchmark {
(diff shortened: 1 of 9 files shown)


benches/fraud.rs
```diff
@@ -233,16 +233,13 @@ impl FraudBenchmark {
...
                 );
 
             // TODO: this should be returned directly by `partitioned_rolling_aggregate`
-            let avg_spend_pw_indexed = avg_spend_pw.map_index(|(cc_num, (ts, avg_amt))| {
-                (
-                    (*cc_num, *ts),
-                    avg_amt.as_ref().and_then(|avg| avg.compute_avg()),
-                )
-            });
+            let avg_spend_pw_indexed =
```
mbudiu-vmw, line 241: It looks like you can remove the TODO comment.

src/operator/aggregate/mod.rs
```diff
@@ -39,13 +39,42 @@ pub use min::Min;
...
+/// This design requires aggregate values to form a semigroup with a `+`
+/// operation such that `Agg(x + y) = Agg(x) + Agg(y)`, i.e., aggregating
+/// a union of two Z-sets must produce the same result as the sum of
+/// individual aggregates.  Not all aggregates have this property.  E.g.,
+/// the average value of a Z-set cannot be computed by combining the
+/// averages of its subsets.  We can get around this problem by representing
+/// average as `(sum, count)` tuple with point-wise `+` operator
+/// `(sum1, count1) + (sum2, count2) = (sum1+sum2, count1+count2)`.
+/// The final result is converted to an actual average value by dividing
+/// the first element of the tuple by the second.  This is a general technique
```
mbudiu-vmw, line 60: for some aggregators, such as exact median, it may need to simply keep the full set around.

src/operator/aggregate/mod.rs
```diff
@@ -75,9 +105,20 @@ pub trait Aggregator<K, T, R>: Clone + 'static {
...
-    fn aggregate<'s, C>(&self, cursor: &mut C) -> Option<Self::Output>
+    fn aggregate<'s, C>(&self, cursor: &mut C) -> Option<Self::Accumulator>
     where
         C: Cursor<'s, K, (), T, R>;
+
+    /// Compute the final value of the aggregate.
+    fn finalize(&self, accumulator: Self::Accumulator) -> Self::Output;
+
+    /// Applies `aggregate` to `cursor` followed by `finalize`.
+    fn aggregate_finalize<'s, C>(&self, cursor: &mut C) -> Option<Self::Output>
```
mbudiu-vmw, line 116: this name is misleading, it suggests that only the finalize method of the aggregator is applied.
it should be called "aggregate_and_finalize"

mbudiu-vmw (approved):

mbudiu-vmw:
This is exactly what I was hoping for.

mbudiu-vmw:
But how about #265?
Doesn't that need to change as well?

mbudiu-vmw:
Fixes #265 
I guess it does.

PR #276, opened 2023-01-17
Uifixes

### LINKED ISSUE #270: update -> compile sequence doesn't work in UI.
`Update code` -> Uploading new code for the project -> `Compile` sends a `Compile` command to the server with the old version (before the update).

### LINKED ISSUE #271: Use different icons for different HTTP services.
Can we have different icons for the main server and the controller?  This would make it easier to navigate multiple browser tabs.

### LINKED ISSUE #272: Compile -> Update code doesn't work
The `Update code` button doesn't work (nothing happens) after an unsuccessful compilation attempt (not sure if this also happens after a successful compilation).

Reproduction: `Update code` -> load new code for the project -> refresh -> `Compile` -> `Status` (compilation fails) -> `Update code`.

### LINKED ISSUE #273: In a pipeline with multiple inputs, all inputs are labeled `input0`
image

### LINKED ISSUE #274: Show error message when `New pipeline` fails
When the `New pipeline` action fails, the UI shows `Error received: 500`.  There should be a more detailed error message in the body of that reply.

### LINKED ISSUE #275: Poll for status changes
Can we have the `Status` page poll for changes to status, so we don't need to keep checking?

mbudiu-vmw:
Fixes #275 
Fixes #274 
Fixes #273 
Fixes #272 
Fixes #271 
Fixes #270 
Notice that I could not test all fixes since I can't start new pipelines or compile code with the current version.


ryzhyk:
Merged manually.

PR #278, opened 2023-01-18
Fix the aarch64 neon vectorized implementation of `trace::consolidation::fill_indices`

### LINKED ISSUE #277: Out-of-bounds access in `consolidate_paired_slices`
On https://github.com/vmware/database-stream-processor/commit/50e58aa34d66b3fbaa6f7481a90af2ae870f132d, tests segfault on the unchecked access for `idx1` at https://github.com/vmware/database-stream-processor/blob/50e58aa34d66b3fbaa6f7481a90af2ae870f132d/src/trace/consolidation/mod.rs#L185

I temporarily replaced the unchecked access with regular `Index`, and got errors like:
```
---- operator::distinct::test::distinct_trace_test_mt stdout ----
thread 'dbsp-worker-0' panicked at 'index out of bounds: the len is 3 but the index is 8070450532586684418', src/trace/consolidation/mod.rs:186:13
thread 'operator::distinct::test::distinct_trace_test_mt' panicked at 'called `Result::unwrap()` on an `Err` value: Any { .. }', src/operator/distinct.rs:696:25
```
Which looks like the indices are nonsensical; maybe underflow?

These are all the failures:
```
failures:
    operator::aggregate::test::proptest_aggregate_test_mt
    operator::aggregate::test::proptest_aggregate_test_st
    operator::distinct::test::distinct_trace_test_mt
    operator::join::test::join_trace_test
    operator::join::test::propagate_nested_test
    operator::recursive::test::reachability
    operator::recursive::test::reachability2
    trace::consolidation::tests::fill_indices
    trace::consolidation::tests::test_consolidate_paired_slices
```

I don't yet know why this doesn't happen on CI. One potential environment difference is that I'm on Apple Silicon / M1.

utaal:
Fixes #277.

`indices_ptr` is u32/u64-sized, despite the SIMD operations acting on two u32/u64 datums at a time; this is also the pointer type expected by `vst1q_u64`.

I have only tested this on my 64 bit machine, the 32 bit fix is my best guess.

* [x] Can `usize` ever be 32 bits in `aarch64` here https://github.com/vmware/database-stream-processor/blob/50e58aa34d66b3fbaa6f7481a90af2ae870f132d/src/trace/consolidation/fill_indices.rs#L217?
Apparently, yes: https://en.wikipedia.org/wiki/AArch64

### DIFF
--- a/src/trace/consolidation/fill_indices.rs
+++ b/src/trace/consolidation/fill_indices.rs
@@ -200,12 +200,9 @@ unsafe fn fill_indices_x86_avx2(original_length: usize, indices: &mut Vec<usize>
 #[cfg(target_arch = "aarch64")]
 #[target_feature(enable = "neon")]
 pub unsafe fn fill_indices_aarch64_neon(original_length: usize, indices: &mut Vec<usize>) {
-    use std::{
-        arch::aarch64::{
-            uint32x4_t, uint64x2_t, vaddq_u32, vaddq_u64, vdupq_n_u32, vdupq_n_u64, vst1q_u32,
-            vst1q_u64,
-        },
-        mem::transmute,
+    use std::arch::aarch64::{
+        uint32x4_t, uint64x2_t, vaddq_u32, vaddq_u64, vdupq_n_u32, vdupq_n_u64, vst1q_u32,
+        vst1q_u64,
     };
 
     if original_length == 0 {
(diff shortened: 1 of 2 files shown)


Kixiron (approved):
Thanks!

mbudiu-vmw:
If this works maybe you can remove the "draft" label and we can merge it.

PR #279, opened 2023-01-23
Pipeline manager

ryzhyk:
* HTTP server that manages cataloging, compilation, and execution of SQL programs.
* A simple web UI for it
* A demo script that shows SQL streaming with DBSP end-to-end.


pipeline_server/src/runner.rs
```diff
@@ -0,0 +1,479 @@
...
+                // Create Prometheus config file for the pipeline.
+                // The Prometheus server should pick up this file automatically.
+                self.create_prometheus_config(
+                    &project_descr.name,
+                    request.project_id,
+                    pipeline_id,
+                    port,
+                )
+                .await
+                // Don't abandon pipeline, just log the error.
```
mbudiu-vmw, line 217: this comment should be one line lower

pipeline_server/src/runner.rs
```diff
@@ -0,0 +1,479 @@
...
+        };
+        let metadata_file_path = self.config.metadata_file_path(pipeline_id);
+        fs::write(
+            &metadata_file_path,
+            serde_json::to_string(&metadata).unwrap(),
+        )
+        .await?;
+
+        let log_file_path = self.config.log_file_path(pipeline_id);
+        let log_file = File::create(&log_file_path).await?;
```
mbudiu-vmw, line 267: do you need to?

pipeline_server/src/runner.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    }
+    */
+
+    async fn log_suffix_inner(log_file_path: &Path) -> AnyResult<String> {
+        let mut buf = Vec::with_capacity(10000);
+
+        let mut file = File::open(log_file_path).await?;
+
+        Pin::new(&mut file).start_seek(SeekFrom::End(-LOG_SUFFIX_LEN))?;
```
mbudiu-vmw, line 396: in lines?

pipeline_server/src/runner.rs
```diff
@@ -0,0 +1,479 @@
...
+            return Ok(HttpResponse::Ok().body("pipeline already killed"));
+        };
+
+        let url = format!("http://localhost:{port}/kill");
+        let response = match reqwest::get(&url).await {
+            Ok(response) => response,
+            Err(_) => {
+                // We failed to reach the pipeline, which likely means
+                // that it crashed or was killed manually by the user.
+                return Ok(HttpResponse::Ok().body(format!("pipeline at '{url}' already killed")));
```
mbudiu-vmw, line 443: will the db find out?

pipeline_server/src/main.rs
```diff
@@ -0,0 +1,873 @@
...
+//!   attached to it.  The client can add/remove/modify/compile a project.
+//!   Compilation includes running the SQL-to-DBSP compiler followed by the Rust
+//!   compiler.
+//!
+//! * Configuration.  A project can have multiple configurations associated with
+//!   it.  A configuration specifies input and output streams as a YAML file
+//!   that is deserialized into a `ControllerConfig` object.  Similar to
+//!   projects, one can add/remove/modify configs.
+//!
+//! * Pipeline.  A pipeline is a running instance of a compiled project.  One
```
mbudiu-vmw, line 16: based on one of the configs

pipeline_server/src/main.rs
```diff
@@ -0,0 +1,873 @@
...
+//!   can start multiple pipelines for a project with the same or different
+//!   configs.
+//!
+//! The API is currently single-tenant: there is no concept of users or
+//! permissions.  Multi-tenancy can be implemented by creating a server
+//! instance per tenant, which enables better separation of concerns,
+//! resource isolation and fault isolation compared to buiding multitenancy
+//! into the server.
```
mbudiu-vmw, line 24: what if there's very little traffic to the pipeline?

pipeline_server/src/main.rs
```diff
@@ -0,0 +1,873 @@
...
+//! database, this means that the last version of the project observed by the
+//! user is outdated, so the request is rejected.  Note that we don't store old
+//! versions, only the latest one.
+
+// TODOs:
+// * Tests.
+// * Generate a proper JSON API + docs using, e.g., OpenAPI.
+// * Support multi-node DBSP deployments (the current architecture assumes that
+//   pipelines and the Prometheus server run on the same host as this server).
+// * Proper UI.
```
mbudiu-vmw, line 60: this todo does not belong to this file

pipeline_server/src/main.rs
```diff
@@ -0,0 +1,873 @@
...
+include!(concat!(env!("OUT_DIR"), "/generated.rs"));
+
+fn build_app<T>(app: App<T>, state: WebData<ServerState>) -> App<T>
+where
+    T: ServiceFactory<ServiceRequest, Config = (), Error = ActixError, InitError = ()>,
+{
+    // Creates a dictionary of static files indexed by file name.
+    let generated = generate();
+
+    // Extrac the contents of `index.html`, so we can serve it on the
```
mbudiu-vmw, line 188: typo

pipeline_server/src/main.rs
```diff
@@ -0,0 +1,873 @@
...
+}
+
+/// Enumerate the project database.
+///
+/// Returns an array of [project descriptors](`db::ProjectDescr`).
+#[get("/list_projects")]
+async fn list_projects(state: WebData<ServerState>) -> impl Responder {
+    state
+        .db
+        .lock()
```
mbudiu-vmw, line 259: when does unlock happen?

pipeline_server/src/main.rs
```diff
@@ -0,0 +1,873 @@
...
+/// * `NOT_FOUND` - `project_id` does not exist in the database.
+#[post("/delete_project")]
+async fn delete_project(
+    state: WebData<ServerState>,
+    request: web::Json<DeleteProjectRequest>,
+) -> impl Responder {
+    let db = state.db.lock().await;
+
+    match db.list_project_pipelines(request.project_id).await {
+        Ok(pipelines) => {
```
mbudiu-vmw, line 555: I am not sure that the logic of this should be in the web server. 
the web server is just an API on top of some abstract controller.

pipeline_server/src/main.rs
```diff
@@ -0,0 +1,873 @@
...
+/// Create a new project configuration.
+///
+/// # HTTP errors
+///
+/// * `NOT_FOUND` - `project_id` does not exist in the database.
+#[post("/new_config")]
+async fn new_config(
+    state: WebData<ServerState>,
+    request: web::Json<NewConfigRequest>,
+) -> impl Responder {
```
mbudiu-vmw, line 599: I wonder whether this code could be simplified.
A controller can have a set of apis with input/output data structures.
the web server just sends the data to the controller and serializes back the data structure or the error.
same structure for all requests as long as the data structures implement some trait.

mbudiu-vmw: And the fact that the state is in a DB should also be abstracted by the controller.

mbudiu-vmw (changes requested):

adapters/src/controller/mod.rs
```diff
@@ -232,12 +243,32 @@ impl Controller {
         let min_batch_size_records = controller.status.global_config.min_batch_size_records;
 
         loop {
+            let dump_profile = controller
+                .dump_profile_request
+                .swap(false, Ordering::AcqRel);
+            if dump_profile {
```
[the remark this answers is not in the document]
ryzhyk: Yes! Not sure about showing the profile, but can we add a `Dump profile` button just to invoke this endpoint.  The user can then look at the profile in the file system.

adapters/src/server/mod.rs
```diff
@@ -103,6 +257,40 @@ async fn status(state: WebData<ServerState>) -> impl Responder {
...
+        .content_type(mime::APPLICATION_JSON)
+        .body(state.metadata.clone())
+}
+
+#[get("/dump_profile")]
+async fn dump_profile(state: WebData<ServerState>) -> impl Responder {
+    match &*state.controller.lock().unwrap() {
+        Some(controller) => {
+            controller.dump_profile();
+            HttpResponse::Ok().body("Profile dump initiated")
```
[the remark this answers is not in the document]
ryzhyk: initiated.

demo/project_demo03-GreenTrip/prepare.sh
```diff
@@ -0,0 +1,20 @@
...
+rpk topic delete green_trip_demo_large_input
+rpk topic delete green_trip_demo_large_output
+
+rpk topic create green_trip_demo_large_input -c retention.ms=-1 -c retention.bytes=-1
+rpk topic create green_trip_demo_large_output
+
+# Download the data file.
+if [ ! -f "${THIS_DIR}"/green_tripdata.csv ]
+then
+    gdown 14cKfJjwhsVPosshmSP7MBrolsTJfz9Xh --output "${THIS_DIR}"/green_tripdata.csv
```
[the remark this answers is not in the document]
ryzhyk: Same here. Didn't find another way to download from Google docs via command line.

pipeline_server/src/config.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// writes Prometheus config files for all pipelines.
+    ///
+    /// The default is `false`.
+    #[serde(default)]
+    pub with_prometheus: bool,
+
+    /// Compile pipelines in debug mode.
+    ///
+    /// The default is `false`.
+    #[serde(default)]
```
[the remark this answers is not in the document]
ryzhyk: I don't think so. In a workspace build, profiles defined inside individual crates are ignored.

pipeline_server/src/db.rs
```diff
@@ -0,0 +1,684 @@
...
+    /// Note: doesn't check that the project exists.
+    async fn set_project_status(
+        &self,
+        project_id: ProjectId,
+        status: ProjectStatus,
+    ) -> AnyResult<()> {
+        let (status, error) = status.to_columns();
+
+        self.dbclient
+            .execute(
```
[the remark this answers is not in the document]
ryzhyk: It will bubble up to the client as Internal Server Error.

pipeline_server/src/db.rs
```diff
@@ -0,0 +1,684 @@
...
+            return Ok(());
+        }
+
+        self.set_project_status(project_id, ProjectStatus::Pending)
+            .await?;
+
+        Ok(())
+    }
+
+    /// Cancel compilation request.
```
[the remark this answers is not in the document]
ryzhyk: Marks it in the DB.  The compiler polls the DB and will kill the compilation job within a second. From the user's perspective the job is cancelled after its status has been changed in the DB.

pipeline_server/src/db.rs
```diff
@@ -0,0 +1,684 @@
...
+
+    /// Retrieves the first pending project from the queue.
+    ///
+    /// Returns a pending project with the most recent `status_since` or `None`
+    /// if there are no pending projects in the DB.
+    pub(crate) async fn next_job(&self) -> AnyResult<Option<(ProjectId, Version)>> {
+        // Find the oldest pending project.
+        let rows = self
+            .dbclient
+            .query("SELECT id, version FROM project WHERE status = 'pending' AND status_since = (SELECT min(status_since) FROM project WHERE status = 'pending')", &[])
```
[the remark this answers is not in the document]
ryzhyk: Doesn't `min` guarantee at most one output record?

pipeline_server/src/runner.rs
```diff
@@ -0,0 +1,479 @@
...
+/// "Started HTTP server on port XXXXX" or "Failed to create server
+/// [detailed error message]".  In the former case, the port number is
+/// recorded in the database.  In the latter case, the error message is
+/// returned to the client.
+///
+/// # Killing a pipeline
+///
+/// To stop the pipeline, the runner sends a `/kill` HTTP request to the
+/// pipeline.  This request is asynchronous: the pipeline may continue running
+/// for a few seconds after the request succeeds.
```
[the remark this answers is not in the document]
ryzhyk: The server received the kill request and will shutdown within few seconds.

pipeline_server/src/runner.rs
ryzhyk: yeah, the `stderr` method takes a file handle, not just filename.

pipeline_server/src/main.rs
ryzhyk: This is crate-level documentation, and the UI is part of the crate.

pipeline_server/src/main.rs
ryzhyk: Then it shouldn't use a lot of resources.

pipeline_server/src/main.rs
ryzhyk: when the lock goes out of scope, in this case end of this statement.

pipeline_server/src/runner.rs
ryzhyk: Bytes actually, which is a good catch, because I myself got confused and wrote "lines" in the comment few lines below.

It would be better to seek N lines from the end, but I can't think of a simple way to do this.

pipeline_server/src/runner.rs
ryzhyk: good point

pipeline_server/src/main.rs
ryzhyk: I think the planned switch to OpenAPI will take care of this.

Kixiron (approved):

pipeline_manager/src/compiler.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    async fn compiler_task(config: ManagerConfig, db: Arc<Mutex<ProjectDB>>) -> AnyResult<()> {
+        Self::do_compiler_task(config, db).await.map_err(|e| {
+            error!("compiler task failed; error: '{e}'");
+            e
+        })
+    }
+
+    async fn do_compiler_task(
```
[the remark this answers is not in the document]
ryzhyk: The main complication is that we must periodically check for cancellation, which requires polling the database.  I don't know how to do this without a `select` in a loop.

PR #281, opened 2023-01-25
Kind deploy

ryzhyk:


deploy/docker.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -e
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+DEFAULT_SQL_COMPILER_PATH="${THIS_DIR}/../../sql-to-dbsp-compiler"
+SQL_COMPILER="${1:-$DEFAULT_SQL_COMPILER_PATH}"
+DBSP_PATH="${THIS_DIR}/../"
+
+(cd ${DBSP_PATH} && (git ls-files | tar -cvf ${THIS_DIR}/dbsp_files.tar -T -))
+(cd ${SQL_COMPILER} && (git ls-files | tar -cvf ${THIS_DIR}/sql_compiler_files.tar -T -))
```
lalithsuresh, line 10: This is still using git ls-files as the previous version, so how it is not sending the entire repository tree as before?

deploy/Dockerfile
```diff
@@ -19,10 +22,25 @@ RUN sed -i "s/scram-sha-256/trust/g" /etc/postgresql/14/main/pg_hba.conf \
...
+
 ADD ./dbsp_files.tar /database-stream-processor
 ADD ./sql_compiler_files.tar /sql-to-dbsp-compiler
 
-RUN cd /database-stream-processor && ~/.cargo/bin/cargo build --release
+RUN cd /database-stream-processor/pipeline_manager && ~/.cargo/bin/cargo build --release
 RUN cd /sql-to-dbsp-compiler/SQL-compiler && mvn -DskipTests package
 
+# Provision Prometheus data source + DBSP dashboard in Grafana.
+RUN cp /database-stream-processor/deploy/prometheus_datasource.yaml /etc/grafana/provisioning/datasources/
```
lalithsuresh, line 42: Every RUN command creates another docker layer. For these three, it's fine to make them a single set of chained commands in one RUN.

deploy/docker.sh
ryzhyk: I don't have a problem with adding all files under source control to the container image.  What I didn't like is that running `docker build` in the root of the repo caused docker to package and send to the demon all files in the project directory, including tens of GB or rust compiler artifacts.  I know this can be avoided by adding `.dockerignore` files, but that's another thing to maintain.  So I tried to solve the problem by running `docker build` in the `deploy` directory.  This is still not perfect, because now a bunch of `kind` working files get transferred, but it's better than sending everything.

deploy/Dockerfile
ryzhyk: thanks, fixed

lalithsuresh:
@ryzhyk can we merge this?

ryzhyk:
I just rebased.  Let's wait for CI and merge if there are no regressions.

Issue #229, opened 2022-11-09
nexmark benchmark have unstable performance


gz:
here are the worst offenders (stdev > 2) using 10 runs on the CI machine same invocation as the ci.bash script:

|  query                 |   ('elapsed', 'mean') |   ('elapsed', 'std') |   ('elapsed', 'min') |   ('elapsed', 'max') |
|:------------------|----------------------:|---------------------:|---------------------:|---------------------:|
| 7  |               29.5857 |              5.72089 |              24.0367 |              38.6235 |
| 13 |               35.6026 |              6.49526 |              28.4508 |              47.6523 |
| 18 |              125.103  |              2.85732 |             119.044  |             127.782  |
| 19 |              108.673  |              4.28322 |             102.351  |             118.69   |

gz:
1 core, 4 generator, things are fine:
```
|   name |   ('elapsed', 'mean') |   ('elapsed', 'std') |   ('elapsed', 'min') |   ('elapsed', 'max') |
|-------:|----------------------:|---------------------:|---------------------:|---------------------:|
|      0 |               7.70756 |            0.0121255 |              7.69477 |              7.73248 |
|      1 |              11.2202  |            0.0302091 |             11.1786  |             11.2689  |
|      2 |               7.80124 |            0.0169123 |              7.78096 |              7.82983 |
|      3 |               8.28608 |            0.0256119 |              8.2513  |              8.32688 |
|      4 |              13.9969  |            0.0648266 |             13.9327  |             14.1596  |
|      5 |               8.15482 |            0.014006  |              8.1358  |              8.18143 |
|      6 |              15.1791  |            0.0764544 |             14.987   |             15.2693  |
|      7 |              18.2843  |            0.0545263 |             18.1988  |             18.3711  |
|      8 |               8.3394  |            0.0280169 |              8.31073 |              8.39335 |
|      9 |              57.2127  |            0.185577  |             56.9653  |             57.5052  |
|     12 |              10.9224  |            0.028771  |             10.8843  |             10.993   |
|     13 |              21.2305  |            0.154159  |             20.9139  |             21.3537  |
|     14 |              11.36    |            0.016742  |             11.3369  |             11.3922  |
|     15 |              16.7001  |            0.0136383 |             16.6682  |             16.7181  |
|     16 |             119.59    |            0.221919  |            119.163   |            119.981   |
|     17 |              26.7408  |            0.0783941 |             26.6099  |             26.867   |
[...]

PR #282, opened 2023-01-27
Add a simple benchmark of the nexmark generator

utaal:
to help diagnose the unstable performance in https://github.com/vmware/database-stream-processor/issues/229


Kixiron:
I think that criterion benchmarks would be a good fit for this, that way we'd get much more scientific results for it

PR #283, opened 2023-01-31
Dataflow JIT

Kixiron:


crates/dataflow-jit/src/codegen/mod.rs
```diff
@@ -1399,6 +1474,79 @@ impl<'a> CodegenCtx<'a> {
         self.add_expr(expr_id, value, value_ty, None);
     }
 
+    fn cast(&mut self, expr_id: ExprId, cast: &Cast, builder: &mut FunctionBuilder<'_>) {
```
mbudiu-vmw, line 1477: notice that in the SQL compiler all casts are implemented manually as part of the runtime sql library. The numeric ones are relatively homogeneous, but there are some really weird ones in SQL. If we could support the sql library you would not need to implement casts. The generated code from the SQL compiler invokes these functions, which are named in some homogeneous way. Please see casts.rs

crates/dataflow-jit/src/ir/expr.rs
```diff
@@ -0,0 +1,483 @@
...
+
+    pub const fn value(&self) -> &RValue {
+        &self.value
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct IsNull {
+    value: ExprId,
+    row: usize,
```
ryzhyk, line 296: what is this field?

crates/dataflow-jit/src/ir/expr.rs
```diff
@@ -0,0 +1,483 @@
...
+    pub const fn row(&self) -> usize {
+        self.row
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct SetNull {
+    target: ExprId,
+    row: usize,
+    is_null: RValue,
```
ryzhyk, line 317: and this one?

crates/dataflow-jit/src/ir/types.rs
```diff
@@ -0,0 +1,393 @@
...
+    codegen::NativeType,
+    ir::{function::InputFlags, LayoutId, RowLayoutCache},
+};
+use bitvec::vec::BitVec;
+use std::fmt::{self, Debug, Display, Write};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ColumnType {
+    Bool,
+    U16,
```
ryzhyk, line 11: How come constants start with 32 bits but there are 16-bit types?

crates/dataflow-jit/src/ir/types.rs
```diff
@@ -0,0 +1,393 @@
...
+    I16,
+    U32,
+    I32,
+    U64,
+    I64,
+    F32,
+    F64,
+    /// Represents the days since Jan 1 1970 as an `i32`
+    Date,
```
ryzhyk, line 20: It seems odd to have `Date` and `Timestamp` as primitives in a low-level IR. Is there a reason they cannot be represented with integers?

crates/dataflow-jit/src/ir/types.rs
```diff
@@ -0,0 +1,393 @@
...
+}
+
+impl Display for ColumnType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.to_str())
+    }
+}
+
+pub struct RowLayoutBuilder {
+    rows: Vec<ColumnType>,
```
ryzhyk, line 172: not `columns`?

ryzhyk (commented):
Just started reviewing. Two big asks so far:
- I really prefer not to combine meaningful changes and moving stuff around (into the `crates` folder) in the same PR. Any chance we could separate them?
- Comments would really help (reading the `ir` module so far). It will probably all become obvious once I've got a hand of what's going on, but right now I am struggling to understand some of the structs and fields.

crates/dataflow-jit/src/codegen/mod.rs
Kixiron: It'd be a lot better if the SQL compiler emitted a cast instruction, otherwise we either pay big costs at every cast operation or I have to manually write a bunch of `if function.name == "cast_to_i64_i32"`

crates/dataflow-jit/src/ir/expr.rs
```diff
@@ -0,0 +1,483 @@
...
+        self.truthy
+    }
+
+    pub const fn falsy(&self) -> BlockId {
+        self.falsy
+    }
+}
+
+#[derive(Debug, Clone, From, PartialEq)]
+pub enum RValue {
```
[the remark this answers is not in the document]
Kixiron: An rvalue ("right value", name stems from historical stuff) is either an expression or an immediate constant value, it's useful in constant propagation and such

crates/dataflow-jit/src/ir/types.rs
Kixiron: Just because I haven't fully built all of that out, hasn't been necessary yet

crates/dataflow-jit/src/ir/types.rs
Kixiron: Needs to be renamed, it's a holdover from older code

crates/dataflow-jit/src/ir/types.rs
Kixiron: They totally could be, this just somewhat simplifies handling since it's well-known to the compiler

crates/dataflow-jit/src/ir/expr.rs
```diff
@@ -0,0 +1,483 @@
...
+        Self { target }
+    }
+
+    pub const fn target(&self) -> BlockId {
+        self.target
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Branch {
```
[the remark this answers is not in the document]
Kixiron: Not really, branch is the more normal name for block terminators like this

crates/dataflow-jit/src/ir/expr.rs
Kixiron: It should be `column`, it's the column that's being checked for null-ness

crates/dataflow-jit/src/ir/expr.rs
Kixiron: `row` is the same as the last one, `is_null` is the boolean value that's stored as the given column's null flag

crates/dataflow-jit/src/codegen/mod.rs
mbudiu-vmw: This has been fixed in https://github.com/vmware/sql-to-dbsp-compiler/pull/105

ryzhyk (changes requested):
This change breaks the demo.  I can fix it up once you've addressed the other issues.

Also, please consolidate the PR into a smaller number of commits.

crates/adapters/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 license = "MIT OR Apache-2.0"
 
 [features]
-default = ["with-kafka", "server"]
```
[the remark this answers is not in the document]
Kixiron: I didn't currently have the kafka stuff installed on my system, I'll revert that

crates/dataflow-jit/src/ir/node.rs
```diff
@@ -0,0 +1,1236 @@
...
+pub trait DataflowNode {
+    fn inputs(&self, inputs: &mut Vec<NodeId>);
+
+    fn output_kind(&self, inputs: &[Stream]) -> Option<StreamKind>;
+
+    fn output_stream(&self, inputs: &[Stream]) -> Option<Stream>;
+
+    fn signature(&self, inputs: &[Stream], layout_cache: &RowLayoutCache) -> Signature;
+
+    fn validate(&self, inputs: &[Stream], layout_cache: &RowLayoutCache);
```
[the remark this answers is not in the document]
Kixiron: Not yet, the validation infrastructure isn't fully built out

crates/dataflow-jit/src/ir/function.rs
```diff
@@ -0,0 +1,696 @@
...
+
+    /// Returns `true` if the parameter is only a input and not an output
+    pub const fn is_readonly(&self) -> bool {
+        self.is_input() && !self.is_output()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Function {
+    args: Vec<(LayoutId, ExprId, InputFlags)>,
```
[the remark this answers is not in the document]
Kixiron: It's the id of the argument

crates/dataflow-jit/src/ir/expr.rs
```diff
@@ -0,0 +1,692 @@
...
+}
+
+/// An rvalue (right value) is either a reference to another expression or an
+/// immediate constant
+#[derive(Debug, Clone, From, PartialEq)]
+pub enum RValue {
+    /// An expression
+    Expr(ExprId),
+    /// An immediate constant value
+    Imm(Constant),
```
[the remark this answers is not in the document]
Kixiron: Because it's not indirect, `Expr` only contains an `ExprId`. It's a fairly minor difference that may disappear later on

Kixiron:
Is the nexmark noise issue still present @utaal? I'm noticing substantial regressions from this PR which shouldn't be possible, it doesn't actually change anything related to those benchmarks or anything underlying them. I'm also noticing the regressions are really similar to those in #300 which is very strange since that PR actually does modify fundamental behavior within the library

ryzhyk:
We haven't done anything about the nexmark noise yet.

crates/dbsp/src/operator/filter_map.rs
```diff
@@ -123,7 +123,7 @@ pub trait FilterMap<C> {
     /// IntoIterator`, e.g., `Option<>` or `Vec<>`.
     fn flat_map<F, I>(&self, func: F) -> Stream<C, OrdZSet<I::Item, Self::R>>
     where
-        F: Fn(Self::ItemRef<'_>) -> I + 'static,
+        F: FnMut(Self::ItemRef<'_>) -> I + 'static,
```
ryzhyk, line 126: what's the use case for this?

crates/dbsp/src/trace/layers/erased/mod.rs
```diff
@@ -282,9 +289,12 @@ impl ErasedLayer {
             diffs.set_len(self.diffs.len());
         }
 
+        // TODO: We can eliminate elements from `0..lower_bound` when creating the
```
ryzhyk, line 292: Nope, in general we can't, as this will screw up the offsets in the parent layer.

crates/dbsp/src/trace/layers/ordered/mod.rs
```diff
@@ -126,6 +125,8 @@ where
     L: Trie + NegByRef,
     O: OrdOffset,
 {
+    // TODO: We can eliminate elements from `0..lower_bound` when creating the
```
ryzhyk, line 128: likewise

crates/dbsp/src/trace/layers/ordered/mod.rs
```diff
@@ -146,6 +147,8 @@ where
 {
     type Output = Self;
 
+    // TODO: We can eliminate elements from `0..lower_bound` when creating the
```
ryzhyk, line 150: same here

crates/dbsp/src/operator/filter_map.rs
Kixiron: It allows buffer reuse, the jit generates flat_map calls that reuse vectors

PR #286, opened 2023-02-03
Add 'orgchart' example similar to Differential Dataflow first example.

blp:
This is my first try at using DBSP. I'm sure stuff is wrong.

The most puzzling thing here for me is in print_output(). It seems
like there should be an easier way to iterate through the output
than two levels of key/value advancement. It also seems odd that
cursor.weight() mutates the cursor.

Signed-off-by: Ben Pfaff <bpfaff@vmware.com>

### DIFF
--- a/examples/orgchart.rs
+++ b/examples/orgchart.rs
@@ -7,6 +7,7 @@
 //! also include the employee's second-level manager.
 
 use anyhow::Result;
+use bincode::{Decode, Encode};
 use clap::Parser;
 use dbsp::{
     operator::FilterMap,
@@ -21,7 +22,7 @@ type EmployeeID = usize;
 /// Indicates that `manager` is the immediate manager of `employee`.
 ///
 /// If `manager == employee` then `manager` is the CEO.
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, SizeOf)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, SizeOf, Encode, Decode)]
 struct Manages {
     manager: EmployeeID,
     employee: EmployeeID,
@@ -29,7 +30,7 @@ struct Manages {
 
 /// Indicates that `manager` is the immediate manager of `employee` and that
 /// `grandmanager` is the immedate manager of `manager`.
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, SizeOf)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, SizeOf, Encode, Decode)]
 struct SkipLevel {
     grandmanager: EmployeeID,
     manager: EmployeeID,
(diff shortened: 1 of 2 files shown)


examples/orgchart.rs
```diff
@@ -0,0 +1,141 @@
...
+struct SkipLevel {
+    grandmanager: EmployeeID,
+    manager: EmployeeID,
+    employee: EmployeeID,
+}
+
+type Weight = i32;
+type SkipLevels = OrdIndexedZSet<EmployeeID, SkipLevel, Weight>;
+
+fn print_output(output: &OutputHandle<OrdIndexedZSet<usize, SkipLevel, Weight>>) {
```
ryzhyk, line 42: `usize` -> `EmployeeID`?

examples/orgchart.rs
```diff
@@ -0,0 +1,141 @@
...
+        let manages_by_manager = manages.map_index(|m| (m.manager, m.clone()));
+        let manages_by_employee = manages.map_index(|m| (m.employee, m.clone()));
+
+        // If Manages { manager, employee: common } and Manages { manager: common,
+        // employee } then SkipLevel { grandmanager: manager, manager: common,
+        // employee }.
+        let skiplevels: Stream<_, SkipLevels> = manages_by_employee
+            .join_index::<(), _, _, _, _, _>(&manages_by_manager, |common, m1, m2| {
+                Some((
+                    *common,
```
ryzhyk, line 87: I'm not sure you need to index output by `common`.  Could you just output `OrdZSet<SkipLevel, Weight>` instead of `OrdIndexedZSet<usize, SkipLevel, Weight>`?

ryzhyk (commented):
Looks great, thanks!  Not sure what happened in CI.  Will have a look.

ryzhyk:
Yes, `print_output` should be factored into a reusable function.  You should be able to use `.to_string()`, to format batches as strings using the `Display` trait, but the output isn't very nice.  We probably want  an `IndexedZSet<K:Display, V:Display, R:Display>::display()` method with default implementation similar to `print_output`.

Agreed about `weight` taking `mut` cursor being odd.  It has to do with internal implementation details of some cursor types. I tried to fix it in the past, but don't remember what the showstopper was.

examples/orgchart.rs
blp: Fixed with my push just now.

examples/orgchart.rs
blp: Oh, thanks for pointing that out. I thought that smelled funny but I didn't understand the difference between `join` and `join_index`. I fixed it with my push just a minute ago.

ryzhyk (approved):

mbudiu-vmw:
How about filing issues for the problems you discovered with APIs?

blp:
> How about filing issues for the problems you discovered with APIs?

Good idea. Filed issue #289 and issue #290.

blp:
Closing in favor of PR #295.

Issue #65, opened 2022-05-31
Move time tracking to the circuit.


ryzhyk:
### Is your feature request related to a problem? Please describe.

Instead of having each operator track current time, move time tracking into the circuit.

### Describe the solution you'd like

-

PR #291, opened 2023-02-04
Generalized incremental `distinct` operator.

ryzhyk:
We used to have two separate implementations of incremental `distinct`: one that only worked in the top-level scope and one that worked in the nested scope.  We replace them with a generalized implementation that works for arbitrary levels of nesting (although internally it still invokes a specialized implementation for the top-level scope, which is slightly more efficient).  The new implementation is also more general in that it works for indexed Z-sets, not just regular Z-sets. Finally, we rename the operator to be consistent with the rest of the API: `distinct` for the incremental version and `stream_distinct` for the non-incremental version.

Our implementation of `distinct` follows a different approach than the closely related `aggregate` operator (in fact, we _could_ implement `distinct` on top of `aggregate` and probably should actually have such an implementation for comparison): `aggregate` indexes both its input and output collections.  The latter means that the implementation must only compute the value of the aggregate at the current time and the delta can be diffing it with old outputs stored in the output trace.

For `distinct` we avoid indexing the output by computing the delta directly on the input trace.  This trades memory for CPU, but the hypothesis is that the cost of computing `distinct` is low enough that this doesn't matter in practice.  In reality, this design does not always save memory: oftentimes the output of `distinct` is fed to an operator that maintain the trace of its input, in which case the output trace that we optimized away still ends up being created.

Theory
------

According to the definition of a lifted incremental operator, we are computing a partial derivative:

```text
    ∂^n                      ∂^(n-1)                      ∂^(n-1)
 ------------f(t1,..,tn) = ------------f(t1,..,tn)  -  ------------f(t1-1,..,tn).
 ∂_t1 .. ∂_tn              ∂_t2 .. ∂_tn                ∂_t2 .. ∂_tn
```

where `n` is the nesting level of the circuit, i.e., 1 for the top-level circuit, 2 for its subcircuit, etc., `ti` is the i'th component of the n-dimensional timestamp, and `f(t1..tn)` is the value of the function we want to compute (in this case `distinct`) at time `t1..tn`.

For example, for a twice nested circuit, we get:

```text
     ∂^3                        ∂^2                         ∂^2
 --------------f(t1,t2,t3) = -----------f(t1,t2,t3)  -  -----------f(t1-1,t2,t3) =
 ∂_t1 ∂_t2 ∂_t3               ∂_t2 ∂_t3                  ∂_t2 ∂_t3

  ┌─                                     ─┐   ┌─                                          ─┐
  │   ∂                   ∂               │   │   ∂                    ∂                   │
  │ -----f(t1,t2,t3)  - -----f(t1,t2-1,t3)│ - │ -----f(t1-1,t2,t3)  - -----f(t1-1,t2-1,t3) │ =
= │ ∂_t3                 ∂_t3             │   │ ∂_t3                  ∂_t3                 │
  └─                                     ─┘   └─                                          ─┘

= ((f(t1,t2,t3) - f(t1,t2,t3-1)) - (f(t1,t2-1,t2) - f(t1,t2-1,t3-1))) -
  ((f(t1-1,t2,t3) - f(t1-1,t2,t3-1)) - (f(t1-1,t2-1,t2) - f(t1-1,t2-1,t3-1))).
```

In general, in order to compute the partial derivative, we need to know the values of `f` for all times within the n-dimensional cube with edge 1 of the current time `t1..tn` (i.e., all `2^n` possible combinations of `ti`'s and `ti-1`'s).

Computing the value of `f` at time `t`, in turn, requires first computing the sum of all updates with time `t' < t` in the input stream:

```text
             __
             \
f(t) =dist(  / stream[t'][k][v].weight )
             --
             t'<=t
```

where `stream[t'][k][v].weight` is pseudocode for selecting the weight of the update at time `t'` for key `k` and value `v`, and

```text
           ┌
           │  1, if w > 0
dist(w) = <
           │  0, otherwise
           └
```

We compute `f(t)` for all `2^n` times `t` of interest simultaneously in a single run of the cursor by maintaining an array of `2^n` accumulators and updating a subset of them at each step.

`keys_of_interest` map
----------------------

The `distinct` operator does not have nice properties like linearity that help with incremental evaluation: computing an update for each key/value pair requires inspecting the entire history of updates for this pair.  Fortunately, we do not need to look at all key/value pairs in the trace.  Specifically, a key/value pair `(k, v)` can only appear in the output of the operator at time `t` if it satisfies one of the following conditions:

* `(k,v)` occurs in the current input `delta`
* there exist `t1<t` and `t2<t` such that `t = t1 T t2` and `(k,v)` occurs in the inputs of the operator at times `t1` and `t2`.

To efficiently compute tuples that satisfy the second condition, we use the `keys_of_interest` map.  For each `(k,v)` observed at time `t1` we lookup the smallest time `t'` such that `t' = t1 T t2` for some `t2` such that `not (t2 <= t1)` at which we've encountered `(k,v)` and record `(k,v)` in `keys_of_interest[t']`.  When evaluating the operator at time `t'` we simply scan all tuples in `keys_of_interest[t2]` for candidates.

### DIFF
--- a/benches/ldbc-graphalytics/main.rs
+++ b/benches/ldbc-graphalytics/main.rs
@@ -455,6 +455,10 @@ fn attach_profiling(dataset: DataSet, circuit: &mut Circuit<()>) {
                             Cow::Borrowed("time"),
                             MetaItem::Duration(profile.total_time()),
                         ),
+                        (
+                            Cow::Borrowed("nodeid"),
+                            MetaItem::String(node_id.to_string()),
+                        ),
                     ];
 
                     for item in default_meta {
--- a/src/nexmark/queries/q15.rs
+++ b/src/nexmark/queries/q15.rs
@@ -176,37 +176,37 @@ pub fn q15(input: NexmarkStream) -> Q15Stream {
     // Compute unique bidders across all bids and for each price range.
     let distinct_bidder = bids
         .map(|(day, (_auction, _price, bidder))| (*day, *bidder))
-        .distinct_incremental()
+        .distinct::<()>()
         .index();
     let rank1_distinct_bidder = rank1_bids
         .map(|(day, (_auction, _price, bidder))| (*day, *bidder))
-        .distinct_incremental()
+        .distinct::<()>()
         .index();
     let rank2_distinct_bidder = rank2_bids
         .map(|(day, (_auction, _price, bidder))| (*day, *bidder))
-        .distinct_incremental()
+        .distinct::<()>()
         .index();
     let rank3_distinct_bidder = rank3_bids
         .map(|(day, (_auction, _price, bidder))| (*day, *bidder))
-        .distinct_incremental()
+        .distinct::<()>()
         .index();
 
     // Compute unique auctions across all bids and for each price range.
     let distinct_auction = bids
         .map(|(day, (auction, _price, _bidder))| (*day, *auction))
-        .distinct_incremental()
+        .distinct::<()>()
         .index();
     let rank1_distinct_auction = rank1_bids
         .map(|(day, (auction, _price, _bidder))| (*day, *auction))
-        .distinct_incremental()
+        .distinct::<()>()
         .index();
     let rank2_distinct_auction = rank2_bids
         .map(|(day, (auction, _price, _bidder))| (*day, *auction))
-        .distinct_incremental()
+        .distinct::<()>()
         .index();
     let rank3_distinct_auction = rank3_bids
         .map(|(day, (auction, _price, _bidder))| (*day, *auction))
-        .distinct_incremental()
+        .distinct::<()>()
         .index();
 
     // Compute bids per day.
--- a/src/operator/aggregate/mod.rs
+++ b/src/operator/aggregate/mod.rs
@@ -611,9 +611,6 @@ where
         // );
         self.empty_input = delta.is_empty();
 
-        // We iterate over keys in order, so it is safe to use `Builder`
-        // as long as we are careful to add values in order for each key in
-        // `eval_key` method.
         let mut result = Vec::with_capacity(delta.key_count());
 
         let mut delta_cursor = delta.cursor();
@@ -810,7 +807,7 @@ mod test {
                 Ok((
                     move || {
                         *counter.borrow_mut() += 1;
-                        Ok(*counter.borrow() == 4)
+                        Ok(*counter.borrow() == MAX_ITERATIONS)
                     },
                     (),
                 ))
--- a/src/operator/distinct.rs
+++ b/src/operator/distinct.rs
@@ -1,35 +1,43 @@
 //! Distinct operator.
 
 use crate::{
-    algebra::{AddAssignByRef, AddByRef, HasOne, HasZero, IndexedZSet, ZRingValue, ZSet},
+    algebra::{AddByRef, HasOne, HasZero, IndexedZSet, Lattice, PartialOrder, Present, ZRingValue},
     circuit::{
         metadata::{MetaItem, OperatorMeta},
         operator_traits::{BinaryOperator, Operator, UnaryOperator},
         Circuit, GlobalNodeId, Scope, Stream,
     },
     circuit_cache_key,
-    time::NestedTimestamp32,
-    trace::{ord::OrdKeySpine, BatchReader, Builder, Cursor as TraceCursor, Trace},
+    trace::{ord::OrdValSpine, Batch, BatchReader, Builder, Cursor as TraceCursor, Trace},
+    DBTimestamp, OrdIndexedZSet, Timestamp,
 };
 use size_of::SizeOf;
 use std::{
     borrow::Cow,
-    cmp::Ordering,
-    collections::{BTreeSet, HashMap},
+    cmp::{min, Ordering},
+    collections::BTreeMap,
     marker::PhantomData,
-    ops::{Add, Neg},
+    ops::Neg,
 };
 
 circuit_cache_key!(DistinctId<C, D>(GlobalNodeId => Stream<C, D>));
 circuit_cache_key!(DistinctIncrementalId<C, D>(GlobalNodeId => Stream<C, D>));
-circuit_cache_key!(DistinctTraceId<C, D>(GlobalNodeId => Stream<C, D>));
 
 impl<P, Z> Stream<Circuit<P>, Z>
 where
     P: Clone + 'static,
 {
-    /// Apply [`Distinct`] operator to `self`.
-    pub fn distinct(&self) -> Stream<Circuit<P>, Z>
+    /// Reduces input batches to one occurrence of each element.
+    ///
+    /// For each input batch `B`, the operator produces an output batch
+    /// that contains at most one occurrence of each tuple in `B`.
+    /// Specifically, for each input tuple `(key, value, weight)` with
+    /// `weight > 0` the operator produces an output tuple `(key, value, 1)`.
+    /// Tuples whose `weight` si `<= 0` are dropped.
+    ///
+    /// Intuitively, the operator converts the input multiset into a set
+    /// by eliminating duplicates.
+    pub fn stream_distinct(&self) -> Stream<Circuit<P>, Z>
     where
         Z: IndexedZSet + Send,
         Z::R: ZRingValue,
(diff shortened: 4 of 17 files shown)


src/operator/distinct.rs
```diff
@@ -281,177 +268,248 @@ where
...
+                            None => Some(self.time.join(trace_ts)),
+                            Some(time_of_interest) => {
+                                Some(min(time_of_interest, self.time.join(trace_ts)))
+                            }
+                        }
+                    }
+                });
+
+                // Compute `dist` for each entry in `distinct_vals`.
+                for (_time, weight) in self.distinct_vals.iter_mut() {
```
Kixiron, line 443:
```suggestion
                for weight in self.distinct_vals.values_mut() {
```

src/operator/distinct.rs
```diff
@@ -281,177 +268,248 @@ where
...
+                &mut vals[0..half_len],
+                ts.as_ref()
+                    .and_then(|ts| ts.checked_recede(half_len.ilog2() as Scope)),
+            );
+            Self::init_distinct_vals(&mut vals[half_len..], ts);
+        }
+    }
+
+    fn clear_distinct_vals(&mut self) {
+        for (_ts, val) in self.distinct_vals.iter_mut() {
```
Kixiron, line 483:
```suggestion
        for val in self.distinct_vals.values_mut() {
```

src/operator/distinct.rs
```diff
@@ -281,177 +268,248 @@ where
...
     fn name(&self) -> Cow<'static, str> {
-        Cow::Borrowed("DistinctTrace")
+        Cow::Borrowed("DistinctIncremental")
     }
 
     fn metadata(&self, meta: &mut OperatorMeta) {
-        let size: usize = self.future_updates.values().map(BTreeSet::len).sum();
-        let bytes = self.future_updates.size_of();
+        let size: usize = self.keys_of_interest.values().map(Vec::len).sum();
+        let bytes = self.keys_of_interest.size_of();
```
Kixiron, line 512:
```suggestion
        let size: usize = self.keys_of_interest.values().map(Vec::len).sum();
        let bytes = size_of::size_of_values([
            &self.keys_of_interes as &dyn SizeOf,
            &self.distinct_vals as &dyn SizeOf,
        ]);
```
I'm assuming from the comment that `distinct_vals` is always kept empty within `self` but if not then its length should be added to `size`

Kixiron: Now that I look at it more, I'd almost prefer if we just put a `#[derive(SizeOf)]` onto the `DistinctIncremental` struct itself and then just call `self.size_of()`

src/operator/distinct.rs
```diff
@@ -463,132 +521,205 @@ where
...
-        // TODO: generalize `DistinctTrace` to support arbitrarily nested scopes.
-        assert_eq!(scope, 0);
+        let epoch_end = self.time.epoch_end(scope);
+        /*println!("time: {:?} epoch_end: {:?} fixedpoint: empty_output: {}, empty_input: {}, keys_of_interest: {}"
+        , self.time
+        , epoch_end
+        , self.empty_output
+        , self.empty_input
+        , self.keys_of_interest.keys()
+           .all(|ts| !ts.less_equal(&epoch_end)));*/
```
Kixiron, line 548:
```suggestion
```

src/operator/distinct.rs
```diff
@@ -463,132 +521,205 @@ where
...
-    // `future_updates[t2]`. Then when evaluating an operatpr at time `t` we
-    // simply scan all keys in `delta` and all keys in `future_updates[t]` for
-    // candidates.
-    //
-    // TODO: Add example.
-    //
     fn eval(&mut self, delta: &Z, trace: &T) -> Z {
-        //eprintln!("distinct_trace {}", self.time);
+        Self::init_distinct_vals(&mut self.distinct_vals, Some(self.time.clone()));
+        self.empty_input = delta.is_empty();
```
Kixiron, line 567: If the input is empty do we need to do any further work?

src/operator/distinct.rs
```diff
@@ -463,132 +521,205 @@ where
...
+        let keys_of_interest = self.keys_of_interest.remove(&self.time).unwrap_or_default();
+
+        let keys_of_interest =
+            <OrdIndexedZSet<Z::Key, Z::Val, Present>>::from_tuples((), keys_of_interest);
+        let mut keys_of_interest_cursor = keys_of_interest.cursor();
+
+        // Iterate over all keys in `delta_cursor` and `keys_of_interest`.
+        while delta_cursor.key_valid() && keys_of_interest_cursor.key_valid() {
+            // println!("delta key = {key_of_interest:?}, val = {val_of_interest:?}");
+
```
Kixiron, line 586:
```suggestion
```

src/operator/distinct.rs
```diff
@@ -463,132 +521,205 @@ where
...
+                                );
+                                delta_cursor.step_val();
+                                keys_of_interest_cursor.step_val();
+                            }
+                        }
+                    }
+
+                    // Iterate over remaining `delta_cursor` values.
+                    while delta_cursor.val_valid() {
+                        // println!("(key, val) = {:?}", delta_cursor.key_val());
```
Kixiron, line 658:
```suggestion
```

src/operator/distinct.rs
```diff
@@ -463,132 +521,205 @@ where
...
-                HasZero::zero(),
-                &mut builder,
-            );
-            candidate = cand_iterator.next();
+
+        // Iterate over remaining `keys_of_interest_cursor` keys.
+        while keys_of_interest_cursor.key_valid() {
+            trace_cursor.seek_key(keys_of_interest_cursor.key());
+
+            // println!("key = {key_of_interest:?}, val = {val_of_interest:?}");
```
Kixiron, line 706:
```suggestion
```

src/operator/distinct.rs
```diff
@@ -463,132 +521,205 @@ where
...
+        // Iterate over remaining `delta_cursor` keys.
         while delta_cursor.key_valid() {
-            let w = delta_cursor.weight();
-            let k = delta_cursor.key();
-
-            self.eval_value(&mut trace_cursor, trace, k, w, &mut builder);
+            trace_cursor.seek_key(delta_cursor.key());
+
+            while delta_cursor.val_valid() {
+                // println!("(key, val) = {:?}", delta_cursor.key_val());
```
Kixiron, line 690:
```suggestion
```

src/operator/distinct.rs
```diff
@@ -463,132 +521,205 @@ where
...
-    // `future_updates` map, where for each key observed in the current epoch
-    // at time `t1` we lookup the smallest time `t2 > t1` (if any) at which we saw
-    // the key during any previous epochs and record this key in
-    // `future_updates[t2]`. Then when evaluating an operatpr at time `t` we
-    // simply scan all keys in `delta` and all keys in `future_updates[t]` for
-    // candidates.
-    //
-    // TODO: Add example.
-    //
     fn eval(&mut self, delta: &Z, trace: &T) -> Z {
```
Kixiron, line 565: It seems like an `eval_owned()` implementation could be beneficial for this operator since it clones inputs

Kixiron (approved):

benches/ldbc-graphalytics/main.rs
```diff
@@ -460,6 +460,13 @@ fn attach_profiling(dataset: DataSet, circuit: &mut Circuit<()>) {
                     for item in default_meta {
                         meta.insert(0, item);
                     }
+                    meta.insert(
+                        0,
+                        (
+                            Cow::Borrowed("nodeid"),
+                            MetaItem::String(node_id.to_string()),
+                        ),
+                    );
```
Kixiron, line 469: This should be added to `default_meta` instead of being broken out by itself

src/operator/distinct.rs
ryzhyk: Yes, there can be more work for the operator in `keys_of_interest`

src/operator/distinct.rs
ryzhyk: We'll never get an owned `trace` (it's always owned by the Z-1 operator), but we can get an owned `delta`, and yeah, that should allow us to reduce cloning. I'll add a todo.

src/operator/distinct.rs
ryzhyk: This doesn't work on `Vec`'s

Kixiron: Ah that's my bad, I thought it was a map

mbudiu-vmw:
Perhaps a \/ b is a better notation than a T b.

mbudiu-vmw (approved):
I couldn't rewrite this, but I am approving

src/nexmark/queries/q15.rs
```diff
@@ -176,37 +176,37 @@ pub fn q15(input: NexmarkStream) -> Q15Stream {
     // Compute unique bidders across all bids and for each price range.
     let distinct_bidder = bids
         .map(|(day, (_auction, _price, bidder))| (*day, *bidder))
-        .distinct_incremental()
+        .distinct::<()>()
```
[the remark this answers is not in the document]
ryzhyk: This is #65. Same for the `NestedTimestam32` question below.  `join` is another victim of the same issue.  This is the next thing I plan to work on (whenever I get around to working on the library).

src/operator/aggregate/mod.rs
```diff
@@ -810,7 +807,7 @@ mod test {
                 Ok((
                     move || {
                         *counter.borrow_mut() += 1;
-                        Ok(*counter.borrow() == 4)
+                        Ok(*counter.borrow() == MAX_ITERATIONS)
```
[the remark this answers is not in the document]
ryzhyk: yes, 4 was wrong

src/operator/distinct.rs
```diff
@@ -169,43 +155,44 @@ where
     }
 }
 
-/// Incremental version of the distinct operator.
+/// Incremental version of the distinct operator that only works in the
```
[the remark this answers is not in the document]
ryzhyk: It works for totally ordered timestamps ... In any case, this API is not exposed to the user.

src/operator/distinct.rs
```diff
@@ -281,177 +268,248 @@ where
...
+    ///
+    /// The `distinct` operator does not have nice properties like linearity
+    /// that help with incremental evaluation: computing an update for each
+    /// key/value pair requires inspecting the entire history of updates for
+    /// this pair.  Fortunately, we do not need to look at all key/value pairs
+    /// in the trace.  Specifically, a key/value pair `(k, v)` can only
+    /// appear in the output of the operator at time `t` if it satisfies one
+    /// of the following conditions:
+    /// * `(k,v)` occurs in the current input `delta`
+    /// * there exist `t1<t` and `t2<t` such that `t = t1 T t2` and `(k,v)`
```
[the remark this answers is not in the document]
ryzhyk: fixed it

PR #292, opened 2023-02-06
Cleanup & document pipeline manager API.

ryzhyk:
Use the `utoipa` crate to generate OpenAPI spec for the pipeline manager API, and Swagger documentation for it, available on `/api-doc/openapi.json` and `/swagger-ui/` endpoints respectively.

Unlike OpenAPI generators, `utoipa` doesn't automatically generate server-side stubs for the API, but rather documents an existing, manually implemented, API.  Although it does interact with the Rust type system, it is not 100% type-safe, i.e., the generated API may not match the actual input and output formats used by the server.  Unfortunately, this appears to be the best option the Rust ecosystem offers today.

While at it, I also cleaned up the API in a couple of ways:

- Use JSON for all inputs and outputs (as opposed to returning plain text strings in some cases).  In particular, errors are now returned as JSON:

  ``` { "message": "Cannot delete a project while some of its pipelines are running" } ```

- Convert `list_project_configs` and `list_project_pipelines` into `GET` methods that take project id as URL-encoded parameter for consistency with `project_status`, `project_code`, etc.

@mbudiu-vmw, the preferred way to work with this API from the browser would be to generate client-side typescript bindings for it, but it's probably not worth bothering with now, since we will eventually re-implement the UI anyway.

### DIFF
--- a/pipeline_manager/scripts/delete_project.sh
+++ b/pipeline_manager/scripts/delete_project.sh
@@ -9,4 +9,4 @@ if [ "$#" -ne 1 ]; then
     exit 1
 fi
 
-curl -X POST http://localhost:8080/delete_project  -H 'Content-Type: application/json' -d '{"project_id":'$1'}'
+curl -X DELETE http://localhost:8080/projects/$1 -H 'Content-Type: application/json' -d '{}'
--- a/pipeline_manager/src/db.rs
+++ b/pipeline_manager/src/db.rs
@@ -1,9 +1,11 @@
 use crate::{ManagerConfig, ProjectStatus};
 use anyhow::{anyhow, Error as AnyError, Result as AnyResult};
+use chrono::{DateTime, Utc};
 use log::error;
-use serde::Serialize;
-use std::{error::Error as StdError, fmt, fmt::Display, time::SystemTime};
+use serde::{Deserialize, Serialize};
+use std::{error::Error as StdError, fmt, fmt::Display};
 use tokio_postgres::{Client, NoTls};
+use utoipa::ToSchema;
 
 /// Project database API.
 ///
@@ -22,10 +24,55 @@ pub(crate) struct ProjectDB {
     dbclient: Client,
 }
 
-pub(crate) type ProjectId = i64;
-pub(crate) type ConfigId = i64;
-pub(crate) type PipelineId = i64;
-pub(crate) type Version = i64;
+/// Unique project id.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
+#[repr(transparent)]
+#[serde(transparent)]
+pub(crate) struct ProjectId(pub i64);
+impl Display for ProjectId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+/// Unique configuration id.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
+#[repr(transparent)]
+#[serde(transparent)]
+pub(crate) struct ConfigId(pub i64);
+impl Display for ConfigId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+/// Unique pipeline id.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
+#[repr(transparent)]
+#[serde(transparent)]
+pub(crate) struct PipelineId(pub i64);
+impl Display for PipelineId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+/// Version number.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
+#[repr(transparent)]
+#[serde(transparent)]
+pub(crate) struct Version(i64);
+impl Display for Version {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl Version {
+    fn increment(&self) -> Self {
+        Self(self.0 + 1)
+    }
+}
 
 #[derive(Debug)]
 pub(crate) enum DBError {
@@ -85,10 +132,7 @@ impl ProjectStatus {
 }
 
 /// Project descriptor.
-///
-/// This is a Rust representation of a record from the `project` table, sans
-/// the potentially large `code` field, which must be retrieved separately.
-#[derive(Serialize)]
+#[derive(Serialize, ToSchema)]
 pub(crate) struct ProjectDescr {
     /// Unique project id.
     pub project_id: ProjectId,
@@ -100,10 +144,8 @@ pub(crate) struct ProjectDescr {
     pub status: ProjectStatus,
 }
 
-/// Project config descriptor.
-///
-/// This is a Rust representation of a record from the `project_config` table.
-#[derive(Serialize)]
+/// Project configuration descriptor.
+#[derive(Serialize, ToSchema)]
 pub(crate) struct ConfigDescr {
     pub config_id: ConfigId,
     pub project_id: ProjectId,
@@ -113,16 +155,14 @@ pub(crate) struct ConfigDescr {
 }
 
 /// Pipeline descriptor.
-///
-/// This is a Rust represenation of a record from the `pipeline` table.
-#[derive(Serialize)]
+#[derive(Serialize, ToSchema)]
 pub(crate) struct PipelineDescr {
-    pub pipeline_id: ConfigId,
+    pub pipeline_id: PipelineId,
     pub project_id: ProjectId,
     pub project_version: Version,
     pub port: u16,
     pub killed: bool,
-    pub created: SystemTime,
+    pub created: DateTime<Utc>,
 }
 
 impl ProjectDB {
@@ -171,9 +211,9 @@ impl ProjectDB {
             let status = ProjectStatus::from_columns(status, error)?;
 
             result.push(ProjectDescr {
-                project_id: row.try_get(0)?,
+                project_id: ProjectId(row.try_get(0)?),
                 name: row.try_get(1)?,
-                version: row.try_get(2)?,
+                version: Version(row.try_get(2)?),
                 status,
             });
         }
@@ -187,12 +227,12 @@ impl ProjectDB {
             .dbclient
             .query_opt(
                 "SELECT version, code FROM project WHERE id = $1",
-                &[&project_id],
+                &[&project_id.0],
             )
             .await?
             .ok_or(DBError::UnknownProject(project_id))?;
 
-        Ok((row.try_get(0)?, row.try_get(1)?))
+        Ok((Version(row.try_get(0)?), row.try_get(1)?))
     }
 
     /// Create a new project.
@@ -205,16 +245,16 @@ impl ProjectDB {
             .dbclient
             .query_one("SELECT nextval('project_id_seq')", &[])
             .await?;
-        let id: ProjectId = row.try_get(0)?;
+        let id: ProjectId = ProjectId(row.try_get(0)?);
 
         self.dbclient
             .execute(
                 "INSERT INTO project (id, version, name, code, status_since) VALUES($1, 1, $2, $3, now())",
-                &[&id, &project_name, &project_code],
+                &[&id.0, &project_name, &project_code],
             )
             .await?;
 
-        Ok((id, 1))
+        Ok((id, Version(1)))
     }
 
     /// Update project name and, optionally, code.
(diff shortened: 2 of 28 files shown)


pipeline_manager/src/main.rs
```diff
@@ -367,25 +482,36 @@ async fn project_status(state: WebData<ServerState>, req: HttpRequest) -> impl R
...
 
 /// Create a new project.
+#[utoipa::path(
+    request_body = NewProjectRequest,
+    responses(
+        (status = CREATED, description = "Project created successfully", body = NewProjectResponse),
+    ),
+    tag = "Project"
+)]
 #[post("/new_project")]
```
lalithsuresh, line 515: `POST /projects`

pipeline_manager/src/main.rs
```diff
@@ -475,12 +608,21 @@ struct CompileProjectRequest {
...
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Unknown project id '42'"))),
+        (status = CONFLICT
+            , description = "Project version specified in the request doesn't match the latest project version in the database."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Outdated project version '{version}'"))),
+    ),
+    tag = "Project"
+)]
 #[post("/compile_project")]
```
lalithsuresh, line 626: `POST /projects/{project_id}/compile`

pipeline_manager/src/main.rs
```diff
@@ -509,13 +651,21 @@ struct CancelProjectRequest {
...
+            , example = json!(ErrorResponse::new("Unknown project id '42'"))),
+        (status = CONFLICT
+            , description = "Project version specified in the request doesn't match the latest project version in the database."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Outdated project version '{3}'"))),
+    ),
+    tag = "Project"
+)]
 #[post("/cancel_project")]
 async fn cancel_project(
```
lalithsuresh, line 670: What's the difference between canceling a project and deleting it?

pipeline_manager/src/main.rs
```diff
@@ -588,10 +747,17 @@ struct NewConfigResponse {
...
+    responses(
+        (status = OK, description = "Configuration successfully created.", body = NewConfigResponse),
+        (status = NOT_FOUND
+            , description = "Specified `project_id` does not exist in the database."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Unknown project id '42'"))),
+    ),
+    tag = "Config"
+)]
 #[post("/new_config")]
```
lalithsuresh, line 761: `POST /configs/`

pipeline_manager/src/main.rs
```diff
@@ -636,10 +802,17 @@ struct UpdateConfigResponse {
...
+    responses(
+        (status = OK, description = "Configuration successfully updated.", body = UpdateConfigResponse),
+        (status = NOT_FOUND
+            , description = "Specified `config_id` does not exist in the database."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Unknown config id '5'"))),
+    ),
+    tag = "Config"
+)]
 #[post("/update_config")]
```
lalithsuresh, line 816: `POST /configs/{config_id}` or maybe even PATCH

pipeline_manager/src/main.rs
```diff
@@ -661,17 +834,25 @@ async fn update_config(
...
+    responses(
+        (status = OK, description = "Configuration successfully deleted."),
+        (status = NOT_FOUND
+            , description = "Specified `config_id` does not exist in the database."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Unknown config id '5'"))),
+    ),
+    tag = "Config"
+)]
 #[post("/delete_config")]
```
lalithsuresh, line 856: `DELETE /configs/{config_id}`

pipeline_manager/src/main.rs
```diff
@@ -687,30 +868,34 @@ async fn delete_config(
...
+            , description = "Specified `project_id` does not exist in the database."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Unknown project id '42'"))),
+    ),
+    params(
+        ("project_id" = i64, Path, description = "Unique project identifier")
+    ),
+    tag = "Config"
+)]
+#[get("/list_project_configs/{project_id}")]
```
lalithsuresh, line 885: `GET /projects/{project_id}/configs`

pipeline_manager/src/main.rs
```diff
@@ -746,21 +931,32 @@ struct NewPipelineResponse {
...
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Config '9' does not belong to project '15'"))),
+        (status = INTERNAL_SERVER_ERROR
+            , description = "Pipeline process failed to initialize."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Failed to run 'project42': permission denied"))),
+    ),
+    tag = "Pipeline"
+)]
 #[post("/new_pipeline")]
```
lalithsuresh, line 960: `POST /pipeline/`

pipeline_manager/src/main.rs
```diff
@@ -773,30 +969,34 @@ async fn new_pipeline(
...
+            , description = "Specified `project_id` does not exist in the database."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Unknown project id '42'"))),
+    ),
+    params(
+        ("project_id" = i64, Path, description = "Unique project identifier")
+    ),
+    tag = "Pipeline"
+)]
+#[get("/list_project_pipelines/{project_id}")]
```
lalithsuresh, line 986: `GET /projects/{project_id}/pipelines`

pipeline_manager/src/main.rs
```diff
@@ -823,13 +1023,25 @@ pub(self) struct KillPipelineRequest {
...
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Unknown pipeline id '64'"))),
+        (status = INTERNAL_SERVER_ERROR
+            , description = "Request failed."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Failed to kill the pipeline; response from pipeline controller: ..."))),
+    ),
+    tag = "Pipeline"
+)]
 #[post("/kill_pipeline")]
```
lalithsuresh, line 1045: `DELETE /pipeline/{pipeline_id}` (or POST?) (what's the difference between kill pipeline and delete? is it just stop vs delete?)

lalithsuresh (commented):
I've made suggestions for how to keep the API design familiar to folks used to REST. In general, describe resources (nouns) and operations on them (GET, POST, etc.), rather than describing verbs. It's also easier to evolve later.

Could we add more documentation about the accepted configs? For example, I couldn't easily find examples for the file transport.

pipeline_manager/src/main.rs
ryzhyk: `/cancel` cancels compilation project compilation.  The project remains in the database and can be compiled/modified/queried.

pipeline_manager/src/main.rs
ryzhyk: So a create operation usually does not use a verb, i.e., not `/configs/new`?

pipeline_manager/src/main.rs
ryzhyk: delete = kill + delete from DB.  A killed pipeline can still be inspected until it is deleted.

pipeline_manager/src/runner.rs
```diff
@@ -441,17 +443,23 @@ scrape_configs:
...
+                    serde_json::to_string(&format!("Pipeline at '{url}' already killed.")).unwrap(),
+                ));
             }
         };
 
         if response.status().is_success() {
             db.set_pipeline_killed(pipeline_id).await?;
-            Ok(HttpResponse::Ok().finish())
+            Ok(HttpResponse::Ok()
+                .body(serde_json::to_string("Pipeline successfully terminated.").unwrap()))
```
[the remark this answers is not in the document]
ryzhyk: Same thing.  The API is called `kill` but then I decided that `terminate` is a better user-facing name.  Yeah, `shutdown` is probably better.

pipeline_manager/src/main.rs
```diff
@@ -327,24 +436,30 @@ struct ProjectStatusResponse {
...
+            , description = "Specified `project_id` does not exist in the database."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Unknown project id '42'"))),
+    ),
+    params(
+        ("project_id" = i64, Path, description = "Unique project identifier")
+    ),
+    tag = "Project"
+)]
 #[get("/project_status/{project_id}")]
```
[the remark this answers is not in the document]
ryzhyk: and `/list_projects` should be just `GET /project`?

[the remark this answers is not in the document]
lalithsuresh: Yup. `GET /projects` for example.

pipeline_manager/src/main.rs
lalithsuresh: Yup. POST is the verb, which usually creates something.

pipeline_manager/src/main.rs
ryzhyk: Isn't it better to do `POST /projects/compile` and extract `project_id` from JSON?  Otherwise, we have a mix of url-encoded and JSON arguments.

pipeline_manager/src/main.rs
lalithsuresh: To follow a create/<modify>/delete cycle for a project, here is an approach that comes to mind.

* `POST /projects/`     # creates a project, not yet compiled
* `GET /projects/{project_id}`
* `POST /compiler/{project_id}` # creates a compilation job
* `GET /compiler/{project_id}` # gets the status of the compilation job
* `DELETE /compiler/{project_id}` # ends the compilation job

An alternative that some people use, which might be overkill for now:

* `POST /projects/`     # creates a project, not yet compiled
* `GET /projects/{project_id}`
* `PUT/PATCH /projects/{project_id}`   # modify a "compilation" field in the project that changes its desired state, which triggers compilation. AWS APIs have this approach to attaching "Action" sub-resources to resources.

pipeline_manager/src/main.rs
lalithsuresh: For pipelines, it makes sense to have a status field for the pipeline that we can change with a `PUT` or a `PATCH` to start or stop it. So we'll have:

```
CREATE /pipelines
PUT/PATCH /pipelines/{pipeline_id}
DELETE /pipelines/{pipeline_id}
```

pipeline_manager/src/main.rs
lalithsuresh: Yes, I made a similar suggestion to that effect in another comment now that I understand the workflow better.

pipeline_manager/src/db.rs
```diff
@@ -523,16 +563,16 @@ impl ProjectDB {
...
             .query_one("SELECT nextval('project_config_id_seq')", &[])
             .await?;
-        let id: ConfigId = row.try_get(0)?;
+        let id: ConfigId = ConfigId(row.try_get(0)?);
 
         self.dbclient
             .execute(
                 "INSERT INTO project_config (id, project_id, version, name, config) VALUES($1, $2, 1, $3, $4)",
-                &[&id, &project_id, &config_name, &config],
+                &[&id.0, &project_id.0, &config_name, &config],
```
mbudiu-vmw, line 571: would a getter be better than accessing .0?

pipeline_manager/src/main.rs
```diff
@@ -434,13 +560,20 @@ struct UpdateProjectResponse {
...
 /// If project code changes, any ongoing compilation gets cancelled,
-/// project status is reset to `ProjectStatus::None`, and project version
+/// project status is reset to `None`, and project version
 /// is incremented by 1.  Changing project name only doesn't affect its
 /// version or the compilation process.
-///
-/// # HTTP errors
-///
-/// * `NOT_FOUND` - `project_id` does not exist in the database.
+#[utoipa::path(
```
mbudiu-vmw, line 567: these don't seem to save you much typing.
Are they mostly for documentation using swagger?

mbudiu-vmw (approved):

pipeline_manager/src/main.rs
ryzhyk: utoipa generates an OpenAPI spec, which is a piece of JSON that other tools can use to generate client-side bindings, e.g., I'm working on Python bindings now.  And it also generates swagger documentation from that.

pipeline_manager/scripts/delete_project.sh
```diff
@@ -9,4 +9,4 @@ if [ "$#" -ne 1 ]; then
     exit 1
 fi
 
-curl -X POST http://localhost:8080/delete_project  -H 'Content-Type: application/json' -d '{"project_id":'$1'}'
+curl -X DELETE http://localhost:8080/projects -H 'Content-Type: application/json' -d '{"project_id":'$1'}'
```
lalithsuresh, line 12: For consistency, it's better to do a delete to `/projects/{project_id}`. Similar issues with delete for pipelines and configs. Given that we already have URLs like `/projects/$1/{config, pipelines}` etc elsewhere.

lalithsuresh: Otherwise, what we're suggesting with the API is a way to delete a collection (e.g. multiple projects), in which case the payload should be able to state an array of IDs to delete. This is also fine, just want to be sure what the API is.

ryzhyk:
@lalithsuresh , I fixed the delete apis. I also implemented request forwarding via the main server.  When you click on `Pipelines` to see the list of running pipelines, there should be buttons to retrieve pipeline status, metadata, start and pause the pipeline. Let me know if it works for you, so we can cleanup and merge this.

lalithsuresh:
LGTM. I was able to run it end to end with a file based transport example. I'll send a PR for that after we merge this.

PR #293, opened 2023-02-06
Make it easy to bring up a local instance of DBSP

lalithsuresh:
Adds some changes to bring up a docker container locally with the DBSP pipeline manager running. For now, anyone trying this out would need to build the docker image locally first and then run the docker command to bring up the deployment. Later, we can make prebuilt images available so it becomes just the docker run command to get started with DBSP.

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -45,4 +45,7 @@ RUN cp /database-stream-processor/deploy/prometheus_datasource.yaml /etc/grafana
    && cp /database-stream-processor/deploy/grafana_dashboard_provision.yaml /etc/grafana/provisioning/dashboards/ \
    && cp /database-stream-processor/deploy/grafana_dashboard.json /etc/grafana/provisioning/dashboards/
 
-CMD service postgresql start && bash
+CMD service postgresql start && \
+    mkdir /working-dir && \
+    /database-stream-processor/deploy/start_manager.sh /working-dir 0.0.0.0 && \
+    bash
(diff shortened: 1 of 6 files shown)


deploy/Dockerfile
```diff
@@ -45,4 +45,7 @@ RUN cp /database-stream-processor/deploy/prometheus_datasource.yaml /etc/grafana
    && cp /database-stream-processor/deploy/grafana_dashboard_provision.yaml /etc/grafana/provisioning/dashboards/ \
    && cp /database-stream-processor/deploy/grafana_dashboard.json /etc/grafana/provisioning/dashboards/
 
-CMD service postgresql start && bash
+CMD service postgresql start && \
+    mkdir /working-dir && \
```
ryzhyk, line 49: I don't think this is necessary, the manager will create the dir if it doesn't exist

deploy/README.md
```diff
@@ -0,0 +1,16 @@
...
+./docker.sh
+```
+
+Next, bring up an instance of the container:
+
+```
+docker run --name dbsp -p 8081:8080 -itd dbspmanager
+```
+
+Open your browser and you should now be able to see the pipeline maanger dashboard on localhost:8081.
```
ryzhyk, line 16: typo

deploy/README.md
```diff
@@ -0,0 +1,16 @@
...
+First, build a DBSP Docker image:
+
+```
+./docker.sh
+```
+
+Next, bring up an instance of the container:
+
+```
+docker run --name dbsp -p 8081:8080 -itd dbspmanager
```
ryzhyk, line 13: Is this in case we're already running the DBSP manager on the host?

lalithsuresh: Yup. 8080 is commonly used across projects, but this is just to show you can pick the port you want to forward to.

ryzhyk (approved):

deploy/README.md
lalithsuresh: I've also clarified in the README that the user can pick any port to forward to.

PR #296, opened 2023-02-08
Add 'degrees' example similar to Differential Dataflow example.

blp:
The results of this example are similar to those for the one in the Differential Dataflow crate documentation root at: https://docs.rs/differential-dataflow/latest/differential_dataflow/

Signed-off-by: Ben Pfaff <bpfaff@vmware.com>

### DIFF
--- a/examples/degrees.rs
+++ b/examples/degrees.rs
@@ -1,0 +1,133 @@
+//! Simple DBSP example for counting the distribution of degrees in a graph.
+//!
+//! This program creates a series of graph edges, then counts the number of
+//! times each source node appears, then counts the number of times each count
+//! appears.  This is similar to the example for Differential Dataflow at
+//! <https://docs.rs/differential-dataflow/latest/differential_dataflow/>.
+
+use anyhow::Result;
+use clap::Parser;
+use dbsp::{
+    operator::FilterMap,
+    time::NestedTimestamp32,
+    trace::{BatchReader, Cursor},
+    DBData, DBWeight, OrdIndexedZSet, OutputHandle, Runtime,
+};
+
+type Node = usize;
+type Weight = isize;
+
+#[derive(Debug, Clone, Parser)]
+struct Args {
+    /// Number of initial edges in the graph.
+    #[clap(long, default_value = "100")]
+    edges: usize,
+
+    /// Number of source nodes in the graph.
+    #[clap(long, default_value = "13")]
+    sources: usize,
+
+    /// Number of extra edges added later to the graph.
+    #[clap(long, default_value = "5")]
+    extra: usize,
+
+    /// Number of threads.
+    #[clap(long, default_value = "2")]
+    threads: usize,
+}
+
+struct BatchIterator<'a, T: BatchReader> {
+    cursor: T::Cursor<'a>,
+}
+
+impl<'a, T: BatchReader> BatchIterator<'a, T> {
+    /// Returns an iterator of `(key, value, weight)` over the items that
+    /// `cursor` visits.
+    fn new(cursor: T::Cursor<'a>) -> Self {
+        Self { cursor }
+    }
+}
+
+impl<'a, K, V, R> Iterator for BatchIterator<'a, OrdIndexedZSet<K, V, R>>
+where
+    K: DBData,
+    V: DBData,
+    R: DBWeight,
+{
+    type Item = (K, V, R);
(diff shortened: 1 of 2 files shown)


examples/degrees.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl<'a, T: BatchReader> BatchIterator<'a, T> {
+    /// Returns an iterator of `(key, value, weight)` over the items that
+    /// `cursor` visits.
+    fn new(cursor: T::Cursor<'a>) -> Self {
+        Self { cursor }
+    }
+}
+
+impl<'a, K, V, R> Iterator for BatchIterator<'a, OrdIndexedZSet<K, V, R>>
```
mbudiu-vmw, line 51: shouldn't this be in dbsp?

mbudiu-vmw (approved):
is there a way to turn this into a benchmark too?

examples/degrees.rs
blp: I don't know. I don't understand why cursors aren't iterators.

examples/degrees.rs
```diff
@@ -0,0 +1,133 @@
...
+        let (edges, hedges) = circuit.add_input_zset::<(Node, Node), Weight>();
+
+        // Count the number of edges with each node as its source (each node's
+        // out-degree).
+        let degrees = edges.map(|&(src, _dst)| (src, 1)).index();
+        let degrees = degrees.aggregate_linear::<NestedTimestamp32, _, _>(|_k, v| *v);
+
+        // Count the number of nodes with each out-degree.
+        let distribution = degrees.map(|(_src, count)| (*count, 1)).index();
+        let distribution = distribution.aggregate_linear::<NestedTimestamp32, _, _>(|_k, v| *v);
```
ryzhyk, line 108:
```
let distribution = degrees
    .map(|(_src, count)| count)
    .aggregate_linear::<(), _, _>(|_count| 1)
```

blp:
> is there a way to turn this into a benchmark too?

I guess that I could add some performance measurement and move it to the benchmarks directory. I was trying to keep it as a simple example rather than make it a benchmark. But I could make it a benchmark if you like that better.

mbudiu-vmw:
More examples are necessary.
if there is a large dataset to benchmark this, could be a good benchmark too.

examples/degrees.rs
```diff
@@ -0,0 +1,133 @@
...
+        extra,
+    } = Args::parse();
+
+    let (mut dbsp, (hedges, degrees, distribution)) = Runtime::init_circuit(threads, |circuit| {
+        let (edges, hedges) = circuit.add_input_zset::<(Node, Node), Weight>();
+
+        // Count the number of edges with each node as its source (each node's
+        // out-degree).
+        let degrees = edges.map(|&(src, _dst)| (src, 1)).index();
+        let degrees = degrees.aggregate_linear::<NestedTimestamp32, _, _>(|_k, v| *v);
```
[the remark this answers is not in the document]
blp: Oh, maybe I'm getting it figured out now, never mind.

examples/degrees.rs
ryzhyk: This is partially for historic reasons.  Before GATs were standardized, the cursor API was more complicated and couldn't be framed as an iterator.

Switching to an iterator-based API is worth exploring if we can do it without a performance penalty. There are several different ways to iterate over a batch. E.g., we may want to iterate over (key,value,weight) tuples as you do here or over keys with a nested iterator over (value, weight) tuples.

mbudiu-vmw: No reason for a BatchAllKeyValuesIterator not to exist.

examples/degrees.rs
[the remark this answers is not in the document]
ryzhyk: I guess you already figured this out, but just to clarify: `aggregate_linear` requires it input to implement `trait IndexedZSet`. As it happens, `trait ZSet` is a specialization of `IndexedZSet` with `Val = ()`, so any `ZSet`, including `OrdZSet` is acceptable.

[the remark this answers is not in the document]
blp: I was surprised to find that out! I had guessed that OrdZSet didn't have an index and therefore couldn't be used as an IndexedZSet.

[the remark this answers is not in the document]
blp: Fixed this with my recent push.

ryzhyk (approved):

PR #297, opened 2023-02-08
Add a hello world example

lalithsuresh:
I didn't explain all the steps to use the current UI given its a placeholder anyway.

Signed-off-by: Lalith Suresh <lsuresh@vmware.com>


examples/orgchart/orgchart.rs
```diff
@@ -0,0 +1,136 @@
+//! Simple DBSP example.
```
mbudiu-vmw, line 1: why is this file here?

lalithsuresh: Check again, messed up the git move.

examples/hello-world/README
```diff
@@ -0,0 +1,18 @@
+First, build the DBSP docker container following the instructions
```
ryzhyk, line 1: `examples` is a special folder where cargo looks for when one runs `cargo run --example`.  We shouldn't abuse it for end-to-end examples.  The right thing to do is to reorganize the project moving all crates to a separate `crates` folder so we can have end-to-end `examples` here.  We can do this once other outstanding PRs have been merged. Maybe we can keep it with other projects in `pipeline_manager/demo` for now?

PR #298, opened 2023-02-09
Iterator interface to `trait IndexedZSet`.

ryzhyk:
Internally DBSP uses the cursor API for iterating over batches.  Cursors are designed to be fast but they have relatively poor ergonomics and are unsafe.  This is ok for internal use by the library, but we expose batches to the user via output handles.  We therefore add an iterator-based interface to batches (specifically, for indexed z-sets only) in order to make it easier and less error-prone for users to iterate over the output of the computation.

Unfortunately, the only way to wrap `Cursor` in `trait Iterator` is to iterate over updates by value (because `Iterator` doesn't take advantage of GATs).

@blp , I moved your `impl Iterator` for batches into the library, so you can simplify #296

### DIFF
--- a/src/algebra/zset/mod.rs
+++ b/src/algebra/zset/mod.rs
@@ -47,9 +47,56 @@ pub trait IndexedZSet: Batch<Time = ()> + GroupValue + NumEntries {
     {
         self.distinct()
     }
+
+    /// Returns an iterator over updates in the indexed Z-set.
+    fn iter(&self) -> IndexedZSetIterator<Self> {
+        IndexedZSetIterator::new(self.cursor())
+    }
 }
 
 impl<Z> IndexedZSet for Z where Z: Batch<Time = ()> + GroupValue + NumEntries {}
+
+/// Iterator over `(key, value, weight)` tuples of an indexed Z-set.
+pub struct IndexedZSetIterator<'a, Z>
+where
+    Z: IndexedZSet,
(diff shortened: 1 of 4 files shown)


blp:
Thank you! I added a commit to convert `orgchart` to use this.

blp (approved):
Look great, thank you.

Kixiron (commented):
I'm not a huge fan of the cloning, but we could address that in the future

Kixiron (approved):

src/algebra/zset/mod.rs
```diff
@@ -47,10 +47,57 @@ pub trait IndexedZSet: Batch<Time = ()> + GroupValue + NumEntries {
...
+    fn new(cursor: Z::Cursor<'a>) -> Self {
+        Self { cursor }
+    }
+}
+
+impl<'a, Z> Iterator for IndexedZSetIterator<'a, Z>
+where
+    Z: IndexedZSet,
+{
+    type Item = (Z::Key, Z::Val, Z::R);
```
Kixiron, line 82: Actually, couldn't we just use the lifetime here?
```suggestion
    type Item = (&'a Z::Key, &'a Z::Val, Z::R);
```

ryzhyk: I tried hard to make it work. Don't think it's possible.

PR #303, opened 2023-02-18
Make sure manager can run without a config.yaml.

gz:
- Define defaults for everything in config.yaml
- Add an option to run it as daemon or not (default is false was true before (on unix)).

With this I can run the pipeline manager to run by just typing `cargo run` on the command line (for easier development).

### DIFF
--- a/pipeline_manager/src/config.rs
+++ b/pipeline_manager/src/config.rs
@@ -18,6 +18,14 @@ fn default_working_directory() -> String {
     ".".to_string()
 }
 
+fn default_logfile() -> String {
+    "pipeline_manager.log".to_string()
+}
+
+fn default_sql_compiler_home() -> String {
+    "../../sql-to-dbsp-compiler".to_string()
+}
+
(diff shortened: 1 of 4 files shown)


pipeline_manager/src/config.rs
```diff
@@ -18,28 +19,44 @@ fn default_working_directory() -> String {
     ".".to_string()
 }
 
-/// Pipeline manager configuration read from a YAML config file.
-#[derive(Deserialize, Clone)]
+fn default_sql_compiler_home() -> String {
+    "../sql-to-dbsp-compiler".to_string()
```
ryzhyk, line 23: PS. We are moving to the directory layout where the SQL compiler lives in the main DBSP folder (even if it's a separate repo)

ryzhyk:
@lalithsuresh , with this PR docker can run the manager directly providing the right CLI flags, without the `start_manager` script.  It also shouldn't need to run bash, just don't pass the `--unix-daemon` flag and it will keep running as the main container process.

PR #305, opened 2023-02-20
Two more flags for the pipeline manager

gz:
- Added permissive CORS settings for debug mode.
- Added an option to inject a SQL file on startup.

### DIFF
--- a/pipeline_manager/src/config.rs
+++ b/pipeline_manager/src/config.rs
@@ -110,6 +110,24 @@ pub(crate) struct ManagerConfig {
     #[serde(skip)]
     #[arg(short, long)]
     pub config_file: Option<String>,
+
+    /// [Developers only] Inject a SQL file into the database when starting the
+    /// manager.
+    ///
+    /// This is useful to populate the DB with state for testing.
+    #[serde(skip)]
+    #[arg(short, long)]
+    pub initial_sql: Option<String>,
+
+    /// [Developers only] Run in development mode.
+    ///
+    /// This runs with permissive CORS settings and allows the manager to be
+    /// accessed from a different host/port.
+    ///
+    /// The default is `false`.
+    #[serde(default)]
+    #[arg(long)]
+    pub dev_mode: bool,
 }
 
 impl ManagerConfig {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,6 +20,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "actix-cors"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b340e9cfa5b08690aae90fb61beb44e9b06f44fe3d0f93781aaa58cfba86245e"
+dependencies = [
+ "actix-utils",
+ "actix-web",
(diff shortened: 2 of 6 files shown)


pipeline_manager/src/config.rs
```diff
@@ -110,6 +110,24 @@ pub(crate) struct ManagerConfig {
...
+
+    /// [Developers only] Run in development mode.
+    ///
+    /// This runs with permissive CORS settings and allows the manager to be
+    /// accessed from a different host/port.
+    ///
+    /// The default is `false`.
+    #[serde(default)]
+    #[arg(long)]
+    pub dev_mode: bool,
```
ryzhyk, line 130: How's this different from binding it to 0.0.0.0 instead of 127.0.0.1?

pipeline_manager/src/config.rs
```diff
@@ -110,6 +110,24 @@ pub(crate) struct ManagerConfig {
...
     #[arg(short, long)]
     pub config_file: Option<String>,
+
+    /// [Developers only] Inject a SQL file into the database when starting the
+    /// manager.
+    ///
+    /// This is useful to populate the DB with state for testing.
+    #[serde(skip)]
+    #[arg(short, long)]
+    pub initial_sql: Option<String>,
```
ryzhyk, line 120: What's the use case for this? E.g., if you just want to add some projects to the DB, you could make some API calls after starting the server.

ryzhyk (commented):
Looks good, just want to clarify why we need these.

pipeline_manager/src/config.rs
gz: Let me check tomorrow if 0.0.0.0 makes a difference. I ran into a problem because the browser prevents the UI running on port X making requests to the manager running on port Y (you can enable it by configuring [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) correctly).

pipeline_manager/src/config.rs
gz: Originally added it to test UI which didn't have create functionality. Figured it might be useful in the future too for patching or schema migration etc.
Yes we can do it with API calls too but reading a SQL file seemed like a way to do it easier/faster. But yes, it's not strictly necessary to have.

pipeline_manager/src/config.rs
gz: I think I get it now, instead of the flag you want to compute it e.g., `dev_mode = host == 'localhost'`?

ryzhyk: > Let me check tomorrow if 0.0.0.0 makes a difference. I ran into a problem because the browser prevents the UI running on port X making requests to the manager running on port Y (you can enable it by configuring [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) correctly).

Oh, because the UI isn't currently hosted by the server.

> I think I get it now, instead of the CLI flag you want to compute it e.g., dev_mode = host == 'localhost'?

No, that's not what I meant. It makes sense now that I know what you're doing.

ryzhyk (approved):

PR #309, opened 2023-02-22
Pass to simplify and optimize docker container

lalithsuresh:
* Pipeline server: use cargo install and run the binary directly instead of using cargo build -> run -> start_manager.sh script
* Cleanup a bunch of build artifacts: the dbsp project /target folder and downloaded dependency packages
* Use OpenJDK JRE instead of JDK

@ryzhyk could you test your demo workflow end-to-end with these changes just to be sure I didn't break anything?

The changes bring the container size down from ~6.2GB to 4.2GB. Most of this is currently taken by the rust and llvm toolchains (~2GB total) and about 1.5GB in /usr/lib.

### DIFF
--- a/demo/hello-world/config.yml
+++ b/demo/hello-world/config.yml
@@ -4,7 +4,7 @@ inputs:
     transport:
       name: file
       config: 
-        path: "/examples/messages.csv"
+        path: "/database-stream-processor/demo/hello-world/messages.csv"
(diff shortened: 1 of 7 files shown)


ryzhyk (approved):

ryzhyk:
Actually looks like something's broken.  Looking into it.

PR #310, opened 2023-02-22
Fix regression introduced in PR#305.

ryzhyk:
PR #305 wraps the HTTP server in CORS middleware using most restrictive settings by default, which blocks access to HTTP endpoints from the browser connected to the same host/port.  This commit changes this behavior by only enabling CORS in permissive mode when the `--dev-mode` flag is specified.


mbudiu-vmw:
The code seems different from the commit message. The deleted code checks for Feb mode

mbudiu-vmw (approved):

PR #311, opened 2023-02-23
Mark the `project_name` field unique.

gz:
This means we can reasonably hide the project ID from the user.

### DIFF
--- a/pipeline_manager/src/db.rs
+++ b/pipeline_manager/src/db.rs
@@ -286,6 +286,26 @@ CREATE TABLE IF NOT EXISTS pipeline (
         Ok((version, code))
     }
 
+    /// Helper to convert rusqlite error into a `DBError::DuplicateProjectName`
+    /// if the underlying low-level error thrown by the database matches.
+    fn maybe_duplicate_project_name_err(e: rusqlite::Error, project_name: &str) -> AnyError {
+        if let rusqlite::Error::SqliteFailure(sqlite_failure, Some(msg)) = &e {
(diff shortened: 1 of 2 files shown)


gz (commented):
Also need to update error handling in https://github.com/vmware/database-stream-processor/pull/311/files#diff-d5b4d76084ac8d75b29640302d2a38081a6553508bc303210aeecb175da60856R300

ryzhyk (changes requested):
- We need to make sure that `new_project` returns a meaningful error message too.  It's often the first API call the client performs, which means that it may fail for more than one reason (e.g., old schema).  So we need to either distinguish the uniqueness violation SQlite error or check if the name exists as a separate query.
- This will also break workflows where project name doesn't matter and is left blank, e.g., because the client creates a project and uses it via an API handle. I guess this should be solved on the client side by either getting a list of projects and generating new unique names, e.g., "Project 5", or generating probabilistically unique names a la UUIDs.

gz:
I fixed the error handling; still need to look into the python API to generate a unique name by default.

ryzhyk:
> I fixed the error handling; still need to look into the python API to generate a unique name by default.

I don't think we need that now.  You upset the formatter though :)

ryzhyk (approved):

gz:
> > I fixed the error handling; still need to look into the python API to generate a unique name by default.
> 
> I don't think we need that now. You upset the formatter though :)

Fixed, for some reason I need to use the nightly version for cargo fmt :/

Issue #312, opened 2023-02-23
End-to-end test using docker


ryzhyk:
- Build a docker container on every commit in CI (just don't upload it to the repository).
- Run an end-to-end API test against it, e.g., using the python API test script.

This will help debug issues with the container and improve our end-to-end test coverage.

One thing missing currently is shared file system access, so the client cannot easily create files visible inside the container.

@lalithsuresh

lalithsuresh:
@ryzhyk we can close this?

PR #313, opened 2023-02-23
Integration tests using a docker container

lalithsuresh:
Addresses #312

This patch builds and runs the docker container
for the pipeline manager on every push. It then
invokes script from another container to run tests
against the pipeline manager container. We
use two containers because container-container
networking is easier on Github Actions than
host-container networking.

As of now, the integration-tests.sh file only does
a rudimentary `curl` call to see if a basic POST
to `/projects` succeeds.


.github/workflows/integration.yml
```diff
@@ -0,0 +1,23 @@
+
+on: push
```
ryzhyk, line 2: Shouldn't this be part other docker workflow?  I think that workflow should first run these steps to build and test the container (on each commit) and then push the tested container to the repo (on release commits only).

ryzhyk (approved):

.github/workflows/containers.yml
```diff
@@ -27,14 +42,13 @@ jobs:
 
       - name: Extract metadata (tags, labels) for Docker
         id: meta
+        if: github.event_name == 'release'
         uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
       - name: Build and push Docker image
```
ryzhyk, line 50: This should now be just "Push Docker image"

ryzhyk (approved):

PR #314, opened 2023-02-23
Move Rust crates to a directory

ryzhyk:
There's a lot going on in the top-level repository now, so it's time to separate Rust code from the rest of it.

- Move all crates, including the main `dbsp` crate, to the `crates` directory.
- Split nexmark query implementation and benchamarks into a separate crate.


mbudiu-vmw (approved):
Does the SQL compiler need changes?

PR #317, opened 2023-02-28
Return entire project meta-data when fetching code.

gz:
Avoids multiple API calls since we usually want to display more project information with the code.

### DIFF
--- a/crates/pipeline_manager/src/db.rs
+++ b/crates/pipeline_manager/src/db.rs
@@ -196,9 +196,7 @@ impl ProjectDB {
     /// database created with `create_db.sql` along with access credentials.
     pub(crate) fn connect(config: &ManagerConfig) -> AnyResult<Self> {
         unsafe {
-            rusqlite::trace::config_log(Some(|errcode, msg| {
-                debug!("sqlite: {msg}:{errcode}")
-            }))?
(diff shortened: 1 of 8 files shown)


crates/pipeline_manager/src/db.rs
```diff
@@ -115,7 +115,16 @@ impl ProjectStatus {
...
             Some("pending") => Ok(Self::Pending),
             Some("compiling") => Ok(Self::Compiling),
-            Some("sql_error") => Ok(Self::SqlError(error_string.unwrap_or_default())),
+            Some("sql_error") => {
+                let error = error_string.unwrap_or_default();
+                if let Ok(messages) = serde_json::from_str(&error) {
+                    Ok(Self::SqlError(messages))
+                }
+                else {
+                    log::debug!("Expected valid json for SqlCompilerMessage but got {:?}, did you update struct without adjusting the database?", error);
```
ryzhyk, line 124: This could be `log::warn` or even `log::error`.  Also, for consistency with the rest of the code base, let's import `log::error` in module preamble.

ryzhyk (commented):
Looks good. Let me take over from here and fix ts/python + the one issue I found.

gz:
Sounds good, at some point we might want to also have return code in the SqlCompilerMessages as it could probably return with an exit code of 0 and a bunch of warning messages only. We can do it later though, at some point maybe we want to rethink the DB schema to store exit code and message it in a separate field.

PR #318, opened 2023-03-02
deploy: Describe how to build the SQL compiler as a prerequisite.

blp:


deploy/README.md
```diff
@@ -1,3 +1,13 @@
+Prerequisites
+=============
+
+Before building the Docker image, clone the SQL-to-DBSP compiler at the root of this repository and then build it:
+
+```
+git clone https://github.com/vmware/sql-to-dbsp-compiler
+(cd sql-to-dbsp-compiler/SQL-compiler/ && mvn -DskipTests package)
```
mbudiu-vmw, line 8: what about required dependencies?
this is ugly, we should have a build.sh/build.bat command to do this.

mbudiu-vmw (approved):

deploy/README.md
ryzhyk: Yes, we need proper automation.  I suggest we rely on `cargo make` to automate the build process, including pulling dependencies. We should also address https://github.com/vmware/sql-to-dbsp-compiler/issues/119, so that the DBSP repo can depend on a specific compiler version. Alternatively (additionally?) we can add the sql compiler as git submodule to dbsp.

ryzhyk:
@blp , let me do some work today to better automate the process. Perhaps this PR won't be needed.

ryzhyk:
@blp, this step has been automated in #326.

PR #319, opened 2023-03-02
HTTP input and output transport

ryzhyk:
Added support for streaming data in and out of a DBSP pipeline via HTTP.  The
primary use case for this is browser and CLI-based testing, where one
wants to push some sample data and read outputs from the pipeline, tail
an output stream, etc.  It may be less suitable for production use,
since it may not be easy to build reliable communication on top of this
mechanism.

Implementation notes:

- We don't create seperate HTTP server(s) to run HTTP endpoints,
  instead, we extend the server that handles control plane communication
  (pause/start/status/...) with two new endpoints:
  `/http_input/{endpoint_name}` and `/http_output/{endpoint_name}`,
  which return a websocket connection to the specified endpoint.

- We use websockets, as they are likely going to give higher
  throughput/lower latency than, e.g., HTTP POST.  In addition,
  they eliminate the need to buffer and poll outputs.  In the future,
  we may add support for just posting data to the input endpoint
  in case the client doesn't care about performance and does not want
  the added complexity of a websocket.

- We implement a simple flow control protocol for the input connection:
  Each input message starts with an 8-byte unique id.  Upon receiving
  the complete message, the server sends back a 16-byte acknowledgement
  consisting of this id and the length of the message.  The server drops
  the message in paused state and sends a negative acknowledgement
  consisting of message id followed by an 8-byte 0.

This commit also removes the restriction that each output stream of
the circuit must be connected to exactly one transport endpoint since one
intended use case for HTTP endpoints is monitoring the contents
of a stream while also writing it to its actual destination.


crates/adapters/src/test/websocket.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /// Serialize `data` as `csv` and send it through web socket as one big
+    /// composite message with one frame per buffer.
+    pub async fn send_to_websocket_continuations(
+        mut ws: Pin<&mut Framed<impl AsyncRead + AsyncWrite, WsCodec>>,
+        data: &[Vec<TestStruct>],
+    ) {
+        if data.len() <= 1 {
+            return Self::send_to_websocket(ws, data).await;
```
mbudiu-vmw, line 57: why does this need to be a special case?

crates/adapters/src/test/websocket.rs
```diff
@@ -0,0 +1,152 @@
...
+use bytes::{BufMut, BytesMut};
+use csv::ReaderBuilder as CsvReaderBuilder;
+use csv::WriterBuilder as CsvWriterBuilder;
+use futures::{SinkExt, StreamExt};
+use std::pin::Pin;
+use tokio::io::{AsyncRead, AsyncWrite};
+
+pub struct TestWsSender;
+
+impl TestWsSender {
```
mbudiu-vmw, line 16: should this be `#[cfg(test)]`?

crates/adapters/src/transport/http/input.rs
```diff
@@ -0,0 +1,411 @@
...
+                    "HTTP input endpoint '{}': rejecting {}-byte message in paused state",
+                    self.endpoint.name(),
+                    bin.len()
+                );
+
+                let mut ack: [u8; 16] = [0; 16];
+                ack[0..8].copy_from_slice(&bin[0..8]);
+                ctx.binary(Bytes::copy_from_slice(&ack));
+            }
+            // First frame of a multipart message.  Save message tag, so we can send an
```
mbudiu-vmw, line 305: there isn't a library function to receive a multipart message?
When I used websockets I never worried about this stuff.
Maybe my messages weren't long enough

crates/adapters/src/transport/http/output.rs
```diff
@@ -0,0 +1,227 @@
...
+/// demand and sends output batches to these websockets.
+///
+/// This implementation provides no support for reliable delivery
+/// and is mostly intended for browser-based testing.
+#[derive(Clone)]
+struct HttpOutputEndpoint {
+    inner: Arc<HttpOutputEndpointInner>,
+}
+
+impl HttpOutputEndpoint {
```
mbudiu-vmw, line 108: could some of this be refactored into a base struct common to both Input and Output?

mbudiu-vmw (approved):

crates/adapters/src/test/websocket.rs
ryzhyk: To break message up into frames in the websocket protocol, one needs at least two frames (first and last).

crates/adapters/src/transport/http/input.rs
ryzhyk: I don't think the actix framework supports that today.  But the framework you used in JS most likely does.

PR #321, opened 2023-03-03
Use docker compose to bring up dbsp and redpanda containers

lalithsuresh:
We can now just run `docker compose up` to bring up a DBSP container alongside a Redpanda container.

Also simplifies the Docker container build process. We now avoid using git-ls and instead send the local folder over as is, with a `.dockerignore` file to avoid sending `/target` and some other files.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+# Ignore target and git directories
+target
+.git
+**/Dockerfile
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,7 @@
 target
 .git
 **/Dockerfile
+**/pipeline_data
+**/ldbc-graphalytics-data
+**/fraud_data
+demo
--- a/crates/adapters/src/server/mod.rs
+++ b/crates/adapters/src/server/mod.rs
@@ -141,6 +141,10 @@ pub fn run_server<F>(
 where
     F: Fn(usize) -> (DBSPHandle, Catalog),
 {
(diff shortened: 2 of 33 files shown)


ryzhyk (approved):

ryzhyk:
@lalithsuresh , can you check if you can run `docker compose` and then run the `demo/docker_demo.sh` script.  The 4 demo projects should appear in the ui on `localhost:8085`. Try compiling and running one of them, e.g., TimeSeries enrich.

lalithsuresh:
@ryzhyk The docker part works fine. The demo script however needs a lot of dependencies. I've so far had to install `jq` and `rpk`, and now I'm at:

```
/Users/lsuresh/code/database-stream-processor/demo/project_demo01-TimeSeriesEnrich/prepare.sh: line 18: gdown: command not found
/Users/lsuresh/code/database-stream-processor/demo/project_demo01-TimeSeriesEnrich/prepare.sh: line 23: mapfile: command not found
/Users/lsuresh/code/database-stream-processor/demo/project_demo01-TimeSeriesEnrich/prepare.sh: line 29: /Users/lsuresh/code/database-stream-processor/demo/project_demo01-TimeSeriesEnrich/transactions.csv: No such file or directory
```

ryzhyk:
@lalithsuresh , I am not too worried about the demo script, it will need to be packaged differently eventually. This is only to test that the docker setup works not only on my machine.

* For `gdown` you need `pip install gdown`.
* mapfile is harder, looks like it's not supported on MacOS.

PR #322, opened 2023-03-04
Add separate state for rust compilation.

gz:
This allows the UI to determine at what point a program is correct SQL.

### DIFF
--- a/crates/pipeline_manager/src/db.rs
+++ b/crates/pipeline_manager/src/db.rs
@@ -115,7 +115,8 @@ impl ProjectStatus {
             None => Ok(Self::None),
             Some("success") => Ok(Self::Success),
             Some("pending") => Ok(Self::Pending),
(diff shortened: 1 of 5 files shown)


crates/pipeline_manager/src/db.rs
```diff
@@ -205,7 +207,7 @@ impl ProjectDB {
 CREATE TABLE IF NOT EXISTS project (
     id integer PRIMARY KEY AUTOINCREMENT,
     version integer,
-    name varchar UNIQUE,
+    name varchar,
```
ryzhyk, line 210: I thought we're keeping this for now.

gz: yes my mistake will revert

gz: fixed

PR #324, opened 2023-03-07
Add a schema parameter to the ProjectDesc.

gz:
The schema is generated by the compiler using `-js` and contains info about tables and views in the SQL program.

### DIFF
--- a/crates/pipeline_manager/src/compiler.rs
+++ b/crates/pipeline_manager/src/compiler.rs
@@ -4,6 +4,7 @@ use fs_extra::{dir, dir::CopyOptions};
 use log::{debug, error, trace};
 use serde::{Deserialize, Serialize};
 use std::{
+    path::PathBuf,
     process::{ExitStatus, Stdio},
     sync::Arc,
 };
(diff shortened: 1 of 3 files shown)


crates/pipeline_manager/src/compiler.rs
```diff
@@ -177,18 +178,26 @@ impl Compiler {
```
gz, line 178: is the db.lock() here also starting a transaction on the db?

I put the schema update in the `db.set_project_status_guarded` because I want to ensure the schema exists when status >= CompilingRust to avoid a race but if this is a transaction maybe setting the schema can be a separate query.

crates/pipeline_manager/src/db.rs
```diff
@@ -498,44 +521,56 @@ CREATE TABLE IF NOT EXISTS pipeline (
...
     ///
     /// # Note
     /// This intentionally does not throw an error if there is a project version
-    /// mismatch and instead does just not update. It's used by the compiler
-    /// to update status and in case there is a newer version it is expected
-    /// that the compiler just picks up and runs the next job.
+    /// mismatch and instead does just not update. It's used by the compiler to
+    /// update status and in case there is a newer version it is expected that
+    /// the compiler just picks up and runs the next job.
     pub(crate) fn set_project_status_guarded(
```
gz, line 550: One thing: maybe it's better to just do it with two queries in a transaction rather than polluting the set_project_status_guarded api.

ryzhyk (commented):
In an offline discussion, we agreed to make set_schema a separate method.

PR #325, opened 2023-03-07
#321 postfix: read in batches of inputs at a time

lalithsuresh:
This method to read a batch of inputs at a time should work across various shells and be consistent between Linux and Mac OSX. I suspect it's not as fast as mapfile/readarray, but fast enough to run `demo/docker_demo.sh` in about 10 minutes.


ryzhyk:
It works for me on Linux.

PR #326, opened 2023-03-08
Make SQL compiler a git submodule

ryzhyk:
- Add SQL-to-DBSP compiler as a submodule to the repo.
- Change Dockerfile to update the submodule before building the SQL compiler.
- Remove the no longer needed docker.sh.
- Improve the `start_manager.sh` script:
  - Update and build the SQL compiler submodule before running the manager.
  - Point the manager to `~/.dbsp` as default working directory -- this way the script can be run without arguments.
- Change pipeline manager `Makefile.toml` to build the compiler before running end-to-end tests.

With these changes, one should be able to fork the repo and then start the service with one command (`./scripts start_manager.sh`).


lalithsuresh:
What's the expected way to test the SQL compiler during the typical local dev cycle? Will we always need a commit that can be referenced on the sql compiler repo?

PR #337, opened 2023-03-16
Fixed quicksort bug

Kixiron:
Fixed bug in columnar quicksort impl that caused values to be forgotten


crates/dbsp/src/trace/consolidation/tests/proptests.rs
```diff
@@ -236,4 +236,17 @@ proptest! {
         prop_assert_eq!(values2, expected_values2);
         prop_assert_eq!(values3, expected_values3);
     }
+
+    #[test]
+    fn quicksort_correctness(mut batch in vec(any::<(u16, ())>(), 0..50_000)) {
```
ryzhyk, line 241: This test wouldn't catch the bug addressed by this PR, which would require a non-unit value type.

ryzhyk (approved):

crates/dbsp/src/trace/consolidation/tests/proptests.rs
Kixiron: It'll still detect missed/duplicate values, with non-zst values sorting stability comes into play

PR #340, opened 2023-03-17
`Trie::truncate_below`

ryzhyk:
This is the first step toward garbage collecting old time series state.
Consider a time series trace indexed by (some representation of) physical time
that is only used by "windowed" operators, i.e., operators that don't look
further than a bounded number of seconds into the past.  Once all users
of the trace have moved their windows beyond a certain timestamp,
older values no longer need to be stored in the trace.

Internally traces consist of batches that are in turn built on top of
`Trie`s, so we start with adding GC functionality to the `Trie` API.  We
add two new methods:

`truncate_below()` - informs the trie that keys below a given bound will no
longer be accessed.  No state gets deallocated at this point.  This
operation affects the behavior of the trie in two ways:
  - The default cursor returned by the `cursor()` method starts
    iterating from the current bound and not from the start of the trie.
  - Merging two tries using `merge()` does not include truncated tuples
    in the result.

`lower_bound()` - returns the current value of the bound.


crates/dbsp/Cargo.toml
```diff
@@ -59,8 +59,9 @@ rand_xoshiro = "0.6.0"
 indicatif = "0.17.0-rc.11"
 clap = { version = "3.2.8", features = ["derive", "env"] }
 reqwest = { version = "0.11.11", features = ["blocking"] }
-time = { version = "0.3.14", features = [
+time = { version = "0.3.20", features = [
     "formatting",
+    "macros",
     "serde",
     "serde-human-readable",
 ] }
```
Kixiron, line 67: Probably should break this into a table, it'll need to be moved below the other deps first

```suggestion
[dependencies.time]
version = "0.3.20"
features = [
    "formatting",
    "macros",
    "serde",
    "serde-human-readable",
]
```

crates/dbsp/src/trace/layers/column_layer/mod.rs
```diff
@@ -202,8 +215,8 @@ impl<K, R> ColumnLayer<K, R> {
         let mut shifter = BackshiftOnDrop {
             keys: &mut self.keys,
             diffs: &mut self.diffs,
-            processed_len: 0,
-            deleted_cnt: 0,
+            processed_len: self.lower_bound,
+            deleted_cnt: self.lower_bound,
```
Kixiron, line 219: I'm not sure this is correct, this will leak all items at `0..lower_bound`

Kixiron (changes requested):

crates/dbsp/src/trace/layers/column_layer/mod.rs
ryzhyk: Thanks for catching this.  Does the new version look better?

crates/dbsp/src/trace/layers/column_layer/mod.rs
```diff
@@ -221,7 +238,7 @@ impl<K, R> ColumnLayer<K, R> {
                 let current_diff =
                     unsafe { &mut *shifter.diffs.as_mut_ptr().add(shifter.processed_len) };
 
-                if !retain(current_key, current_diff) {
+                if !retain(current_key, current_diff) || shifter.processed_len < lower_bound {
```
Kixiron, line 241: The `lower_bound` stuff can probably be broken out into a single `drop_in_place::<*mut [T]>()` call to drop `0..lower_bound`, probably outside of `process_loop()`

crates/dbsp/src/trace/layers/column_layer/mod.rs
```diff
@@ -148,6 +161,9 @@ impl<K, R> ColumnLayer<K, R> {
         // since we may make some holes during the process.
         unsafe { self.set_len(0) };
 
+        let lower_bound = self.lower_bound;
+        self.lower_bound = 0;
```
Kixiron, line 165: Around here we should drop all items from `0..lower_bound` if those items are useless to us

crates/dbsp/src/trace/layers/test.rs
```diff
@@ -0,0 +1,674 @@
...
+    consolidate(&mut tuples);
+
+    let mut builder = <Tr::TupleBuilder>::with_capacity(tuples.len());
+    builder.extend_tuples(
+        tuples
+            .iter()
+            .cloned()
+            .map(|((k, v, t), r)| (k, (v, (t, r)))),
+    );
+    println!("builder: {builder:?}");
```
Kixiron, line 174:
```suggestion
```

crates/dbsp/src/trace/layers/test.rs
```diff
@@ -0,0 +1,674 @@
...
+    let result = builder.done();
+    println!("result: {result:?}");
+    result
```
Kixiron, line 177:
```suggestion
    builder.done()
```

crates/dbsp/src/trace/layers/column_layer/tests.rs
Kixiron, line 1: I'd kinda like a test that uses drop bombs to make sure the correct number of values are dropped (drop bomb would consist of something like a wrapped `Rc<RefCell<BTreeMap<usize, bool>>>` that sets a value to `false` for each unique drop bomb, which each has a `usize` id)

Kixiron: This is mostly unrelated to panics, it's also about memory leaks occurring within normal operation

crates/dbsp/src/trace/layers/column_layer/mod.rs
```diff
@@ -144,6 +157,17 @@ impl<K, R> ColumnLayer<K, R> {
...
+        // SAFETY: We initialize BackshiftOnDrop with `lower_bound` offset below,
+        // so these elements will never get touched again.
+        unsafe {
+            ptr::drop_in_place(&mut self.keys[0..lower_bound]);
+            ptr::drop_in_place(&mut self.diffs[0..lower_bound]);
+        }
+
         // Avoid double drop if the drop guard is not executed,
         // since we may make some holes during the process.
         unsafe { self.set_len(0) };
```
Kixiron, line 173:
```suggestion
        // SAFETY: We initialize BackshiftOnDrop with `lower_bound` offset below,
        // so these elements will never get touched again.
        unsafe {
            // Avoid double drop if the drop guard is not executed,
            // or if dropping elements up to `lower_bound` panics
            // since we may make some holes during the process.
            self.set_len(0);
        
            // Drop all keys up to `lower_bound`
            let keys = ptr::slice_from_raw_parts_mut(self.keys.as_mut_ptr(), lower_bound);
            ptr::drop_in_place(keys);
            
            // Drop all diffs up to `lower_bound`
            let diffs = ptr::slice_from_raw_parts_mut(self.diffs.as_mut_ptr(), lower_bound);
            ptr::drop_in_place(diffs);
        }
```

ryzhyk: what was the problem with my code?

Kixiron: If a panic occurred while dropping either of the slices then the drop implementations of `keys` and `diffs` would still run, causing a double free on the elements `0..lower_bound` within them

PR #342, opened 2023-03-24
Migrate to sqlx

lalithsuresh:
Initial lift to migrate the code from rusqlite to sqlx to interface with a database.

Some key changes:
* DB methods are all async now
* Schema changes to accommodate sqlx quirks
* Cannot test a connection before daemonization because tokio does not survive forks

Remaining todos:
* Test with postgres. Will require ManagerConfig to be populated with connection strings directly.
* More tests for db.rs.
* Use proptests?

### DIFF
--- a/crates/pipeline_manager/src/db.rs
+++ b/crates/pipeline_manager/src/db.rs
@@ -1,10 +1,14 @@
-use crate::{ManagerConfig, ProjectStatus};
+use crate::ProjectStatus;
 use anyhow::{anyhow, Error as AnyError, Result as AnyResult};
 use chrono::{DateTime, NaiveDateTime, Utc};
 use log::{debug, error};
-use rusqlite::Connection;
 use serde::{Deserialize, Serialize};
-use std::{error::Error as StdError, fmt, fmt::Display};
+use sqlx::{
+    any::{AnyConnectOptions, AnyRow},
+    pool::PoolOptions,
+    Any, ConnectOptions, Pool, Row,
+};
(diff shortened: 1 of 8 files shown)


crates/pipeline_manager/src/db.rs
```diff
@@ -886,15 +947,171 @@ CREATE TABLE IF NOT EXISTS pipeline (
...
+                killed: row.get::<i32, _>(3) != 0,
                 created: DateTime::<Utc>::from_utc(created_naive, Utc),
             });
         }
 
         Ok(result)
     }
 }
+
+#[cfg(test)]
```
ryzhyk, line 963:
```rust
#[cfg(test)]
mod test {
    // All code below this line.
}
```

crates/pipeline_manager/src/db.rs
```diff
@@ -886,15 +947,171 @@ CREATE TABLE IF NOT EXISTS pipeline (
...
+    let expected = ProjectDescr {
+        project_id: res.0,
+        name: "test1".to_string(),
+        description: "project desc".to_string(),
+        version: res.1,
+        status: ProjectStatus::None,
+        schema: None,
+    };
+    let actual = rows.get(0).unwrap();
+    assert_eq!(expected.name, actual.name);
```
ryzhyk, line 1016: `#[derive(Eq)]` for `ProjectDescr`, and:

```rust
assert_eq!(expected, actual);
```

ryzhyk (approved):

PR #347, opened 2023-03-31
Timestamp and Date functions

Kixiron:
Implements the following functions/intrinsics in the JIT (all functions can be called with the `Call` instruction)

- Timestamp functions
  - `@dbsp.timestamp.epoch()`
  - `@dbsp.timestamp.year()`
  - `@dbsp.timestamp.month()`
  - `@dbsp.timestamp.day()`
  - `@dbsp.timestamp.quarter()`
  - `@dbsp.timestamp.decade()`
  - `@dbsp.timestamp.century()`
  - `@dbsp.timestamp.millennium()`
  - `@dbsp.timestamp.iso_year()`
  - `@dbsp.timestamp.week()`
  - `@dbsp.timestamp.day_of_week()`
  - `@dbsp.timestamp.iso_day_of_week()`
  - `@dbsp.timestamp.day_of_year()`
  - `@dbsp.timestamp.millisecond()`
  - `@dbsp.timestamp.microsecond()`
  - `@dbsp.timestamp.second()`
  - `@dbsp.timestamp.minute()`
  - `@dbsp.timestamp.hour()`
  - `@dbsp.timestamp.floor_week()`
  - `@dbsp.timestamp.to_date()`

- Date functions
  - `@dbsp.date.year()`
  - `@dbsp.date.month()`
  - `@dbsp.date.day()`
  - `@dbsp.date.quarter()`
  - `@dbsp.date.decade()`
  - `@dbsp.date.century()`
  - `@dbsp.date.millennium()`
  - `@dbsp.date.iso_year()`
  - `@dbsp.date.week()`
  - `@dbsp.date.day_of_week()`
  - `@dbsp.date.iso_day_of_week()`
  - `@dbsp.date.day_of_year()`
  - `@dbsp.date.epoch()`
  - `@dbsp.date.hour()`
  - `@dbsp.date.second()`
  - `@dbsp.date.minute()`
  - `@dbsp.date.millisecond()`
  - `@dbsp.date.microsecond()`
  - `@dbsp.date.to_timestamp()`
  
  Additionally this implements a few missing binary operations, namely `DivFloor` (floored division),`Rem` (remainder), `Mod` (modulus/euclidean remainder) and `ModFloor` (floored modulus)


crates/dataflow-jit/src/codegen/intrinsics.rs
```diff
@@ -296,6 +330,94 @@ unsafe extern "C" fn dataflow_jit_string_with_capacity(capacity: usize) -> ThinS
...
+    day => |time| time.month() as i64,
+    hour => |time| time.hour() as i64,
+    minute => |time| time.minute() as i64,
+    second => |time| time.second() as i64,
+    millisecond => |time| (time.second() * 1000 + time.timestamp_subsec_millis()) as i64,
+    microsecond => |time| (time.second() * 1_000_000 + time.timestamp_subsec_micros()) as i64,
+    day_of_week => |time| (time.weekday().num_days_from_sunday() as i64) + 1,
+    iso_day_of_week => |time| (time.weekday().num_days_from_monday() as i64) + 1,
+    day_of_year => |time| time.ordinal() as i64,
+    floor_week => |time| {
```
mbudiu-vmw, line 373: This does not look correct. 
The definition of the week number is insane:
"ISO 8601 includes the ISO week-date system, a numbering system for weeks. Each week is associated with the year in which Thursday occurs. For example, week 1 of 2004 (2004W01) covered Monday, 29 December 2003 to Sunday, 4 January 2004. European countries/regions typically use this style of numbering. Non-European countries/regions typically don't use it." 
This comes from https://learn.microsoft.com/en-us/sql/t-sql/functions/datepart-transact-sql?view=sql-server-ver16
I really would prefer to use the existing Chrono libraries instead of implementing this by hand.
All of these functions already exist in sqllib/src/timestamp.rs, it would be great to reuse that code.

mbudiu-vmw: sorry, this comment is not valid; floow_week is right. it applies to week.

mbudiu-vmw:
This looks fine, we should test this stuff end to end and then I will be more confident.
One thing I don't know is about the safety for nullable values. For arithmetic I can execute the arithmetic operation even if the value is null, but I should avoid calling the functions if the values are null. So I will probably need a branch.

ryzhyk:
Looks good to me, assuming we draw the line here and implement other types and functions via the (future) FFI mechanism.

Kixiron:
> This looks fine, we should test this stuff end to end and then I will be more confident. One thing I don't know is about the safety for nullable values. For arithmetic I can execute the arithmetic operation even if the value is null, but I should avoid calling the functions if the values are null. So I will probably need a branch.

Correct, these functions aren't null-safe apart from `epoch`

mbudiu-vmw:
I think this can be merged

mbudiu-vmw:
@Kixiron can you provide an example on how to use these functions for a nullable timestamp?

Kixiron:
> @Kixiron can you provide an example on how to use these functions for a nullable timestamp?

Not at the moment, but you'd just branch to only execute the function when the value is non-null, it's identical to any other nullish branching

PR #350, opened 2023-04-01
Synchronize Pipeline runner with GUI

gz:
- Added Connector Entities
- Added AttachedConnector Entities.
- Introduce 1:1 relationship between config and pipeline.
- Changed logic to generate a config based attached connectors.
- Downgrade sqlx to v0.6 as 0.7 is alpha and has a bug that can't deal with NULL values.


crates/pipeline_manager/src/db.rs
```diff
@@ -261,40 +321,69 @@ impl ProjectDB {
...
+            is_input boolean NOT NULL,
+            FOREIGN KEY (config_id) REFERENCES project_config(id) ON DELETE CASCADE
+            FOREIGN KEY (connector_id) REFERENCES connector(id) ON DELETE CASCADE)"#,
+        )
+        .execute(&conn)
+        .await?;
+
+        sqlx::query(
+            r#"
+            CREATE TABLE IF NOT EXISTS connector (
```
lalithsuresh, line 380: We need to grow test coverage for the pipeline manager. Given that you've added some new tables and methods, could you also add some tests?

scripts/start_manager.sh
```diff
@@ -37,4 +37,5 @@ cd "${MANAGER_DIR}" && ~/.cargo/bin/cargo run --release -- \
     --sql-compiler-home="${SQL_COMPILER_DIR}" \
     --dbsp-override-path="${ROOT_DIR}" \
     --static-html=static \
-    --unix-daemon
+    --unix-daemon \
+    --dev-mode
```
lalithsuresh, line 41: dev-mode shouldn't be the default in start_manager.sh.

crates/pipeline_manager/src/main.rs
```diff
@@ -1348,3 +1419,219 @@ async fn pipeline_delete(state: WebData<ServerState>, req: HttpRequest) -> impl
...
+        (status = OK, description = "connector successfully updated.", body = UpdateConnectorResponse),
+        (status = NOT_FOUND
+            , description = "Specified `connector_id` does not exist in the database."
+            , body = ErrorResponse
+            , example = json!(ErrorResponse::new("Unknown connector id '5'"))),
+    ),
+    tag = "Connector"
+)]
+#[patch("/connectors")]
+async fn update_connector(
```
lalithsuresh, line 1542: @ryzhyk can confirm, but I think you'd also need to update the python API + tests for these changes I think.

lalithsuresh (changes requested):

crates/pipeline_manager/src/main.rs
ryzhyk: That's correct.  I can help with that.  We should probably mark this as draft PR for now.

crates/pipeline_manager/src/db.rs
gz: I'm planning to solve test-coverage with proptests & a model; but ideally I can do it with the next PR after this one which revamps the way we do versioning (which would mean tweak many tests again anyways).

lalithsuresh: Okay, sounds good.

python/dbsp/config.py
```diff
@@ -118,11 +128,50 @@ def run(self) -> DBSPPipeline:
         """
 
         # print("yaml:\n" + self.yaml())
+        attached_connectors = []
+        for (key, input) in self.pipeline_config.inputs.to_dict().items():
+            # TODO: Stream gets added by the attached connector
```
gz, line 133: @ryzhyk with the new connector API the stream gets injected through the attached connector, shall I remove stream from InputEndpointConfig and OutputEndpointConfig. I wasn't really aware of these types for the UI so I just stumbled upon them in the python API.

crates/pipeline_manager/src/db.rs
```diff
@@ -789,77 +888,125 @@ impl ProjectDB {
...
+            .bind(config_description)
             .bind(config)
             .fetch_one(&self.conn)
             .await?;
-        Ok((ConfigId(row.get(0)), Version(1)))
+        let config_id = ConfigId(row.get(0));
+
+        if let Some(connectors) = connectors {
+            // Add the connectors.
+            // TODO: This should be done in a transaction with the query above.
```
gz, line 979: How does it work, is this whole function executed in a transaction? @lalithsuresh @ryzhyk ?

crates/pipeline_manager/src/db.rs
```diff
@@ -868,18 +1015,48 @@ impl ProjectDB {
...
+        if let Some(connectors) = connectors {
+            // Delete all existing attached connectors.
+            sqlx::query("DELETE FROM attached_connector WHERE config_id = ?")
+                .bind(config_id.0)
+                .execute(&self.conn)
+                .await?;
+
+            // Rewrite the new set of connectors.
+            for ac in connectors {
+                // TODO: This should be done in a transaction with the query above.
```
gz, line 1047: same q here

python/dbsp/config.py
ryzhyk: Yes, I think it's ok to remove them.  This is basically a strongly typed representation of the config yaml, which mirrors what `serde` creates when deserializing the yaml on the server side. Which is why these types don't appear directly in the OpenAPI spec.

One nice thing about your new API is that it doesn't require these hidden types any longer. But now we need to model the new entities (connectors and attachments) in the API, and the 1-1 relationship between pipelines and configs. We can discuss offline.

crates/pipeline_manager/src/runner.rs
```diff
@@ -254,24 +246,75 @@ scrape_configs:
...
+        // Assemble the final config by including all attached connectors.
+        async fn generate_attached_connector_config(
+            db: &ProjectDB,
+            config: &mut String,
+            ac: &AttachedConnector,
+        ) -> AnyResult<()> {
+            let ident = 4;
+            config.push_str(format!("{:ident$}{}:\n", "", ac.config.as_str()).as_str());
+            let ident = 8;
+            config.push_str(format!("{:ident$}stream: {}\n", "", ac.config.as_str()).as_str());
```
ryzhyk, line 267: Something's off here.  We're pushing the same string twice.

crates/pipeline_manager/src/db.rs
```diff
@@ -320,21 +402,36 @@ impl ProjectDB {
...
+        sqlx::query(
+            format!(
+                r#"
+        CREATE TABLE IF NOT EXISTS attached_connector (
+            id {},
+            uuid varchar UNIQUE NOT NULL,
+            config_id integer NOT NULL,
+            connector_id integer NOT NULL,
+            config varchar,
+            is_input boolean NOT NULL,
```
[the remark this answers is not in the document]
gz: If connector is a boolean and can't be both we don't need it here anymore yes.

crates/pipeline_manager/src/db.rs
```diff
@@ -431,11 +528,15 @@ impl ProjectDB {
...
-    fn maybe_duplicate_project_name_err(e: sqlx::Error, project_name: &str) -> AnyError {
+    fn maybe_duplicate_project_name_err(e: sqlx::Error, _project_name: &str) -> AnyError {
         match e {
             sqlx::Error::Database(db_error) => {
-                if (*db_error).is_unique_violation() {
-                    anyhow!(DBError::DuplicateProjectName(project_name.to_string()))
+                if db_error
+                    .message()
+                    .to_lowercase()
+                    .contains("unique constraint")
```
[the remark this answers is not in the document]
gz: This change was to make it portable -- maybe the sqlx downgrade to 0.6 changed it but postgres test was failing.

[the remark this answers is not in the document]
gz: I remember now: `is_unique_violation` does not exist yet in sqlx0.6

crates/pipeline_manager/src/db.rs
```diff
@@ -502,7 +601,7 @@ impl ProjectDB {
                 // current version.
                 version = version.increment();
                 sqlx::query(
-                        "UPDATE project SET version = $1, name = $2, description = $3, code = $4, status = '', error = '' WHERE id = $5")
+                        "UPDATE project SET version = ?, name = ?, description = ?, code = ?, status = NULL, error = NULL, schema = NULL WHERE id = ?")
```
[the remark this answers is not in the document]
gz: not supporting is a bug in sqlx 0.7 I believe (it's alpha) it works fine in 0.6 and ? is less error-prone

crates/pipeline_manager/src/db.rs
```diff
@@ -900,19 +1077,72 @@ impl ProjectDB {
...
+
+        Ok(result)
+    }
+
+    async fn attach_connector(
+        &self,
+        config_id: ConfigId,
+        ac: &AttachedConnector,
+    ) -> AnyResult<AttachedConnectorId> {
+        let _descr = self.get_config(config_id).await?;
```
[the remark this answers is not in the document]
gz: only called within db.rs (not pub) so never supposed to not have config_id, will add a comment

crates/pipeline_manager/src/runner.rs
gz: it generates this part:

DEMOGRAPHICS:
   stream: DEMOGRAPHICS

shall we call the top-level thing something else than stream (e.g., connector-{id} or something?)

crates/pipeline_manager/src/db.rs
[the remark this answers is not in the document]
lalithsuresh: Does 0.6 support $n? If so, let's keep it that way.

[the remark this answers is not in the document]
gz: The dollar $n is very error prone: whenever you insert/remove a field you need to adjust many unrelated lines (every row.get()) and it's easy to introduce a mistake there -- I made a couple while working on this.

The ? approach is much saner in this regard.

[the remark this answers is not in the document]
gz: if sqlx 0.7 really won't support it anymore, I'll revert it back :P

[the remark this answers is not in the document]
lalithsuresh: All right!

crates/pipeline_manager/src/db.rs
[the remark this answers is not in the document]
ryzhyk: Then it's probably worth keep commented out `is_unique_violation` line, so we can enable it when 0.7 is stable.

crates/pipeline_manager/src/db.rs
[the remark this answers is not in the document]
ryzhyk: Would be nice to use `?` consistently across this module then.

crates/pipeline_manager/src/runner.rs
ryzhyk: - This will generate an invalid config when two input connectors connect to the same input stream of the circuit (because endpoint name (the first `DEMOGRAPHICS` above) must be unique).  Maybe we can just use uuid?
- The  `ac.config` stream should be renamed to `stream`.

> shall we call the top-level thing something else than stream (e.g., connector-{id} or something?)

Which one's the top-level thing?

gz: top-level-thing is not the "stream:", will rename it.

python/dbsp/config.py
```diff
@@ -17,92 +18,169 @@
...
+        self.name = name
+        self.description = description
 
-    def add_input(self, name: str, input_endpoint_config: InputEndpointConfig):
+    def add_input(self, stream: str, input_endpoint_config: DBSPConnector):
         """Add an input endpoint to the pipeline configuration.
 
         Args:
-            name (str): Endpoint name (must be unique across input endpoints).
+            stream (str): Input name (what table to connect to).
```
ryzhyk, line 60: This is related to the other issue in `runner.rs`: we need both stream (i.e., table) name and unique endpoint name. Is the plan to use `uuid` for the latter? If so, then this is good.

python/dbsp/config.py
```diff
@@ -17,92 +18,169 @@
...
+        self.attached_connectors.append(AttachedConnector(
+            connector_id=input_endpoint_config.connector_id,
+            direction=Direction.INPUT,
+            uuid=uuid.uuid4().hex,
+            config=stream,
+        ))
 
-    def add_file_input(self, stream: str, filepath: str, format_: FormatConfig, **file_config_options):
+    def add_kafka_input(self, name: str, stream: str, config: KafkaInputConfig, format: FormatConfig):
         """Add an input endpoint that reads data from a file to the pipeline configuration.
```
ryzhyk, line 73: file -> Kafka

python/dbsp/config.py
```diff
@@ -17,92 +18,169 @@
...
 
         Args:
-            stream (str): Input stream name to connect the endpoint to.
+            name (str): Name of the input endpoint.
```
ryzhyk, line 76: I no longer know what the terminology means, but I guess this is connector name, not endpoint name?

python/dbsp/config.py
```diff
@@ -17,92 +18,169 @@
...
-                    format_ = format_))
-
-    def add_output(self, name: str, output_endpoint_config: OutputEndpointConfig):
+            stream,
+            DBSPConnector(self.api_client, name, ConnectorType.KAFKAIN, TransportConfig(
+                name="kafka",
+                config=config), format=format))
+
+    def add_kafka_output(self, name: str, stream: str, config: KafkaOutputConfig, format: FormatConfig):
+        """Add an input endpoint that reads data from a file to the pipeline configuration.
```
ryzhyk, line 89: file -> Kafka

python/dbsp/config.py
```diff
@@ -17,92 +18,169 @@
...
+
+        Args:
```
ryzhyk, line 91: This comment no longer matches the function. Is there a better way to document arguments in Python?

python/dbsp/config.py
```diff
@@ -17,92 +18,169 @@
...
+        ))
 
-    def add_file_output(self, stream: str, filepath: str, format_: FormatConfig, **file_config_options):
+    def add_file_input(self, stream: str, filepath: str, format: FormatConfig):
+        """Add an input endpoint that reads data from a file to the pipeline configuration.
+
+        Args:
+            stream (str): Input table the stream is connected to.
+            filepath (str): File to read data from.
+            format_ (FormatConfig): Data format specification, e.g., CsvInputFormatConfig().
```
ryzhyk, line 126: Underscore no longer needed.

crates/pipeline_manager/src/db.rs
```diff
@@ -216,31 +242,66 @@ pub(crate) struct ProjectDescr {
...
 
-fn to_option(val: String) -> Option<String> {
-    if val.is_empty() {
-        None
-    } else {
-        Some(val)
+/// Type of new data connector.
+#[derive(Serialize, Deserialize, ToSchema, Debug, Copy, Clone)]
+pub enum ConnectorType {
+    KafkaIn = 0,
```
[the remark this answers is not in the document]
gz: Will fix in next PR

crates/pipeline_manager/src/db.rs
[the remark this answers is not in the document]
gz: Will fix in next PR

python/dbsp/config.py
```diff
@@ -17,92 +18,169 @@
...
         self.config_version = None
-        # self.workers = workers
-        # print("config: " + str(self.pipeline_config))
+        self.attached_connectors = []
+        self.pipeline_id = None
+        self.name = name
+        self.description = description
 
-    def add_input(self, name: str, input_endpoint_config: InputEndpointConfig):
+    def add_input(self, stream: str, input_endpoint_config: DBSPConnector):
```
ryzhyk, line 56: input_endpoint_config -> connector (also update the doc comment)

crates/pipeline_manager/src/runner.rs
```diff
@@ -254,24 +246,75 @@ scrape_configs:
...
+        fs::write(&config_file_path, config.as_str()).await?;
 
-        let (_version, code) = db.project_code(request.project_id).await?;
+        let (_version, code) = db.project_code(project_id).await?;
 
         let metadata = PipelineMetadata {
-            project_id: request.project_id,
-            version: request.project_version,
+            project_id,
+            version: request.config_version,
```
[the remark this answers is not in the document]
gz: agreed, I'll get rid of versions as discussed offline in next PR.

python/dbsp/config.py
gz: Yes, I changed it to uuid.

crates/pipeline_manager/src/db.rs
```diff
@@ -321,20 +403,35 @@ impl ProjectDB {
...
+        CREATE TABLE IF NOT EXISTS connector (
             id {},
-            project_id integer NOT NULL,
-            project_version integer NOT NULL,
-            -- TODO: add 'host' field when we support remote pipelines.
-            port integer,
-            killed integer NOT NULL,
-            created integer NOT NULL,
-            FOREIGN KEY (project_id) REFERENCES project(id) ON DELETE CASCADE)",
+            name varchar NOT NULL,
```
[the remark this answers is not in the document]
lalithsuresh: I think it's fine to ask users to name even their connectors (e.g. "kafka-staging-cluster"). The API could then prevent entities with duplicate names from being POST-ed, unless there's an argument that forces an "upsert".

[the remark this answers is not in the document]
ryzhyk: It's probably fine.  I suppose with connectors the risk of accidentally overwriting an unrelated connector is higher though.

Another idea is to add an `ephemeral` flag to configurations and connectors, which indicates that the config/connectors should get deleted once not in use.  This is probably the cleanest solution actually. Killing the pipeline associated with config will cleanup the config and any of its ephemeral connectors.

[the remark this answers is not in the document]
lalithsuresh: How about a config option "Lifetime: pipeline" or "Lifetime: permanent" (not sure I like permanent).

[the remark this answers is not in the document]
ryzhyk: Yeah, something like that. Not just config though, connectors as well. Let's think this through and plan for a fix in a (near-)future PR.

ryzhyk (approved):

PR #354, opened 2023-04-04
Fix dependabot complaints about openssl and spin versions

lalithsuresh:


mbudiu-vmw (approved):

ryzhyk (approved):

crates/pipeline_manager/Cargo.toml
```diff
@@ -30,6 +30,9 @@ utoipa-gen = { version = "=3.0.3" }
 utoipa-swagger-ui = { version = "3.0.2", features = ["actix-web"] }
 chrono = { version = "0.4.23", default-features = false, features = ["serde"] }
 sqlx = { version = "0.7.0-alpha.1", features = [  "runtime-tokio-native-tls", "sqlite", "postgres", "any" ] }
+openssl = "0.10.48" # override dependency in sqlx
+spin = "0.9.8"      # override dependency in sqlx
```
Kixiron, line 34: We shouldn't need to explicitly add them as dependencies, simply running `cargo update` should suffice

Kixiron (changes requested):

Kixiron (approved):

PR #355, opened 2023-04-06
Some preparatory commits for being able to run on Kubernetes

lalithsuresh:


python/dbsp/config.py
```diff
@@ -70,6 +70,21 @@ def add_file_input(self, stream: str, filepath: str, format_: FormatConfig, **fi
                         name = "file",
                         config = FileInputConfig.from_dict(dict({ 'path': filepath }, **file_config_options))),
                     format_ = format_))
+
```
ryzhyk, line 73: These Python API changes will need to be aligned with @gz's PR.

ryzhyk (approved):
I don't really understand most of this k8s magic :)

python/dbsp/config.py
lalithsuresh: I'll wait on him to merge his PR and then rebase.

PR #362, opened 2023-04-12
Optimize nexmark q9.

ryzhyk:

### DIFF
--- a/crates/dbsp/src/trace/layers/ordered/mod.rs
+++ b/crates/dbsp/src/trace/layers/ordered/mod.rs
@@ -311,7 +311,7 @@ where
 }
 
 /// Assembles a layer of this
-#[derive(SizeOf, Debug)]
+#[derive(SizeOf, Debug, Clone)]
 pub struct OrderedBuilder<K, L, O = usize> {
     /// Keys
     pub keys: Vec<K>,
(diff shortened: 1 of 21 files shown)


crates/dbsp/src/trace/layers/ordered/mod.rs
```diff
@@ -453,6 +453,108 @@ where
...
+                        self.copy_range_fueled(source2, *lower2, upper2, remaining_fuel as usize);
+                }
+            }
+        }
+
+        effort = (self.vals.keys() - starting_updates) as isize;
+
+        *fuel -= effort;
+
+        // if *fuel < -1_000_000 {
```
lalithsuresh, line 511: Looks like leftovers from debugging

crates/dbsp/src/trace/layers/ordered/mod.rs
```diff
@@ -453,6 +453,108 @@ where
...
+    /// Like `copy_range`, but uses `fuel` to bound the amount of work.
+    ///
+    /// Builds at most `fuel` values plus values for one extra key.
+    /// If `fuel > 0` when the method returns, this means that the merge is
+    /// complete.
+    pub fn push_merge_fueled(
+        &mut self,
+        input1: (&<Self as Builder>::Trie, &mut usize, usize),
+        input2: (&<Self as Builder>::Trie, &mut usize, usize),
+        fuel: &mut isize,
```
Kixiron, line 471: We could/should just use a `usize` with `.saturating_sub()`s imo

crates/dbsp/src/trace/layers/ordered/mod.rs
```diff
@@ -453,6 +453,108 @@ where
...
+        // }
+    }
+
+    /// Like `copy_range`, but uses `fuel` to bound the amount of work.
+    ///
+    /// Invariants:
+    /// - Copies at most `fuel` values plus values for one extra key.
+    /// - If `fuel` is greater than or equal to the number of values, in the
+    ///   `lower..upper` key range, copies the entire range.
+    pub fn copy_range_fueled(
```
Kixiron, line 522: Is `.copy_range()` intensive enough for fuel limiting to matter? Wouldn't we be better off unconditionally coping the entire range over even if there's not technically enough fuel to do every element? Otherwise I can easily see situations where we've got a handful of elements left just kinda dangling that have to be re-addressed later on

Kixiron (changes requested):

crates/dbsp/src/trace/layers/ordered/mod.rs
ryzhyk: The idea is that we can track over-spent fuel this way.  We don't do this now, but it makes sense for the API to capture this.

crates/dbsp/src/trace/layers/ordered/mod.rs
ryzhyk: This is precisely the behavior that this PR fixes :) Turns out `copy_range` can be expensive when running unchecked.

crates/dbsp/src/trace/layers/ordered/mod.rs
Kixiron: Interesting, I guess that could be a useful metric

Kixiron (approved):

mbudiu-vmw (approved):

PR #364, opened 2023-04-13
Added the `@dbsp.error.abort()` function and the `unreachable` terminator

Kixiron:
Adds the `@dbsp.error.abort()` function to immediately abort the program and the `unreachable` terminator for terminating blocks with no meaningful branching/return behavior, currently should just be used as
```llvm
call @dbsp.error.abort()
unreachable
```

### DIFF
--- a/crates/dataflow-jit/src/codegen/call.rs
+++ b/crates/dataflow-jit/src/codegen/call.rs
@@ -1,6 +1,7 @@
 use crate::{
     codegen::{
-        utils::FunctionBuilderExt, CodegenCtx, VTable, TRAP_ASSERT_EQ, TRAP_CAPACITY_OVERFLOW,
+        utils::FunctionBuilderExt, CodegenCtx, VTable, TRAP_ABORT, TRAP_ASSERT_EQ,
+        TRAP_CAPACITY_OVERFLOW,
     },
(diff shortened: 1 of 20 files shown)


crates/dataflow-jit/src/codegen/call.rs
```diff
@@ -100,6 +103,16 @@ impl CodegenCtx<'_> {
         }
     }
 
+    fn error_abort(&self, builder: &mut FunctionBuilder<'_>) {
+        let trap = builder.ins().trap(TRAP_ABORT);
+
+        if let Some(writer) = self.comment_writer.as_deref() {
+            writer
+                .borrow_mut()
+                .add_comment(trap, "call @dbsp.error.abort()");
```
mbudiu-vmw, line 112: in the future we should add information to help users identify where the error happened. For example, having some kind of error message from the compiler could work.

mbudiu-vmw (approved):
How about adding some JSON-based tests too which show how these should be used by the compiler?

crates/dataflow-jit/src/codegen/call.rs
Kixiron: Yah, I'll get around to that

Kixiron:
Fixes #368, #371 and #369

PR #375, opened 2023-04-15
Use pg-embed for embedded postgres.

gz:
- [x] use pg-embed: This allows use to use postges with similar ergonomics as sqlite for running tests/debugging locally. No installation/setup is required (e.g., it's automated by using pg-embed).
- [x] drop sqlite support
- [x] replace sqlx with tokio-postgres
- [x] add model for test coverage
- [ ] fix docker test
- [x] need to make sure pg-embed installation doesn't run concurrently (test failure)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3249,7 +3249,7 @@ dependencies = [
 [[package]]
 name = "pg-embed"
 version = "0.7.2"
-source = "git+https://github.com/gz/pg-embed.git?rev=116a466#116a466269cc0bfb6c50f4bd21c2533f824f0624"
+source = "git+https://github.com/gz/pg-embed.git?rev=8906af8#8906af840b2cc083e3ec61c42f28638807cf76f6"
 dependencies = [
  "archiver-rs",
  "async-trait",
--- a/crates/pipeline_manager/Cargo.toml
+++ b/crates/pipeline_manager/Cargo.toml
@@ -36,7 +36,7 @@ futures = "0.3"
 tokio-postgres = "0.7"
 async-trait = "0.1"
 # Waiting for https://github.com/faokunega/pg-embed/pull/26
-pg-embed = { git = "https://github.com/gz/pg-embed.git", rev = "116a466" }
+pg-embed = { git = "https://github.com/gz/pg-embed.git", rev = "8906af8" }
 
 [target.'cfg(unix)'.dependencies]
 daemonize = { version = "0.4.1" }


crates/pipeline_manager/src/config.rs
```diff
@@ -237,25 +237,21 @@ impl ManagerConfig {
         Path::new(&self.working_directory).join("cargo_workspace")
     }
 
+    /// Where Postgres embed stores the database.
+    pub(crate) fn postgres_embed_data_dir(&self) -> PathBuf {
+        Path::new(&self.working_directory).join("data")
```
lalithsuresh, line 242: Better to create a temporary folder. Adding folders to the source repo bloats the docker context.

crates/pipeline_manager/src/db/test.rs
```diff
@@ -0,0 +1,1140 @@
...
+            .iter_mut()
+            .map(|p| p.created = DateTime::default())
+            .collect::<Vec<_>>()
+    );
+}
+
+/// Compare the database storage implementation with our model using in-memory
+/// data-structures. Ideally, the two behave the same.
+#[test]
+fn db_impl_behaves_like_model() {
```
lalithsuresh, line 325: Why are we maintaining a model in addition to the DB?

crates/pipeline_manager/src/config.rs
gz: If we're running local then you'd want to keep the data around, no? 
(e.g., if you restart the manager it will find the db in data/ (instead of sqlite which kept manager.db in the working dir)

crates/pipeline_manager/src/db/pg_setup.rs
```diff
@@ -0,0 +1,43 @@
...
+/// - `persistent` - If true, the database will be persistent. If false, the
+///   database will be deleted after the process exits.
+/// - `port` - The port on which the database will be available on.
+pub(crate) async fn install(
+    database_dir: PathBuf,
+    persistent: bool,
+    port: Option<u16>,
+) -> AnyResult<PgEmbed> {
+    let pg_settings = PgSettings {
+        database_dir,
```
lalithsuresh, line 21: Is this only used by tests and pg-embed? The previous implementation actually supported all kinds of postgres connection strings.

gz: Only used by pg-embed+test, postgres install will still take a connection string

crates/pipeline_manager/src/config.rs
lalithsuresh: I see. So it'll be in ~/.dbsp or something by default. Does the data get wiped in between tests?

crates/pipeline_manager/src/db/test.rs
gz: Now we have good test coverage ;) -- ideally the model could be less code, but it did find a number of bugs in the impl while writing it so that was good

crates/pipeline_manager/src/config.rs
ryzhyk: @lalithsuresh , I think "current working directory" is the DBSP working directory, usually `~/.dbsp`, not the Unix `cwd`.

ryzhyk: And I imagine that for test runs the DBSP working directory could be under `/tmp`.

gz: Yes

Test setup is slightly different: see here https://github.com/vmware/database-stream-processor/pull/375/files#diff-66137f70e107475f6293e863373cc39a0dcf6bd92443def5ed6c08d377e15f33R43

gz: well it's test.rs line 43 in case this link doesn't work

crates/pipeline_manager/src/db/test.rs
lalithsuresh: I mean, if we can write a test against the model, we should just as well write that test against the DB no?

gz: Not sure I follow; every sequence of actions is evaluated against both implementations model  PostgresDB?. The test makes sure they return the same things.

crates/pipeline_manager/src/db/pg_setup.rs
ryzhyk: I think the answer is yes. A module-level comment would help to clarify this.

crates/pipeline_manager/src/db/storage.rs
```diff
@@ -0,0 +1,263 @@
+use super::{
+    AttachedConnector, ConfigDescr, ConfigId, ConnectorDescr, ConnectorId, ConnectorType, DBError,
+    PipelineDescr, PipelineId, ProjectDescr, ProjectId, Version,
+};
+use crate::{Direction, ProjectStatus};
+use anyhow::{anyhow, Result as AnyResult};
+use async_trait::async_trait;
+
+#[async_trait]
+pub(crate) trait Storage {
```
ryzhyk, line 10: Can we add a doc comment for this trait?

crates/pipeline_manager/src/db/test.rs
```diff
@@ -0,0 +1,1140 @@
...
+            };
+        }
+    }
+}
+
+async fn test_setup() -> DbHandle {
+    let _temp_dir = tempfile::tempdir().unwrap();
+    let temp_path = _temp_dir.path();
+
+    let port = DB_PORT_COUNTER.fetch_add(1, Ordering::Relaxed);
```
ryzhyk, line 46: This is fragile.  You don't want to assume that all these ports are available.  Having said that there is no easy way to locate unused ports.  This is the best I could do:

```rust
use std::net::TcpListener;

let listener = TcpListener::bind(("127.0.0.1", 0))?;
let port = listener.local_addr()?.port();
```
Problem is, once you drop the listener, the port can get stolen by another process.  Ideally Postgres itself would support dynamic port allocation in some form.

crates/pipeline_manager/src/db/test.rs
```diff
@@ -0,0 +1,1140 @@
+use super::PipelineDescr;
```
ryzhyk, line 1: This is good stuff!

ryzhyk (approved):

crates/pipeline_manager/src/db/test.rs
lalithsuresh: Ah, got it. Sounds good. We'll need to plan for transactions/aborts/rollbacks at some point, but this is a good way to catch bugs as you say.

crates/pipeline_manager/src/db/test.rs
lalithsuresh: I wouldn't spend too much time trying to fix this. Binding to an ephemeral port is an anti-pattern for long running servers anyway, and is something we'll have to fix later. I suggest the dev-mode form factor have a runner service on a fixed port that brings up pipelines locally.

gz: Fixed

gz:
@lalithsuresh  docker test fails with:

```
[2023-04-19T23:47:31Z DEBUG pg_embed::pg_access] Unpacking postgres binaries /root/.cache/pg-embed/linux/amd64/15.2.0/linux-amd64-15.2.0.zip /root/.cache/pg-embed/linux/amd64/15.2.0
[2023-04-19T23:47:32Z ERROR pg_embed::postgres] pg_ctl: cannot be run as root
[2023-04-19T23:47:32Z ERROR pg_embed::postgres] Please log in (using, e.g., "su") as the (unprivileged) user that will
[2023-04-19T23:47:32Z ERROR pg_embed::postgres] own the server process.
```

Need advise on how to fix this. I suspect we should use the postgres through docker now? Alternative is to run with non-root user.

PR #376, opened 2023-04-17
Make `Map` polymorphic

Kixiron:
Makes `Map` polymorphic over set and map inputs and outputs

### DIFF
--- a/crates/dataflow-jit/src/ir/nodes/filter_map.rs
+++ b/crates/dataflow-jit/src/ir/nodes/filter_map.rs
@@ -2,7 +2,7 @@ use crate::ir::{
     function::Function,
     layout_cache::RowLayoutCache,
     nodes::{DataflowNode, StreamKind, StreamLayout},
-    LayoutId, NodeId,
+    InputFlags, LayoutId, NodeId,
 };
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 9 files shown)


crates/dataflow-jit/src/ir/nodes/filter_map.rs
```diff
@@ -52,27 +63,88 @@ impl DataflowNode for Map {
...
-    }
+    fn output_kind(&self, _inputs: &[StreamLayout]) -> Option<StreamKind> {
+        Some(self.output_layout.kind())
+    }
+
+    fn output_stream(&self, _inputs: &[StreamLayout]) -> Option<StreamLayout> {
+        Some(self.output_layout)
+    }
+
+    fn validate(&self, inputs: &[StreamLayout], _layout_cache: &RowLayoutCache) {
```
mbudiu-vmw, line 74: is panic the best way to fail validation?

mbudiu-vmw (approved):

crates/dataflow-jit/src/ir/nodes/filter_map.rs
Kixiron: Right now it is, I need to refactor the validation system

mbudiu-vmw: I would suggest filing an issue about it, this way we'll know it's planned and not an oversight.

PR #380, opened 2023-04-18
Garbage collection work

ryzhyk:
We can now run window queries and rolling aggregate queries with bounded out-of-ordedness using bounded memory.

Next step is to discuss how this can be used by the SQL compiler.


crates/dbsp/src/trace/layers/column_layer/builders.rs
```diff
@@ -108,8 +108,10 @@ where
             trie2.assume_invariants();
         }
 
-        let (mut lower1, upper1) = cursor1.bounds();
-        let (mut lower2, upper2) = cursor2.bounds();
+        let (_, upper1) = cursor1.bounds();
+        let mut lower1 = cursor1.position();
+        let (_, upper2) = cursor2.bounds();
+        let mut lower2 = cursor2.position();
```
Kixiron, line 114: Why isn't `.bounds()` sufficient here?

crates/dbsp/src/trace/test_batch.rs
```diff
@@ -0,0 +1,603 @@
...
+}
+
+impl<K, V, T, R> TestBatch<K, V, T, R>
+where
+    K: DBData,
+    V: DBData,
+    T: DBTimestamp,
+    R: DBWeight,
+{
+    /// Create batch from sorted on unsorted tuples.
```
mbudiu-vmw, line 104: on?

mbudiu-vmw (approved):
Can't say I followed all details, but I have some idea

PR #384, opened 2023-04-28
Refactor PgEmbed usage

lalithsuresh:
* Conditionally build pg-embed use outside tests using a "dev" feature flag, enabled by default
* Separate PgEmbed out from ProjectDB
* Update Docker build to use non-default builds

### DIFF
--- a/crates/pipeline_manager/src/config.rs
+++ b/crates/pipeline_manager/src/config.rs
@@ -24,7 +24,7 @@ fn default_sql_compiler_home() -> String {
 }
 
 fn default_db_connection_string() -> String {
-    "postgres-embed".to_string()
+    "".to_string()
 }
 
 /// Pipeline manager configuration read from a YAML config file or from command
@@ -228,6 +228,7 @@ impl ManagerConfig {
     }
 
     /// Where Postgres embed stores the database.
+    #[cfg(feature = "dev")]
     pub(crate) fn postgres_embed_data_dir(&self) -> PathBuf {
         Path::new(&self.working_directory).join("data")
     }
--- a/crates/pipeline_manager/src/db/mod.rs
+++ b/crates/pipeline_manager/src/db/mod.rs
@@ -1051,7 +1051,7 @@ impl ProjectDB {
         }
 
         debug!("Opening connection to {:?}", connection_str);
-        let (client, conn) = tokio_postgres::connect(&connection_str, NoTls).await?;
+        let (client, conn) = tokio_postgres::connect(connection_str, NoTls).await?;
 
         // The `tokio_postgres` API requires allocating a thread to `connection`,
         // which will handle datbase I/O and should automatically terminate once
--- a/crates/pipeline_manager/src/db/mod.rs
+++ b/crates/pipeline_manager/src/db/mod.rs
@@ -12,7 +12,7 @@ use utoipa::ToSchema;
 #[cfg(test)]
 mod test;
 
-#[cfg(feature = "dev")]
+#[cfg(any(test, feature = "dev"))]
 mod pg_setup;
 pub(crate) mod storage;
 
--- a/crates/pipeline_manager/src/db/mod.rs
+++ b/crates/pipeline_manager/src/db/mod.rs
@@ -12,6 +12,7 @@ use utoipa::ToSchema;
 #[cfg(test)]
 mod test;
 
+#[cfg(feature = "dev")]
 mod pg_setup;
 pub(crate) mod storage;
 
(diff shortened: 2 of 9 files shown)


crates/pipeline_manager/src/config.rs
```diff
@@ -24,7 +24,7 @@ fn default_sql_compiler_home() -> String {
 }
 
 fn default_db_connection_string() -> String {
-    "postgres-embed".to_string()
+    "".to_string()
```
gz, line 27: Should this be postgres:// now or something similar?

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -28,10 +28,14 @@ pub(crate) mod storage;
...
 pub(crate) struct ProjectDB {
     conn: Client,
-    #[allow(dead_code)] // We don't have to interact with it, but dropping it stops the DB server.
-    pub pg: Option<pg_embed::postgres::PgEmbed>,
 }
 
+// Used in dev mode for having an embedded Postgres DB live through the lifetime
+// of the program.
+#[cfg(feature = "dev")]
+static PG: once_cell::sync::OnceCell<pg_embed::postgres::PgEmbed> =
```
gz, line 36: If the program crashes will it call `drop()` on PgEmbed (and hence stop Postgres) if it runs in once_cell? I think it's wasy to test by running it an hitting ctrl+c or kill-9. I remember it was a struggle getting it to call drop with tokio once cell stuff in tests, but maybe a regular once_cell is fine.

crates/pipeline_manager/src/db/pg_setup.rs
```diff
@@ -3,47 +3,50 @@
 //! It caches the downloaded binaries as well. Check the `pg_embed` crate for
 //! more details.
 
-use crate::AnyResult;
-use pg_embed::pg_enums::PgAuthMethod;
-use pg_embed::pg_fetch::{PgFetchSettings, PG_V15};
-use pg_embed::postgres::{PgEmbed, PgSettings};
-use std::path::PathBuf;
+#[cfg(feature = "dev")]
```
gz, line 6: if you instead make the 

```rust
mod pg_setup;
```

in mod.rs 

as 

```rust
#[cfg(feature = "dev")]
mod pg_setup;
```

you wont have to make an extra module (embedded) for it.

crates/pipeline_manager/src/db/test.rs
```diff
@@ -3,10 +3,11 @@ use super::{
     storage::Storage, AttachedConnector, ConfigDescr, ConfigId, ConnectorDescr, ConnectorId,
     ConnectorType, PipelineId, ProjectDB, ProjectDescr, ProjectId, ProjectStatus, Version,
 };
-use crate::db::DBError;
+use crate::db::{pg_setup, DBError};
```
gz, line 6: so `mod test` in mod.rs maybe also needs to be conditionally included with cfg(feature = dev)?

gz (approved):

crates/pipeline_manager/src/config.rs
lalithsuresh: tokio-postgres doesn't fill things in automatically, so "postgres://" does not work.

crates/pipeline_manager/src/db/mod.rs
lalithsuresh: I suspect all bets are off when we call kill-9 unless we trap the signal?

mbudiu-vmw: I don't think you can trap SIGKILL, only SIGTERM

crates/pipeline_manager/src/db/test.rs
lalithsuresh: I've updated the CI workflows to have feature=dev wherever --no-default-features is called.

crates/pipeline_manager/src/db/mod.rs
lalithsuresh: @gz It's better to not rely on drop for this. Currently, the pipeline manager doesn't restart correctly because the DDL statements being issued are not idempotent. The fix for that is to do a pass on the schema as we discussed (some of the 1:1 mappings should be a single table).

gz: you're right about SIGKILL, but I'm wondering if OnceCell content gets dropped if the user hits ctrl+c (otherwise there will be a lingering postgres process). Just mentioned because I think for the tests I had to put it in a struct member to get drop() called.

I can fix the 1:1 mappings as once the webui stuff is merged.

crates/pipeline_manager/src/db/test.rs
gz: I believe it will result in a compilation error if you do `cargo test --no-default-features` which is a bit awkward. You could update to include the `crate::db::pg_setup` cfg(feature = "dev") and in addition also for `cfg(test)`?

gz: e.g., `cfg(any(test, feature = "dev")` might do the trick?

lalithsuresh: Do you mean `all(test, feature = "dev)` ?

crates/pipeline_manager/src/db/mod.rs
gz: I tried it; looks like drop gets called on the OnceCell.

PR #394, opened 2023-05-03
Bump `daemonize` to 0.5.0.

ryzhyk:
We have two RUSTSEC alerts related to daemonize:

1. `daemonize` uses `boxfnonce` crate, which is obsolete https://github.com/vmware/database-stream-processor/issues/328

2. `daemonize` itself is unmaintained. https://github.com/vmware/database-stream-processor/issues/304

We bump `daemonize` to a recent 0.5 release, which resolves (1), and I guess the fact that there _is_ a recent release indicates that it's no longer unmaintained.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -444,9 +444,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.3.0"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e579a7752471abc2a8268df8b20005e3eadd975f585398f17efcfd8d4927371"
+checksum = "0ca84f3628370c59db74ee214b3263d58f9aadd9b4fe7e711fd87dc452b7f163"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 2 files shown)


lalithsuresh:
@ryzhyk frankly, I don't think the daemonization adds much value. I'd drop it. If we want to run it in the background, we can just use `nohup <command>`.

ryzhyk:
@lalithsuresh , not a fan myself, especially because it is indeed not very well maintained. But I'm not too worried because we don't use it in "real" deployment modes.  For local dev use it's kind of convenient.

mbudiu-vmw:
I thought that @blp implemented a very rich crate for this purpose.

lalithsuresh:
> @lalithsuresh , not a fan myself, especially because it is indeed not very well maintained. But I'm not too worried because we don't use it in "real" deployment modes. For local dev use it's kind of convenient.

Why not just use `nohup` during local dev in places where we use --unix-daemon today? It'll be just as convenient and has the benefit that you get the job ID right away.

ryzhyk:
- `daemonize` uses pid file to prevent multiple instances
- with `daemonize` we can fail early before disconnecting from terminal
- it will take me a day to deal with all the hiccups of the transition to `nohup`, which I'd rather spend on other stuff :)

what's the harm of keeping `daemonize` around for now?

PR #396, opened 2023-05-03
Refactor runner code to support multiple runner types

lalithsuresh:
See individual commits

### DIFF
--- a/crates/adapters/src/server/mod.rs
+++ b/crates/adapters/src/server/mod.rs
@@ -245,7 +245,6 @@ where
         .service(dump_profile)
         .service(input_endpoint)
         .service(output_endpoint)
-        .service(kill)
 }
 
 #[get("/start")]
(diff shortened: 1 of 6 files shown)


crates/adapters/src/server/mod.rs
```diff
@@ -330,7 +329,12 @@ async fn shutdown(state: WebData<ServerState>) -> impl Responder {
     let controller = state.controller.lock().unwrap().take();
     if let Some(controller) = controller {
         match controller.stop() {
```
ryzhyk, line 331: Both the `stop` call and the actix termination command below can be slow or even get stuck in some cases. We will need a way to enforce eventual termination from the pipeline manager by killing the process/pod after a timeout.  This is outside the scope of this PR.  I will create an issue.

ryzhyk (approved):

